PR #59, opened 2015-02-27
Enable GL_ARB_texture_non_power_of_two on modern graphics hardware

coelckers:
If the extension is present, which should be the case for all hardware currently in use, this allows using all of Quake 2's textures in their original size, which especially makes the models look a lot better.


DanielGibson:
merged, thanks!

(no idea why github doesn't recognize that)

PR #65, opened 2015-03-29
OSX

Jarvik7:
Minor cleanup
Fix 32bit architecture detection (probably)

Todo: Statically link or copy dylibs to the .app file so end users do not need to install dependencies.


Yamagi:
That looks good. Just two small remarks:

  * I read you comment about reverting the README changes bit had no time do so. 
  * If we need / want to integrate ObjC code it may be time to thing about an own main.c for OS X.

Jarvik7:
@Yamagi 
I think rather than an Objective C GUI that starts before the game, maybe it would be a good idea to integrate a mod selector into the normal Q2 menu. Once the mod is selected the game can restart with that as the gamedir so intro movies etc are shown.
If that is done the user can select the mission pack/ctf/whatever and no platform specific code will be needed.

PR #72, opened 2015-05-21
Fixed compilation on OSX with sdl2

DonnoDK:
SDL2 is referenced as a framework on OSX.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -249,7 +249,7 @@ SDLLDFLAGS := -lSDL
 endif # SDL2
 else ifeq ($(OSTYPE), Darwin)
 ifeq ($(WITH_SDL2),yes)
-SDLLDFLAGS := -lSDL2 -framework OpenGL -framework Cocoa
+SDLLDFLAGS := -framework SDL2 -framework OpenGL -framework Cocoa
 else # not SDL2
 SDLLDFLAGS := -lSDL -framework OpenGL -framework Cocoa
 endif # SDL2


Jarvik7:
I don't see the need for this commit.
The other dependencies will still need to be downloaded using homebrew. Making SDL2's dependency a manually downloaded framework just complicates the procedure.

DonnoDK:
@Jarvik7 Sure, you need to download SDL2. But having done so, the compilation would fail, unless you link against it using `-framework`. It simply does not work using `-lSDL2`, unless the you do some manual placement of SDL2 in `/usr/lib` or similar, which is not according to SDL2 installation procedures. The pull request change only affects OSX and only in the case that SDL2 was chosen as an option.

Jarvik7:
Following the OSX compilation directions and installing SDL2 using Homebrew should work. It works for me and at least one other person who has commented about the OSX build in any case.

Does it not work for you following the directions? What is your error?

DonnoDK:
@Jarvik7 Ok, I was not aware of that. No, it did not work for me, having installed SDL2 using libsdl.org's guidelines, placing SDL2.framework in `/Library/Frameworks`. I did this before using yquake2, and for other purposes.

Jarvik7:
OSX support was officially deleted a few months ago since it wasn't being maintained by anyone.

In response I re-wrote the makefile to be much simpler and actually work (OpenAL was broken in the existing code), while not requiring any OSX-specific support in the actual C code (which the main devs are not able to maintain).

Part of that rewrite is using Homebrew to manage dependencies instead of manually downloading them individually.

DonnoDK:
@Jarvik7 I was not aware, disregard the pull request.

DonnoDK:
@Jarvik7 I am using MacPorts and I have no intention of using Homebrew - thanks you for the suggestion. I agree, static linking and later pre-compiling an .app for end-users would be the holy grail.

Jarvik7:
MacPorts may work as well if there is an SDL2 package on it (that installs libs/headers instead of a framework). If you try it, let me know and we can add alternative instructions to the readme for compilation.

I've always avoided MacPorts since it tends to make a huge mess compared to homebrew. Or, at least it did when I last tried it like 8 years ago.

DonnoDK:
@Jarvik7 Works with SDL2 installed through Macports as well. In my experience, Homebrew tends to break on OS updates, while Macports does not since it encapsulates its own little ecosystem.

PR #84, opened 2015-08-09
Cmake build system

bradc6:
Here is an initial pull request to gauge the interest in switching from Makefiles to CMake. I am sure there are additional code changes to capture all the makefile's functionality and will update this pull request as needed.


Yamagi:
Hello,
Daniel tried to convince me to merge this. And since I'm a user of Jetbrains Clion (an C/C++ IDE based upon CMake as a project management tool) I'll do so. But before some small caveats must be solved. This is what I've found so far:
* Linking -ldl is a linuxism. Nearly all other unixoid platforms have dlopen() and it's friends integrated into libc. So that library needs to be conditional to Linux only, otherwise things will break will break.
* "${SDL2_INCLUDE_DIR}/.." breaks at least on FreeBSD, where the SDL2 headers are in SDL2/. The compiler error is (of course): .../src/backends/sdl/input.c:41:10: fatal error: 'SDL.h' file not found
* You need to pass DEFAULT_OPENAL_DRIVER to the compiler. Otherwise no default OpenAL driver is set and OpenAL won't work without user intervention. This generates a warning, caused by the fallback default of "1": ...src/backends/generic/qal.c:338:36: warning: incompatible integer to pointer conversion passing 'int' to parameter of type 'char *' [-Wint-conversion] 
* OGG/Vorbis support is still missing.
Nevertheless, thank you for your hard work so far :)

bradc6:
No problem, I'll resolve these issues and make a new comment on this pull request when it is ready. Glad to hear there is interest in using Cmake. :)

mackal:
There is ${CMAKE_DL_LIBS} to make at least one of the issues better. http://www.cmake.org/cmake/help/v3.0/variable/CMAKE_DL_LIBS.html

bradc6:
Builds successfully on Linux/Mac/FreeBSD. Need to test Windows.

Yamagi:
Now I'm okay with this and I'll merge it sometime in the in the next few days. Thank you for your adjustments. :)

Regarding Windows: After some tinkering with CMake I was able to generate a GNU Makefile, but it's far away from working. I think that's caused by our broken Windows support. For example we still insist on hardlinking SDL2.dll and the last time I checked some alterations to the JPEG headers were needed. I'll need to spend some time to clean that mess up.

DanielGibson:
JPEG headers? We replaced libjpeg with stb_image :-)

Yamagi:
I'm an old man who forgets things like that... But libjpeg is only part of the problem.

Yamagi:
I've done some additional changes, especially integration of several required compiler flags: https://github.com/yquake2/yquake2/commits/cmake Is everyone okay with this?

bradc6:
For OSX we need to dlopen "game.dylib" instead of "game.so"

Yamagi:
Okay, another iteration:
* Extended the comment in line 17 to include Clang
* Reworked the Makefile to generate a game.dynlib on OS X and changed the source to detect OS X with the canonical __APPLE__ macro. This should have been done years ago.
* Add headers to CMake. While they are not needed for compilation they are necessary when CMake is used as project management tool. For example CLion is unable to enumerate header dependencies without them.

bradc6:
Is the old makefile going to be deleted? (Ensuring that we don't have two different build systems that could get different settings.)

Yamagi:
No, we'll keep the Makefile. Deleting it would force package maintainers to alter their build scripts, automatic build system generators are a PITA on not so common systems like NetBSD or Solaris and at least at this time CMake is unable to build our broken code on Windows.

smcv:
Sorry, I've just realised the patch you merged is incomplete: `src/game/savegame/savegame.c` uses `BUILD_DATE` without including `src/common/header/common.h`, so it won't work in the upstream (non-reproducible) case.

I'm not completely clear on which modules are meant to include which headers; `savegame.c` can either include `common.h`, or redo the `#ifndef/#define` itself. Which resolution would you prefer?

PR #91, opened 2015-10-06
Update README

sdfdfhgh:
minor


Yamagi:
Can we have a better commit message? :)

sdfdfhgh:
Only for having more accurate instructions.

Yamagi:
I've merged the pull request with an updated commit message. Thank you.

PR #95, opened 2015-10-14
Patch to allow ignoring default music playback behaviour

ewe2:
Standard client behaviour is to stop or pause music playback on attainment of a goal. This is ambiguously applied and does not restart music except on a map reload. This patch ignores that behaviour where applicable via a compile-time definition. Ideally, it should be a cvar but this is an initial idea with low impact on the source.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,12 @@
 # User configurable options
 # -------------------------
 
+# Ignores original playback behaviour when goals are reached. Unfortunately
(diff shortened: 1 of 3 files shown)


Yamagi:
Why not intercept the track 0 behavior in src/client/sound/ogg.c:501? Doing it there would be generic, your patch requires changes to all game modules...

ewe2:
Ok, I wasn't sure if it would satisfy the same test, I'll rebase the patch around that file.

PR #96, opened 2015-10-14
Patch to ignore standard playback when Ogg support is enabled.

ewe2:
This patch is rebased on master and only affects Ogg playback if WITH_OLDPLAYBACK is set to no. CD playback if enabled will still exhibit the old behaviour ignoring that setting. We would have to alter the main source to do so. See pull request #95


maraakate:
Why don't you just make it a CVAR?  Compile time flags for something this trivial doesn't make much sense to me?

ewe2:
At least with this patch, the main source is untouched if people want a pristine version; if I added cvars I may as well prevent the CD code from triggering track 0 as well.

maraakate:
It's up to Yamagi, but I think a CVAR would make more sense as it's something so small it's really not going to hurt anybody by having it as an option in the code.

Yamagi:
I've merged the other version with a cvar

PR #97, opened 2015-10-14
Add ogg_ignoretrack0 cvar to control default music playback

ewe2:
This patch is an alternate to #96 using a cvar. As the function logic currently stands, this will result in a harmless message that track 00 is not in the playlist when ogg_ignoretrack0 is set to 1. Some documentation of the cvar will be needed if accepted.


ewe2:
Good point. Have updated the pull request.

DanielGibson:
ogg_track0 is not very descriptive

ewe2:
ogg_toggletrack0, ogg_ignoretrack0, ogg_keeploop, ogg_ignoregoal ? Pick something.

DanielGibson:
ogg_ignoretrack0 sounds good to me, but in the end it's up to @Yamagi :)

ewe2:
It's a bit wordy but at least it means something :D

Yamagi:
I've merged this pull request manually after combining the commits into one.

Issue #54, opened 2014-11-30
models (enemy, items) brightness


crmb:
This is about the Windows pre-compiled binaries 5.23 (Win7 x64 + Geforce)

3D models (player hand, items & enemy) seems to be not gamma corrected

i checked with vanilla 3.20 (with the broken dark picture), look at the heath items, or player hand on below screenshots :
with 3.2 the lightning is consistent, with yquake2 the models are too dark

![3 20_1](https://cloud.githubusercontent.com/assets/10013125/5239581/f614d76e-78e6-11e4-801e-9dd1c906af16.jpg)
![yq2_1](https://cloud.githubusercontent.com/assets/10013125/5239582/f641f1e0-78e6-11e4-929f-70b7ffe0435e.jpg)
![3 20_2](https://cloud.githubusercontent.com/assets/10013125/5239502/40a3b49c-78e4-11e4-9990-8cab9ac388b3.jpg)
![yq2_2](https://cloud.githubusercontent.com/assets/10013125/5239501/40a2d3f6-78e4-11e4-9ad7-b62d457eb3cc.jpg)

Yamagi:
That's a nice find. I wonder why this is only seen on Windows and not on Linux or even FreeBSD. Maybe some kind of driver issue on either side. Nevertheless, not only models are too dark but all dynamic lights. Can you try to set gl_modulate to the same value as vid_gamma? That should solve / work around the problem.

Regarding new builds: I planed to release 3.24 in December but had no time. I'll try it again at the end of January. :) But it'll only be another bugfix release without new features. The biggest changes are
* The HUD scale stuff is now working as expected, the menu and console can be scaled too. 
* Some progress stoppers when playing 'rogue' in coop with friendly fire of were fixed. 
* The underwater sound now works with the SDL backend.

deuxsonic:
I can't wait.

crmb:
Hello !

Yes gl_modulate around 1.5 (default seems to be 1 like vid_gamma) kind of solve the problem but it also make wall/floor textures brighter (after a vid_restart) so i am not sure if the picture is right.

if you plan to release a new build :

- I think the game doesnt start if you don't have the openal dll (even with openal set off in the cfg)
- I have tried a bit openal but the stereo was weird for me. If i shoot parallel to a wall, the sound effect will come from the wall side. I guess this is supposed to be reverberation but this is much too pronounced and unnatural.
[...]

PR #103, opened 2015-10-22
Apply gl_overbrightbits to regular meshes.

mackron:
This addresses issue #54. This commit also removes a couple of redundant OpenGL calls.

### DIFF
--- a/src/client/refresh/r_mesh.c
+++ b/src/client/refresh/r_mesh.c
@@ -668,6 +668,21 @@ R_DrawAliasModel(entity_t *e)
 		}
 	}
 
+
+    /* Apply gl_overbrightbits to the mesh. If we don't do this they will appear slightly dimmer relative to
+       walls. Also note that gl_overbrightbits is only applied to walls when gl_ext_mtexcombine is set to 1,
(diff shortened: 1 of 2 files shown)


Yamagi:
This looks very good and is very much appreciated. But one small nit: gl_overbrightbits set 0 leads to black models. I guess that's expected but we should do something about it. Maybe cap gl_overbrightbits at 1?

deuxsonic:
Is there a difference between 1 and 2? I've always used 2 because it's the default. Perhaps make the inclusion of overbright bits on models conditional on the CVAR being greater than 0.

deuxsonic:
Or just check the CVAR and reset it to 1 if it is 0. This is simpler but obviously takes an option away from the user.

deuxsonic:
if (gl_config.mtexcombine) && (gl_overbrightbits > 0) or something like that. Since it's all new code it should be easy to just make the whole block conditional on overbright bits being on, especially since that's what creates the discrepency in the first place.

mackron:
That should be fixed now. I use the same condition that's used with walls for the sake of consistency.

Yamagi:
Nice. Thank you :)

PR #117, opened 2015-11-12
Also ignore .d files as they got generated every time.

Quix0r:
These files can be securely ignored as they are always re-generated.


Yamagi:
Merged, thank you. :)

PR #141, opened 2016-07-01
Fix spelling errors detected by lintian

smcv:
Debian's `lintian` tool checks for common mistakes in source and binary packages. One of its checks is to look for common mis-spellings in the binaries.

Based on a patch by Fabian Greffrath against an earlier version of yquake2.


Yamagi:
Thank you :)

PR #142, opened 2016-07-01
Override __DATE__ from SOURCE_DATE_EPOCH if set

smcv:
For deterministic/reproducible builds (where the same source and
toolchain can be verified to produce the same binary, allowing
maliciously substituted binaries to be detected) it is desirable to
take the software's idea of the build date from the build system;
otherwise, the real-time clock at the time of building affects the
result, making it non-reproducible.

SOURCE_DATE_EPOCH is a distribution-neutral specification for how
to do that. It is meant to be set by meta-build systems such as
dpkg or RPM, using a date/time that is already part of the source code,
for example the date of the latest git commit, the date in
the package's debian/changelog, or the date in the RPM spec file.

See https://reproducible-builds.org/specs/source-date-epoch/ for the
specification of SOURCE_DATE_EPOCH, or https://reproducible-builds.org/
for more information on reproducible builds in general.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,6 +165,13 @@ CFLAGS += -DOSTYPE=\"$(OSTYPE)\" -DARCH=\"$(ARCH)\"
 
 # ----------
 
+# https://reproducible-builds.org/specs/source-date-epoch/
+ifdef SOURCE_DATE_EPOCH
+CFLAGS += -DBUILD_DATE=\"$(shell date --utc --date="@${SOURCE_DATE_EPOCH}" +"%b %_d %Y" | sed -e 's/ /\\ /g')\"
+endif
+
+# ----------
+
 # Systemwide installation
 ifeq ($(WITH_SYSTEMWIDE),yes)
 CFLAGS += -DSYSTEMWIDE
--- a/src/common/header/common.h
+++ b/src/common/header/common.h
@@ -44,6 +44,10 @@
 #error ARCH should be defined by the build system
 #endif
 
+#ifndef BUILD_DATE
+#define BUILD_DATE __DATE__
+#endif
+
 #ifdef _WIN32
  #define CFGDIR "YamagiQ2"
 #else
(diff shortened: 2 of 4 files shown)


Yamagi:
This looks like a nice addition and I'm going to merge it. But before that the Makefile glue must be fiddled into into the CMakeLists.txt to be available in both build system.

smcv:
I'm curious what benefit you're deriving from having the two build systems. I think in every other project I've seen that has two build systems (whether make + CMake,  Automake + CMake or Automake + nmake) the motivation for the second build system has been to support MSVC on Windows, but yquake2 explicitly doesn't?

If one build system is intended to be strongly supported and the other is a second-class citizen, I'd prefer to use the more strongly supported one (whichever that is) in the Debian packages. I don't personally like CMake, but I already maintain another Debian package that relies on it (OpenJK).

Yamagi:
CMake is crap. It may be better then Automake, but that doesn't make it a good build system. The only reason for yq2 having CMake support is that the CMake supporters complained loudly and someone volunteered to do the dirty work... You're right. At this point there's no reason for supporting the SOURCE_DATE_EPOCH  stuff in CMake since most (all?) distributions use the Makefile to build the packages. Even the official Windows binaries are build through the Makefile. If someone really want's to have reproduceable builds with CMake he can provide support for it.

smcv:
Sorry, I realised too late that the patch you've merged has a bug: in the non-reproducible case where `-DBUILD_DATE=...` isn't used, `savegame.c` uses `BUILD_DATE` without a definition.

I'm not completely clear on which modules are meant to be allowed to include which headers. This could either be resolved by having `savegame.c` include `common.h`, or by having `savegame.c` repeat the same `#ifndef/#define` that `common.h` does. Which would you prefer?

Yamagi:
`savegame.c` needs to repeat the defines. `common.h` is shared between the client, renderer and server, but not with the game module.

DanielGibson:
> I'm curious what benefit you're deriving from having the two build systems.

The Makefile is easy to build on Unices/msys and preferred to just produce Binaries.

CMake makes it easy to get the project into IDEs, including CLion which still only supports CMake projects.

smcv:
> `savegame.c` needs to repeat the defines

Regression fixed in #144

PR #143, opened 2016-07-01
Optionally link libopenal at compile time

smcv:
In Linux distributions, having the executable depend on the right
libraries and arrange for them to be installed is straightforward,
and there's a lot of infrastructure for tracking which library
version a particular executable needs, including making sure we have
a version that contains all of the symbols that were used. Loading
libopenal at runtime defeats that infrastructure.

The ability to substitute a different-but-compatible libopenal,
or operate with reduced functionality without libopenal, might
still be desirable for generic/portable binary releases.

The CMake build system already linked the executable to
${OPENAL_LIBRARY} anyway, so it is already a hard dependency in that
build system.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,12 @@ WITH_OGG:=yes
 # installed
 WITH_OPENAL:=yes
 
+# Enables optional runtime loading of OpenAL (dlopen
+# or similar).
+# If set to "no", the library is linked in at compile
+# time in the normal way.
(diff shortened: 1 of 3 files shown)


Yamagi:
Thank you. :)

Merging this has broken the Windows build. A follow up commit fixing it again will be committed a few minutes.

Yamagi:
And there it is.

PR #144, opened 2016-07-05
savegame: ensure that BUILD_DATE is defined

smcv:
The game code does not include common.h, so it needs to redo this
part for builds without SOURCE_DATE_EPOCH, where BUILD_DATE will
not have been passed in from the outside.

Signed-off-by: Simon McVittie <smcv@debian.org>


Yamagi:
Nice.

PR #168, opened 2016-12-23
Arrays bounds issue for OGG playback in OSX

tomgreen66:
Think this is a general issue but found this on OSX.  Stripped change to bare minimum (also had changed in Makefile to build code).  I fixed this by increasing the temporary array to hold 2 integers and a null character.  Might be able to swap the arguments since strcat appends to the first argument.


Yamagi:
Am I right that the change in src/client/menu/menu.c line 1153 is the real bugfix and everything else is just build system stuff for OpenAL on OS X?

tomgreen66:
Sorry - missed removing the sdl2.zip file which I added to my local fork.  The only fix required is in menu.c file.

Yamagi:
I've just pushed a fix for the overflow. The build system stuff shouldn't be necessary since @DanielGibson committed a fix for that problem several days ago. Thank you!

PR #181, opened 2017-04-07
OpenGL3 Renderer, renderers are DLLs again

DanielGibson:
A totally awesome OpenGL3 (3.2 core) renderer, that looks mostly like the old one, which is still available, as they're both separate libs.

Also some small fixes to the GL1 renderer.

Note that the loading of the renderer lib and its public interface is pretty similar to the old one we removed when we moved the renderer into the executable, but it's slightly different and incompatible to old ones.

**Windows binaries:** https://deponie.yamagi.org/quake2/misc/quake2-gl3-test1.zip

**Source Code archive** to build yourself (esp. relevant on non-Windows platforms like Linux and OSX): https://github.com/DanielGibson/yquake2/archive/render_dll.zip

*In the **Video** menu select the **OpenGL 3.2 renderer** by pressing your left/right arrow keys, afterwards apply (select **apply** with down-arrow, then press your Enter key).
The window title will tell you which renderer is currently used.*

Ok, so some notes about what this can do:
* Gamma (`vid_gamma` cvar), intensity (`gl3_intensity` cvar, makes textures look more intense and brighter) and overbrightbits (`gl3_overbrightbits` cvar, lets you adjust darkness of shadows) are applied in the shader and can be changed in realtime (without vid_restart or anything, just change the cvars). Intensity and overbrightbits used to be limited to integral values or even just 1,2,4; now these can be any decimal value, like 1.5
* Dynamic lights are also applied via shader and thus are not limited to the lightmap resolution and look smoother
* Dynamic lights now should work on all brush based entities. In the old renderer they only worked on some doors (and normal walls), but not on rotating things.
* Particles look a bit softer, as they fade out towards their edges
* Lightmaps look the same, but they are blended in the shader (both onto the base surface texture and with each other in case of dynamic lightmaps, like flickering lights, switchable lights etc)
* Screenshots can now be taken in JPG, PNG, BMP (and of course still TGA) format, with both the new and the old renderer. See https://github.com/yquake2/yquake2/pull/181/commits/9462063869dcbe9487e43db9c7311b40ddb0024d for some details
* I tried to keep the code compatible to OpenGL ES3 - but I didn't try that yet. Unfortunately, OpenGL ES2 support (for RPi etc) would be harder, because it lacks UBOs (Uniform Buffer Objects) that I use. Anyway, if anyone wants to port this to mobile platforms or whatever might be using OpenGL ES3, it should be easy.
* The code should be mostly cleaner and more straightforward than the old renderer, because it doesn't have special cases depending on what the hardware supports, it only uses things available on all hardware that supports OpenGL3.2. (To be fair, the worst thing in the GL1 code probably was multitexturing which has already been removed)


Yamagi:
Daniel put a lot of work and even more time into this. This new renderer is likely the best "classic look" renderer ever written for Quake II. To use it an OpenGL 3.2 compliant GPU is requiered. Be aware that some drivers - say 'hello' to the Intel Windows driver - are really crappy and slow. In out tests this new renderer is considerably faster than the old OpenGL 1.4 renderer if a strong GPU and a good driver is used. With slow GPUs and bad drivers it may be slightly slower, but in general stays above ~90 FPS. Anti aliasing is somewhat more as expensive than with OpenGL 1.4, mostly due to the use of fragment shaders. 

Before this can be merged more testing is required. I'll play through the game at least once. In the medium term (after the merge) some smaller features are still missing:

- vid_ref should be renamed to vid_renderer for compatibility with old config files.
- Add the renderer selection back to the video menu.
- Document all cvars. That's long overdue.
- cmake is currently broken.

And some tasks I've forgotten. ;)

iButane:
Sounds promising.  Excited to play through again once a new windows prebuild is available.

Yamagi:
Windows binaries can be found here: https://deponie.yamagi.org/quake2/misc/quake2-gl3-test1.zip 

They're build from this branch plus one missing pull request fixing the item order in the video menu. Since Daniel opened the pull request several bugs were fixed and the look and feel was tweaked to match the removed multitexturing render path of the old OpenGL 1.4 renderer. In fact this new render is much closer to the old renderer with multitexturing than the old renderer with "classic" rendering.

Have fun with testing this. :)

Yamagi:
I've just uploaded a new build. This new binaries have no support for the OpenGL 1.4 renderer to ensure that OpenGL 3.2 is tested.  The URL stays the same: https://deponie.yamagi.org/quake2/misc/quake2-gl3-test1.zip

Yamagi:
Merged, thanks. :)

iButane:
damn, I completely missed (overlooked) the links to the Windows binaries on this thread until now.

Just tried out this new render and it looks amazing! The colors on the console background which I haven't seen in ages! Wow! 
I can even take screenshots again using Steam's overlay (common Nvidia issue)
Gonna try to run through the entire campaign this weekend using this build. So far I haven't noticed any issues.

DanielGibson:
Cool, glad you like it! Have fun and report any issues :)

BTW, you can fine-tune the visual appearance even further with the `gl3_intensity`, `gl3_overbrightbits` and `vid_gamma` CVars (ok, the last one is old and can even be configured via video menu)
The looks of the particles can be tuned with  `gl3_particle_fade_factor` (higher value makes them look less soft/fuzzy) and `gl3_particle_size`

Yamagi:
A new Windows build with all bugfixes since the first test build can be found here: https://deponie.yamagi.org/quake2/misc/quake2-gl3-test2.zip

PR #193, opened 2017-05-07
Add partial doppler shift support for 3D audio sources.

xorw:
In game noticeable when dodging blaster projectiles or missiles :)

See: https://en.wikipedia.org/wiki/Doppler_shift


Yamagi:
This looks (sounds) good and I'm going to merge the change. But before I do that: Is someone out there who can give it a try on OS X? I must admit that I'fe lost track on OpenAL on OS X.

Yamagi:
Merged, thank you :)

@All: if this breaks OS X pull requests are welcome...

PR #196, opened 2017-05-10
Finished doppler shift support (added listener's velocity update)

xorw:


Yamagi:
That sounds good. Nice work. Thank you :)

PR #198, opened 2017-05-15
GL3: Render Model shadows last, refactor gl3_mesh.c

### LINKED ISSUE #194: GL3 Shadow Bug
I found a bug with the GL3 shadows. When a shadow casts on a destructable wall or floor, it will not render the shadow and instead allow you to see through the texture. Heres the video.

https://www.youtube.com/watch?v=Pm-WO-VKhko

DanielGibson:
The model shadows are rendered after all entities are rendered, fixes #194 

refactored gl3_mesh.c to use less global variables (see commit for details)

### DIFF
--- a/src/client/refresh/gl3/gl3_main.c
+++ b/src/client/refresh/gl3/gl3_main.c
@@ -887,6 +887,8 @@ GL3_DrawEntitiesOnList(void)
 		return;
 	}
 
+	GL3_ResetShadowAliasModels();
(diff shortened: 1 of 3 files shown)


Yamagi:
I can confirm that this fixes the issue on AMD, Intel and Nvidia. My old Ivy Bridge GOU even sees a small (1 to 2 percent) performance increase.

PR #207, opened 2017-06-18
Only use /usr/local/opt/openal-soft on Darwin

smcv:
Commit 883781c selected these paths for all OSs, but when linking to
a system copy of OpenAL for Linux distribution binaries we want the
normal OpenAL in /usr.


Yamagi:
You're right. Thank you for the patch. It'll be part of the upcoming 7.01 bugfix release.

PR #213, opened 2017-07-19
Don't use -Wl,--no-undefined on OpenBSD

mulander:
Breaks the build since version 7.00

Results in errors (a lot cropped):
```
===> LD release/baseq2/game.so                                                                                                                                                                                                    
cc build/baseq2/src/common/shared/flash.o build/baseq2/src/common/shared/rand.o build/baseq2/src/common/shared/shared.o build/baseq2/src/game/g_ai.o build/baseq2/src/game/g_chase.o build/baseq2/src/game/g_cmds.o build/baseq2/s
rc/game/g_combat.o build/baseq2/src/game/g_func.o build/baseq2/src/game/g_items.o build/baseq2/src/game/g_main.o build/baseq2/src/game/g_misc.o build/baseq2/src/game/g_monster.o build/baseq2/src/game/g_phys.o build/baseq2/src/
game/g_spawn.o build/baseq2/src/game/g_svcmds.o build/baseq2/src/game/g_target.o build/baseq2/src/game/g_trigger.o build/baseq2/src/game/g_turret.o build/baseq2/src/game/g_utils.o build/baseq2/src/game/g_weapon.o build/baseq2/
src/game/monster/berserker/berserker.o build/baseq2/src/game/monster/boss2/boss2.o build/baseq2/src/game/monster/boss3/boss3.o build/baseq2/src/game/monster/boss3/boss31.o build/baseq2/src/game/monster/boss3/boss32.o build/bas
eq2/src/game/monster/brain/brain.o build/baseq2/src/game/monster/chick/chick.o build/baseq2/src/game/monster/flipper/flipper.o build/baseq2/src/game/monster/float/float.o build/baseq2/src/game/monster/flyer/flyer.o build/baseq
2/src/game/monster/gladiator/gladiator.o build/baseq2/src/game/monster/gunner/gunner.o build/baseq2/src/game/monster/hover/hover.o build/baseq2/src/game/monster/infantry/infantry.o build/baseq2/src/game/monster/insane/insane.o
 build/baseq2/src/game/monster/medic/medic.o build/baseq2/src/game/monster/misc/move.o build/baseq2/src/game/monster/mutant/mutant.o build/baseq2/src/game/monster/parasite/parasite.o build/baseq2/src/game/monster/soldier/soldi
er.o build/baseq2/src/game/monster/supertank/supertank.o build/baseq2/src/game/monster/tank/tank.o build/baseq2/src/game/player/client.o build/baseq2/src/game/player/hud.o build/baseq2/src/game/player/trail.o build/baseq2/src/
game/player/view.o build/baseq2/src/game/player/weapon.o build/baseq2/src/game/savegame/savegame.o -L/usr/local/lib -lm -fvisibility=hidden -Wl,--no-undefined -shared -o release/baseq2/game.so
build/baseq2/src/common/shared/shared.o: In function `Q_fabs':                                                   
src/common/shared/shared.c:299: undefined reference to `__stack_smash_handler'                                   
build/baseq2/src/common/shared/shared.o: In function `BoxOnPlaneSide2':                                          
src/common/shared/shared.c:364: undefined reference to `__stack_smash_handler'                                   
build/baseq2/src/common/shared/shared.o: In function `FloatSwap':                                                
src/common/shared/shared.c:858: undefined reference to `__stack_smash_handler'                                   
build/baseq2/src/common/shared/shared.o: In function `tolower':                                                  
/usr/include/ctype.h:150: undefined reference to `_tolower_tab_'                                                 
build/baseq2/src/common/shared/shared.o: In function `Info_Validate':                                            
src/common/shared/shared.c:1292: undefined reference to `strchr'                                                 
src/common/shared/shared.c:1297: undefined reference to `strchr'                                                 
build/baseq2/src/common/shared/shared.o: In function `Swap_Init':                                                
src/common/shared/shared.c:895: undefined reference to `__stack_smash_handler'                                   
build/baseq2/src/common/shared/shared.o: In function `Info_ValueForKey':                                              
src/common/shared/shared.c:1207: undefined reference to `strcmp'                                                 
src/common/shared/shared.c:1219: undefined reference to `__stack_smash_handler'                                  
build/baseq2/src/common/shared/shared.o: In function `Info_RemoveKey':                                           
src/common/shared/shared.c:1229: undefined reference to `strchr'                                                 
src/common/shared/shared.c:1272: undefined reference to `strcmp'                                                 
src/common/shared/shared.c:1283: undefined reference to `__stack_smash_handler'                                  
src/common/shared/shared.c:1274: undefined reference to `strlen'                                                 
src/common/shared/shared.c:1274: undefined reference to `memmove'     
build/baseq2/src/common/shared/shared.o: In function `Com_sprintf':                                              
src/common/shared/shared.c:1088: undefined reference to `vsnprintf'                                              
src/common/shared/shared.c:1095: undefined reference to `__stack_smash_handler'
build/baseq2/src/common/shared/shared.o: In function `Info_SetValueForKey':                                      
src/common/shared/shared.c:1312: undefined reference to `strchr'                                                 
src/common/shared/shared.c:1312: undefined reference to `strchr'           
```
the diff results in a passing build

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -252,8 +252,8 @@ endif
 CFLAGS += -fvisibility=hidden
 LDFLAGS += -fvisibility=hidden
 
-ifneq ($(YQ2_OSTYPE), Darwin)
-# for some reason the OSX linker doesn't support this
+ifneq ($(YQ2_OSTYPE), $(filter $(YQ2_OSTYPE), Darwin, OpenBSD))
+# for some reason the OSX & OpenBSD linker doesn't support this
 LDFLAGS += -Wl,--no-undefined
 endif
 


mulander:
Needs additional changes as:

```
----- refresher initialization -----
LoadLibrary(ref_gl1.so)
Sys_LoadLibrary failed: File not found
Shutting down OpenAL.
Cmd_RemoveCommand: force_centerview not added
Cmd_RemoveCommand: +mlook not added
Cmd_RemoveCommand: -mlook not added
Shutting down input.
recursive shutdown
Error: Loading ref_gl1.so as renderer lib failed!
```

mulander:
I would appreciate pointers on what changed with ref_gl1 / ref_gl3 since v6.00 that worked prefectly on OpenBSD.

The pull request here *does* allow a clean compile but the resulting binary is crashing with the error mentioned in the previous commit.

DanielGibson:
Hmm it could be that Sys_GetBinaryDir() (see src/backends/generic/misc.c) doesn't work on OpenBSD.
It's supposed to return the directory the quake2 executable is in and used in VID_LoadRefresh() (src/backends/generic/vid.c) to load the renderers (which are supposed to lie next to the executable)

DanielGibson:
OTOH, when Sys_GetBinaryDir() fails it should just return an empty string, so `VID_LoadRefresh()` should basically call (through Sys_LoadLibrary()) `dlopen("ref_gl1.so", RTLD_LAZY);` which should at least work if $PWD is the directory the quake2 executable and ref_gl1.so are in.

mulander:
There is no way to obtain the currently running executable path on OpenBSD. Additionally we install the binary into `/usr/local/bin/quake2` and all game data and .so files to `/usr/local/share/yquake2/`.

This has been working fine so far when compiled using:
`MAKE_FLAGS = config WITH_SYSTEMWIDE=yes WITH_SYSTEMDIR=${PREFIX}/share/${N}`

systemdir resolves to `/usr/local/share/yquake2/`

I did try running it by changing the current working directory to `/usr/local/share/yquake2/` but face the same error.

I will add a gist here in a second with full output from a test run.

DanielGibson:
can you try running the game without installing it?
from the release/ directory where the binaries are located after compiling

you can set the directory the data is in with `./quake2 +set basedir /path/to/quake2/`

for Linux we suggest to keep all binaries in the same directory (game.so in the baseq2/ subdirectory next to the executable of course) and then symlinking the executable to /usr/bin/ or whatever is on your $PATH.
For OpenBSD it won't work if it's not possible to get the executable path, so you may need a shellscript in /usr/bin/ that switches to the directory the binaries are in first.

mulander:
It's the same when run directly from the release directory: https://gist.github.com/mulander/bf29684432ef8cc9c360fdf5786d45b1

Here is the build configuration:
```
Build configuration                                                                                                                                                                                                               
============================                                                                                                                                                                                                      
WITH_CDA = no                                                                                                                                                                                                                     WITH_OPENAL = yes                                                                                                                                                                                                                 
WITH_SDL2 = yes                                                                                                                                                                                                                   
WITH_X11GAMMA = no                                                                                                                                                                                                                
WITH_ZIP = yes                                                                                                                                                                                                                    WITH_SYSTEMWIDE = yes                                                                                                                                                                                                             
WITH_SYSTEMDIR = /usr/local/share/yquake2                                                                                                                                                                                         
============================                                                                                                                                                                                                      
                                                                                                                                                                                                                                  WARNING: CDA disabled because SDL2 doesn't support it! 
```

DanielGibson:
Weird.. can you debug this somehow?
maybe step through the Sys_LoadLibrary() call in VID_LoadRefresh(), check if it path is indeed just "ref_gl1.so"?
and maybe around there print the result of getcwd() to make sure the game still is in the right directory, just in case we change the directory somewhere and I forgot about it

Thanks :)

mulander:
added a print before Sys_LoadLibrary

```c
Com_Printf("reflib_path='%s'\n", reflib_path);                                                                                                                                                                          
Com_Printf("cwd='%s'\n", getcwd(NULL, 256))
```

reflib_path='ref_gl1.so'
cwd='/usr/local/pobj/yquake2-7.01/quake2-7.01/release'

will step through Sys_LoadLibrary and see where it fails. Thanks for the pointers so far.

DanielGibson:
this is weird.. maybe openbsd's dlopen() doesn't check $PWD when given just a library name (i.e. no path)?

a silly idea, not sure if it'll help at all: in VID_LoadRefresh() replace
`snprintf(reflib_path, sizeof(reflib_path), "%s%s", Sys_GetBinaryDir(), reflib_name);`
with
`snprintf(reflib_path, sizeof(reflib_path), "./%s", reflib_name);`

mulander:
`Sys_LoadLibrary <ref_gl1.so> failed: File not found`

is returned from:
```c
 11 void *                                                                                                                                                                                                                        
 10 Sys_LoadLibrary(const char *path, const char *sym, void **handle)                                                                                                                                                             
  9 {                                                                                                                                                                                                                             
  8         void *module, *entry;                                                                                                                                                                                                 
  7                                                                                                                                                                                                                               
  6         *handle = NULL;                                                                                                                                                                                                       
  5                                                                                                                                                                                                                               
  4         module = dlopen(path, RTLD_LAZY);                                                                                                                                                                                     
  3                                                                                                                                                                                                                               
  2         if (!module)                                                                                                                                                                                                          
  1         {                                                                                                                                                                                                                     
517                 Com_Printf("%s <%s> failed: %s\n", __func__, path, dlerror());                                                                                                                                                
  1                 return NULL;                                                                                                                                                                                                  
  2         } 
```

note I changed the output to make sure what's used as the path.

Regarding dlopen on OpenBSD the [man page](https://man.openbsd.org/dlopen) states.
> The path parameter can be specified as either an absolute pathname to a shared library or just the name of the shared library itself. When an absolute pathname is specified, only the path provided will be searched for the shared library. When just a shared library is specified, the same paths will be searched that are used for “intrinsic” shared library searches.

the string "ref_gl1.so" is not an absolute path hence it's not searched for in the current directory. I assume it searches /usr/lib, /usr/local/lib and other places where system wide libs are expected.

Just to test this assumption. I copied the ref_gl1.so and ref_gl3.so to /usr/local/lib and have the game running.

DanielGibson:
the manpage doesn't mention it, but do relative paths maybe work ("./ref_gl1.so")?

mulander:
Just for a quick test I patched:
```diff
-       snprintf(reflib_path, sizeof(reflib_path), "%s%s", Sys_GetBinaryDir(), reflib_name);
+       snprintf(reflib_path, sizeof(reflib_path), "./%s%s", Sys_GetBinaryDir(), reflib_name);
 ```

and removed the system wide binaries form `/usr/local/lib`

the game loaded, so the relative path works.

DanielGibson:
great, thanks for testing!

maybe we'll just make Sys_BinaryDir() return "./" then if it can't figure out the actual binary dir?

(why does OpenBSD not have a way to get the binary dir? :-/)

mulander:
There already is a `WITH_SYSTEMDIR` which allowed the binaries to be separated from the game folder. Perhaps defaulting to that would be a cleaner solution as it's an absolute path and can be defined externally by package managers? I am also just considering installing the .so files in /usr/local but waiting for feedback on that.

DanielGibson:
Well, originally the WITH_SYSTEMDIR was meant for game data..
but it's also used to load the game.so (because for mods it kinda is part of the game data)..
and before we put the renderer back into a lib, quake2 was pretty much self-contained (except for game.so)..
and then we added that rpath hack with SYSTEMDIR/lib so one could more easily use a custom libsdl2 (back when it wasn't readily available in most distros)..
and some packages used the systemdir for both game data and binaries..

It's a bit of a mess and I'm not sure yet how best to resolve it without breaking too much stuff.

One reason we want to have all the binaries (esp. quake2 executable and render libs) in the same path (vs looking for render libs in all the places we look for game data) is to make sure people don't just update one part of their q2 installation and report weird bugs.

mulander:
I appreciate the context, I'm looking for peer feedback from OpenBSD people on how we would like it resolved, will get back to you as soon as I have some useful suggestions.

Should I re-open this as a regular issue in the tracker so we can further discuss a solution? The linker change in this pull request does break compilation when missing so I would like to have that upstreamed.

mulander:
> maybe we'll just make Sys_BinaryDir() return "./" then if it can't figure out the actual binary dir?

I'm leaning towards agreeing with this proposal. I can repackage yquake2 to place binaries the way you suggest (all in the same directory with game data and .so files) then providing a shell wrapper that will change the current directory to the same one as the binary. I do need however a way to make sure the game will look for everything in `/usr/local/share/yquake2/`. This seems covered by `WITH_SYSTEMDIR` but the `Sys_BinaryDir` change will still be needed for the game to load the .so from the current working directory.

Let me know what you decide and I will act accordingly on our side.

DanielGibson:
ok, this is merged.

the SYSTEMDIR problem may need some more discussion.. will probably open an issue for that soon

thanks a lot for the patch and for helping to figure the other problem out! :)

DanielGibson:
#214 discusses the WITH_SYSTEMWIDE problems

DanielGibson:
We now have documentation for package/port maintainers: https://github.com/yquake2/yquake2/blob/master/stuff/packaging.md

PR #215, opened 2017-07-25
Add joystick support

0lvin:


DanielGibson:
Hmm, I think the old joystick code (and thus also this code) is not that useful today, where the most popular joystick-like devices (especially for FPS) are XBox/Playstation-style gamepads ("controllers") with two analog sticks, where one stick is used for forward/backward + strafe left/right, and the other for looking up/down and turning left/right.

A further problem with joystick-like devices is that the axis numbers are not consistent between devices, so hardcoded assumptions about the meaning of axis0 and axis1 (even worse: every further axis) will cause trouble.

So we'd need:
* support for at least 6 axes (4 axes for the two analog sticks, and I think that sometimes the controller's DPad is represented as two axes - which axis is which varies per device)
* configurable action per axis
  * forward/backward
  * strafe left/right
  * look up/down
  * turn left/right
* maybe make all axis invertible?

Furthermore I'm not sure if lookstrafe and freelook should be taken into account here, or can just be ignored (does anyone want to use them with joysticks/gamepads/...?)

xDShot:
Joysticks support is what I was waiting for a long time! Could you please implement axis support and loading mappings config from gamecontrollerdb.txt (https://github.com/gabomdq/SDL_GameControllerDB)?

DanielGibson:
SDL2's gamecontroller API allows loading those mappings and then one can either use the controller events or https://wiki.libsdl.org/SDL_GameControllerGetBindForAxis etc to do the logical mapping yourself and continue to use joystick events.

But IMHO that is not overly important (as long as action for axis can be configured), more of a convenience feature.

xDShot:
@0lvin I've managed to compile yamagi quake 2 with this pull. My SDL mapping line works, sticks work, and I could manage to bind buttons in menu. Thank you!

Here I'd suggest some improvements:
* Make buttons work in menu and bind actions on buttons: DPAD to navigate, A button to confirm, B button to move to previous menu, START to open/close menu, etc.;
* Here is a little issue: when I put left stick in diagonal, it feels like I move slower. Maybe there is something related to normalizing diagonal inputs, I dunno...;
* Maybe make console variable for SDL mapping instead of environment variable?

Overall, it's great. Thanks again! :)

DanielGibson:
now this looks a lot better!

however, some things that could be improved:

* I think for Quake2 (and FPS games in general), using the  triggers as axes for movement/looking is not that useful - I guess most people will want to use them like normal buttons, probably for shooting
* related: On some (many?) controllers (e.g. PS2-like), the triggers *are* buttons and thus will probably create two events: one `SDL_CONTROLLER_AXIS_TRIGGERLEFT` (or RIGHT) event and one `SDL_JOYBUTTONDOWN` event
* some kind of calibration or deadzone support (maybe as "deadzone start" and "deadzone end" values?) would be useful probably.. but might need UI (menu) support to be usable, which might be quite painful in Quake2. Maybe people can use their operating system for calibration instead..

I haven't actually tested the code yet, this are just some thoughts after reading the code

0lvin:
* I have DualShock like joystic with 4 axis, triggers generated only "button" events. So if player like to map trigger to attack action, by defaults code for now export all buttons without relation to kind of such buttons. Code see all buttons as joystick buttons, so we can use more than ten buttons as we have with [gamecontrollers](https://wiki.libsdl.org/SDL_GameControllerButton). Also triggers axis is not mapped to move actions by default for now on code. I have added logic for minimal threshold/deadzones for axis.  
* User can use dpad(hat buttons) for navigation in menu.
* Also i have added minimal code for haptic feedback(disabled by defaults), current code can generate only feedback to actions for navigation in menu and click to attack button.

DanielGibson:
With Xbox360 controllers, the triggers are real axes and only generate axis events, no button events, so with those the fallback to SDL_JOYBUTTONDOWN doesn't work.

I'd suggest adding `K_TRIG_LEFT? and `K_TRIG_RIGHT` key definitions.  
Then, when handling `SDL_CONTROLLER_AXIS_TRIGGERLEFT` you check whether `joy_axis_triggerleft` is either "none" or "", and if it is you create a button down event (`Key_Event(K_TRIG_LEFT, down, true);`), otherwise you do whatever you're already doing for the axis.

Only problem is that you'll also need to generate a key up event once the trigger is not pressed anymore.

0lvin:
Ok, will create such code. 

Also are you intrested in haptic feedback for gun's and explosions? (I don't have solution for it, but tink about it).

DanielGibson:
Haptic feedback sounds cool! (As long as people can switch it off if they don't like it)
I have no idea how hard it'd be to achieve that, or if it's even possible without changing (=> breaking) the API between game.dll and client or the network protocol (I think  we'd like to avoid both things).

But please try it :)
I'd suggest doing that after the "normal" joystick/controller code is done, in a separate branch - so we could merge this without waiting for haptic feedback to be done (I can imagine it needs quite some testing and maybe discussion on improvements etc).

0lvin:
Do you can check what values generated by trigger axis? I think for now use abs(axis.value) > 16384 ([half of full range](https://wiki.libsdl.org/SDL_JoyAxisEvent)) as condition.

DanielGibson:
The trigger axis has value 0 when not pressed and up to 32767 when pressed.
Luckily (no idea if we can rely on this for every platform and device), I even get an event with value 0 when not pressing anymore (=> you could use that case for button up)

I think you could already register the event at <16.000, around 5000 should be enough I think? (Maybe you could just use the configurable threshold for that?)

0lvin:
User can change value in condition by set bigger value in joy_axis_triggerleft_threshold and joy_axis_triggerright_threshold. For now used such condition:
```
qboolean new_right_trigger = abs(axis_value) > (32767 / 4);
```

0lvin:
@DanielGibson @Yamagi Could you please review changes?

0lvin:
I have made #227 - it contains initial code for check haptic feedback, if you like to test - you can try this branch.

DanielGibson:
Sorry we haven't gotten around to properly review and merge this yet, Yamagi and me are both really busy at the moment.
I hope I'll find some time at the end of next week, but I can't promise it - but this is not forgotten :)

DanielGibson:
Ok, we tested it a bit, some notes:

* please make sure the game still builds with SDL1.2 (even if joystick support doesn't work there)
* I had to set `joy_sidesensitivity -1` because strafing left and right was wrong on my (xbox360) gamepad - I guess this should fixed in code so it works as expected (move left analog stick to the left -> strafe left) by default
* I had to set jaw and pitch sensitivity way down (~`0.3`) to be able to aim => maybe scale it down a bit by default?
* when navigating in the menu there is no way to go back with the controller (like with `Esc` on a keyboard)
* maximum movement speed seems to be walking speed - should be running speed (at least if "always run" is on)
* when navigating the menu with the hat the console is spammed with "HAT_* is unbound, hit F4 to set" messages - that's uncool (especially in the menu where the hat is actually used)
* when pressing other unbound keys (even in game), like `joy8`, no such messages are printed (here it might be useful to bind those keys via console)
* haptic feedback feels inconsistent - shooting (at enemies) rumbles more than being shot, if a parasite (dog-like thing) attacks you, there is no rumbling at all. running (only possible via keyboard, see above) also causes rumbling, it probably shouldn't

All in all it seems pretty good, thanks a lot so far and sorry for the late reply!

(Yamagi will comment on the code later)

Yamagi:
Okay, now my part:

* vid.c line 104: We can and and should add more resolutions to Quake II. For example 2560x1440 is missing too. We can merge you code like it is but should add the other missing resultions before the next release. Or you could add them.
* input.c line 625: If more than a few lines are taken from ioquake3 we should add their author the license header of that file.
* input.c: Like @DanielGibson already said the old SDL 1.2 is broken, several #ifdef are missing. It may be  a good idea to move the haptic stuff to a separate file and #ifdef the whole file out.
* Key_GetKey() is now unused and should be removed.
* stb_img.h is contributed code. Don't change it even if the compiler warnings are annoying.
* menu.c: Using the empty lines for the joystick stuff is rather confusing since the options are scattered around. What about replacing lookspring and lookstrafe with the joystick stuff if it's enabled?
* In general axis should be treated as axis and buttons as buttons. Treating axis as buttons makes navigation with them hard to impossible.

Yamagi:
And don't register the cvars if the joystick subsystem isn't loaded. That's super confusing and will lead to bug reports. :)

0lvin:
Thank you.

I have fixed/reverted such:
* Fix build with SDL 1.2
* Move video modes
* Oposite direction on joy_sidesensitivity
* Add notes about code reuse from ioquake3
* Revert changes in stb_img.h

Not fixed/changed:

@DanielGibson:
> "I had to set jaw and pitch sensitivity way down (~0.3) to be able to aim => maybe scale it down a bit by default?"
>maximum movement speed seems to be walking speed - should be running speed (at least if "always run" is on)

It can be related to controller/joystick - i don't have any universal solutions for it. I have same movement speed with my joystick as with buttons, or like it. Try to change sensitivity to bigger value. 

> "haptic feedback feels inconsistent - shooting (at enemies) rumbles more than being shot, if a parasite (dog-like thing) attacks you, there is no rumbling at all. running (only possible via keyboard, see above) also causes rumbling, it probably shouldn't" 

Also can be related to joystick - i have some noname joystick - and i am not sure what will be correct feedback. I can set other value to magnitude and time period for some weapons, if you like. Code also generated feedback for steep sound - so running also have some feedback.   

> "when navigating the menu with the hat the console is spammed with "HAT_* is unbound, hit F4 to set" messages - that's uncool (especially in the menu where the hat is actually used)
when pressing other unbound keys (even in game), like joy8, no such messages are printed (here it might be useful to bind those keys via console)"
> "when pressing other unbound keys (even in game), like joy8, no such messages are printed (here it might be useful to bind those keys via console)"

I will try to resolve it in next version.

> when navigating in the menu there is no way to go back with the controller (like with Esc on a keyboard)

Maybe add additional line in each menu like "go back" ?

@Yamagi:
>  menu.c: Using the empty lines for the joystick stuff is rather confusing since the options are scattered around. What about replacing lookspring and lookstrafe with the joystick stuff if it's enabled?

Maybe someone like to use lookspring / lookstrafe. If you like i can replace this lines with my "joystick lines", or delete "in_joystick" at all and move "haptic" to end of menu. 

> In general axis should be treated as axis and buttons as buttons. Treating axis as buttons makes navigation with them hard to impossible.

Do you mean delete "K_AXIS0" fake buttons?

0lvin:
@DanielGibson @Yamagi What do you thick about update #235? It has only one warning :-)
```
===> CC src/client/refresh/files/stb.c
In file included from src/client/refresh/files/stb.c:42:0:
src/client/refresh/files/stb_image.h:915:14: warning: ‘stbi__malloc_mad4’ defined but not used [-Wunused-function]
 static void *stbi__malloc_mad4(int a, int b, int c, int d, int add)
              ^~~~~~~~~~~~~~~~~
```

0lvin:
I have fixed/reverted such:
* Fix build with SDL 1.2
* Add more video modes
* Fix to opposite direction for joy_sidesensitivity
* Add notes about code reuse from ioquake3
* Fixed "bind" messages
* Changed megnitude level for pain
* Removed fake events for AXIS*_{DOWN,UP}
* Changed logic in menu -> If detected hapric capabilities -> lookspring menu item will be replaced by haptic magnitude item. 

@DanielGibson @Yamagi 'lookspring' looks as unused in code base. Does it make sense to remove this cvar?

Yamagi:
Hey, that looks good. I just had a look and found only two small problems:
* In input.c line 1193 `Com_Printf("------------------------------------\n\n");` should be moved to the bottom of that function. Otherwise some output is not in the "input block".
* In the options menu the internal order of the elements doesn't match the visible order. That can be seen when scrolling the menu from top to bottom. That's likely caused by adding the menu elements in a wrong order. I would suggest to remove the blank line under "invert mouse" and move everything below it one line up. And yes, the menu system is crap. ;)

Otherwise I'm happy. Just one questions: Is there a way to bind axis through the menu? I'm not even sure that the bind system is able to do such thing but it would be nice.

 And the 'lookspring' cvar is indeed unused. Go ahead and remove it.

0lvin:
@Yamagi thank you, fixed.

> Just one questions: Is there a way to bind axis through the menu?

Off the top of my head - we can do it in such way: 
* create separate menu where we will set some flag "bind axis" before enter and unset after exit of such menu (It will fully hide any additional fake events ralated to axis movement).
* inside of such menu we will have rows with "direction options" like:
  * sidemove `<leftx>`
  * updown `<righty>`
  * ......
* in inputs.c  if "bindaxis" is true - move for axis will produce key_event. 

So it will be good candidate for improvements in next PR :-)

Yamagi:
Yes, let's make a separate PR for that or even leave it as it is. I think we can merge this now if @DanielGibson is okay with it.

Yamagi:
Merged. Thank you! :)

Pickle:
I tried the latest commit out last night. Things work well for the most
part. I had one issue with my joystick (which has 4 shoulder buttons). The
bottom row of shoulder buttons mapped as shoulder triggers left and right.
When in game using these buttons the unmapped button/key message pops up
for the joy#.
I suspect this might be a problem with SDL. but maybe its worth mentioning.

0lvin:
It is expected behavior for now, code use initial [joystick button events](https://github.com/yquake2/yquake2/blob/e7199f5c416bbeb4b30ae71aaa93d7616fe22261/src/backends/sdl/input.c#L634) and [controller axis events](https://github.com/yquake2/yquake2/blob/e7199f5c416bbeb4b30ae71aaa93d7616fe22261/src/backends/sdl/input.c#L526). SDL hide button events for triggers, so when user clicks trigger button - code will receive only trigger axis event. So joysticks(not controllers) without trigger axis will never receive any events(trigger button).

So use joystick events will be more flexible because it provide up to [32 buttons](https://wiki.libsdl.org/SDL_JoyButtonEvent) and never hide events. Controller can have up to [14 buttons](https://wiki.libsdl.org/SDL_GameControllerButton). Also controller can have only [6 axis](https://wiki.libsdl.org/SDL_GameControllerAxis) but if user has joystick with bigger amount of axis user can remap it by SDL_GAMECONTROLLERCONFIG.

As temporary solution set 'joy_axis_trigger{left,right}' to 'none' and bind 'joy#' to actions for triggers.

Pickle:
thanks that worked

PR #223, opened 2017-08-02
Fix opengl linking for the renderers.

fucknohtml:
The variable yquake2RendererLinkerFlags is used to store the linker flags for opengl but the empty variable yquake2OpenGLLinkerFlags is used when linking libraries to the renderer.

Also there's no need to link the opengl libs to the opengl3 renderer when the glad extension loader is being used.


Yamagi:
You're right. Thank you :)

PR #227, opened 2017-08-07
initial haptic feedback (rfc, not final version)

0lvin:
All gun's have feedback, sometime such feedback is fully random. If you like to improve/change some feedback change params to IN_Haptic_Effect_Init. 

For check set joy_haptic_level == 1

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,6 @@ the original executables, documentation and so on. They aren't needed anymore:
 - ref_gl.dll
 - ref_soft.dll
 - baseq2/gamex86.dll
-- baseq2/maps.lst
 - ctf/ctf2.ico
 - ctf/gamex86.dll
 - ctf/readme.txt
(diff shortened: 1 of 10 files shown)


DanielGibson:
this doesn't rumble when the player is it, does it?
maybe you could use `player_state_t::kick_angles`? I think they're related to the player being shot at (or affected by a nearby explosion etc)

0lvin:
I have found that all sounds related to player have empty fixed_origin(S_StartSound). 
I have close PR for now and merge changes to #215. Joystick and haptic will be disabled by default but can be enabled from menu.

0lvin:
@DanielGibson - i will search way to get on client original kick_angless, for now - i see only entity number.
```
S_StartSound(NULL, cl.playernum + 1, 0, sfx, 1.0, 1.0, 0);
```

PR #245, opened 2017-10-07
Add support for use controller back button as ESC

0lvin:
As example first button(...,back:b1,...) in SDL_GAMECONTROLLERCONFIG

Quit from game by press any button on joystick in "quit dialog".


Yamagi:
Merged, thank you :)

PR #248, opened 2017-10-17
Makefile osx fixes

tomgreen66:
Due to a couple of issues on OSX the Makefile can be improved (along with bug fix in code that highlighted due to Makefile issues).
1. The filter seems to include an extra comma that causes an error in Make.
1. Homebrew does not install openal-soft completely due to OpenAL included in OSX.  Either set LD_LIBRARY_PATH to point to it or we can compile with rpath to point to it (and we assume location anyway).  
1. Due to issues with setting LD_LIBRARY_PATH my first few tests failed to load the OpenAL library and causes an error when Ogg_Pause tries to stop the stream at end of first level in lift.  Make sure we test for OpenAL.


DanielGibson:
Looks good on first sight, thank you! :)
Can you squash it into one commit? (or maybe two, once for the makefile changes and one for the change in ogg.c)

I'll have to test this and then we should get this and the changes for Sys_Microseconds() from the osx_clock branch into master.

tomgreen66:
Will do what you suggest. Spotted the osx_clock branch (similar issue in a Doom3 port), see: RobertBeckebans/RBDOOM-3-BFG#375.  

Final issue I cant seem to trace is some garbage in the shutting down output.

```
-===================================-

 Installation
models/deadbods/duShutting down Ogg Vorbis.
Shutting down OpenAL.
Shutting down input.
```
but when adding fflush before and after fputs in `Sys_ConsoleOutput` it seems to correctly print the lines:
```
-===================================-

 Installation

----------- shutting down ----------
==== ShutdownGame ====
Shutting down Ogg Vorbis.
Shutting down OpenAL.
Shutting down input.
```

Anyone else seeing it?

tomgreen66:
Closing due to recommendation.  Now in #249.

PR #249, opened 2017-10-20
A few fixes for OSX

tomgreen66:
As described in original pull request #248 (where commit was too cumbersome). Due to a couple of issues on OSX the Makefile can be improved (along with bug fix in code that highlighted due to Makefile issues).
1. The filter seems to include an extra comma that causes an error in Make.
1. Homebrew does not install openal-soft completely due to OpenAL included in OSX.  Either set LD_LIBRARY_PATH to point to it or we can compile with rpath to point to it (and we assume location anyway).  
1. Due to issues with setting LD_LIBRARY_PATH my first few tests failed to load the OpenAL library and causes an error when Ogg_Pause tries to stop the stream at end of first level in lift.  Make sure we test for OpenAL.

Now just need to find out why the shutting down text seems corrupted as described in https://github.com/yquake2/yquake2/pull/248#issuecomment-338304417

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -277,7 +277,7 @@ endif
 CFLAGS += -fvisibility=hidden
 LDFLAGS += -fvisibility=hidden
 
-ifneq ($(YQ2_OSTYPE), $(filter $(YQ2_OSTYPE), Darwin, OpenBSD))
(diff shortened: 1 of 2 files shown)


Yamagi:
We've just merged this by hand since it conflicted with the OS X timing fixes branch. Thank you :)

PR #256, opened 2017-11-16
Soft render from https://icculus.org/quake2/

0lvin:


0lvin:
@Yamagi @DanielGibson What do you think about restore software render? 

I have not cleaned up code yet, so it is not ready to merge or review. Render based on code from https://icculus.org/quake2/ with deleted all asm code.

I have checked code on i386 and x86_64 debian with SDL1.2 and SDL2.0.

Yamagi:
Cool, that looks good. At least on a very first sight. It even works on the Raspberry PI 3. :) Of course, some questions arise:

* What to do whith the cvars? Currently all render independent cvars are prefixed with gl_ and render specific cvars with gl1_ or gl3_. Moving the softrenderer into the gl_ namespace would be easiest but feels wrong. Staying with r_ complicates things. Moving the GL renderers back to r_ breaks existing configs...
* It would be really nice if at least overbrightbits could be implemented into the softrenderer. It shouldn't be too hard, maybe the GL1 code could be used as a starting point. Some other features added to the GL renderers are support for tga textures and jpeg screenshots.

And of course the big question if we're able to maintain another renderer over the years to come.

0lvin:
Thank you. 

@Yamagi I have changed code to use sw_mode for soft renader, and also added logic for screenshots. 

And have such known issues:
* Code is used palette based colors so overbright will take little more time. 
* Also code does not have implementation for [scale](https://github.com/0lvin/yquake2/blob/cc2b8aaa73d058c96a477914c2ee36f093c6e8ac/src/client/refresh/soft/r_draw.c#L75).
* And code segfault with 2K resolution. 

>> And of course the big question if we're able to maintain another renderer over the years to come.

I think we can disable soft renger by default and user can enable when want.

iButane:
So with this software render you're working on, will we finally be able to play on higher resolutions without crashing every time we jump in water? If I remember correctly, any resolution above 1152x864 will crash the game every time you submerge in water.

0lvin:
@iButane I have tried 1366x768 - looks as works without any crashes in softrender, but above 2k resolutions is crashed for sure.

maraakate:
The game will crash at resolutions above 1280x1024 when jumping into water with ref_soft.  I believe, but am not 100% on this, that there are some sizes that must be changed and that sezero may have solved this issue.  I will forward him this conversation.

maraakate:
I remember that MAXHEIGHT and MAXWIDTH must be changed, but if I recall from previous conversations with sezero there is still a fundamental limit of 2084.

sezero:
About the crashes:  MAXWIDTH and MAXHEIGHT are important, and care
should be taken when changing (bumping) them.

- be careful if you ever want to change MAXWIDTH: 12.20 fixed
  point math used in R_ScanEdges() overflows at width 2048.

- the screen warp code expects hardcoded 1600x1200 max resolution
  instead of using the macros.


I'm attaching two patches here (you can `git am` them):

- first one replaces some (long) casts on pointers with intptr_t,
  which is needed for WIN64, because WIN64 is LLP64 and not LP64.

- second one changes MAXHEIGHT back to 1200 (because I did a lazy
  copy/paste), adds back MAXWIDTH as 1600, and adjusts the screen
  warp code to use those macros as it was done in q1, so it would
  not crash in water so long as your resolution is a legitimate
  one: you should take care to filter your resolutions accordingly.

Additional note about horplus fov and softrender:  the changes
need implementing in r_misc.c:R_ViewChanged() and R_SetupFrame(),
and ref_soft needs notifying about horplus being in effect or
not.  See the changes in q2dos port I did for this; the relevant
commits should be these four:
https://bitbucket.org/neozeed/q2dos/commits/30b42a7c563cd51a138af739c7dd4d1563a61f28
https://bitbucket.org/neozeed/q2dos/commits/aba4a2dede0011839ffc04bb4d7f71e1422b8532
https://bitbucket.org/neozeed/q2dos/commits/3b0711a096a31da2e0102eb18e4ea5390cc85412
https://bitbucket.org/neozeed/q2dos/commits/928c763c67d34791b68670d0ec3cf87b21010d53

[0001-replace-several-long-casts.txt](https://github.com/yquake2/yquake2/files/1502380/0001-replace-several-long-casts.txt)
[0002-sintable-size-warp-buffer-size.txt](https://github.com/yquake2/yquake2/files/1502381/0002-sintable-size-warp-buffer-size.txt)

0lvin:
@sezero  Thank you. Do you know exact line/state where value overflows? I like to support 2048+ resolutiions. 

I am trying to get rid any harcoded values that can block of use bigger resolutions/textures or waste memmory in case 320*240 resolution, so am working on reallocate buffers on resolution change.

sezero:
> Do you know exact line/state where value overflows?

Really can't remember, sorry.

0lvin:
@sezero look as change shift size is enough. 
`
if ((vid.width >= 2048) && (sizeof(shift20_t) == 4)) // 2k+ resolution and 32 == shift20_t
{
	shift_size = 18;
}
else
{
	shift_size = 20;
}
`

So no crashes for 2k+ resolutions.

0lvin:
> @Yamagi It would be really nice if at least overbrightbits could be implemented into the softrenderer.

I have implemented sw_overbrightbits. Could you please review changes? 

I am looking to idea for use  true color(24bit) instead 8 bit palette. As i know, code use 16 colors with 16 levels for now and we can't use any full color images.

DanielGibson:
who needs true color in the SW renderer?
if people want the game to be pretty they can just use on of the OpenGL renderers

0lvin:
@DanielGibson Ok, so i  squashed all commits and you can review/check changes, code disabled by default.

Yamagi:
Hey, that's very good work. I've tested a little bit and so far nearly everything seems to work. UI scaling, the overbright bit are looking very good, extremly high resolutions are running fine and even the annoying problem with water surfaces doesn't occure. :)

Just two small bugs: 

* If the videos are played in an widescreen window they're are placed in the middle and the surround background should be black. With the softrenderer that background is magenta. That's likely an "set everything to black" missing somewhere.
* The brightness slider in the video menu is inversed. 

There may be a problem with the vsync. The game holds 60hz just  perfectly but there seems to be some micro stuttering, like some frame are delayed for too long. But that may be an Xorg or driver problem on my side. I'll check that in the next few days and report back.

Otherwise I'm fine with merging that. One remaining question is what to do with the cvars. I have an idea how we could rewrite the cvars at config loading time. With such mechanism we could rename gl_* back to r_*. I'll to discuss that with @DanielGibson.

And you may add yourself to the credits in menu.c right below "Sander van Dijk". :)

0lvin:
@Yamagi 
* Gamma/brigness fixed
* background cleaned up by minimal color in pallete
* added self to credits

About vsync - i have not seted VSYNC flag before 60b444834e41faf4ad96e5d33a328921bd2c0b4c - am not sure that it relly change anything.

iriyap:
I'd vote for renaming all the *r_ cvars back. They are named this way in all id software games and renaming them is breaking an important tradition and engine inter-compatibility (or even with a render as you've noticed).

From the end user standpoint, say, I just downloaded Yamagi and started making maps, I need to set r_lightmap 1 to see if my lightmaps are looking good, or do r_showtris 1 to see if the compiler generated any garbage faces, or do r_fullbright 1 to check out the texturing in dark places. You get my point... Why should the user re-learn just for this specific engine?

Yamagi:
I gave it another try and the vsync looks good now. I'm not quite sure if it was your change or just the fact that I'm on an machine with the Nvidia blob and not on my rather slow laptop. Nevertheless: Good work!

0lvin:
@Yamagi Thank you:-)

Yamagi:
Hi,
I just tried to implement CMake support for the soft renderer and stumbled over a somewhat scary problem. There're some symbol collissions between the client and the new ref_soft.so: At least some of the symbols in vid.c (starting with line 438) are  defined in both the client and the ref_soft.so library, which symbol takes precendence is dependend on the linker flags. If the symbols in the client have higher precedence, the ref_soft.so calls back into the client as soon as one of these symbols is referenced for the first time. The function in the client tries to call the one in the ref_soft.so, but instead it calls itself. This results in an endless recursion until the stack overflows... So, can you rename the colliding symbols in ref_soft.so?

Thank you,
Yamagi

Yamagi:
The CMake integration (and some other very small changes) can be found here: https://github.com/yquake2/yquake2/tree/softrenderer

0lvin:
@Yamagi fixed by rename to RE_*

Yamagi:
And now it's merged. Thank's again. :) I'm going to disable the softrenderer by default until @DanielGibson  and I have renamed the renderer cvars. That'll happen some time after christmas, likely in early january. After that we'll ship it in the default install.

Issue #257, opened 2017-11-21
Mouse not working after last windows update


LordReven666:
I had the same problem with quakespasm(SP for Q1) but they recently fixed this with a patch, ever since the last major windows update my mouse does not move in game.

DanielGibson:
are you using 7.02 or 7.02a?
7.02a had critical fixes for Windows, you can get it at http://deponie.yamagi.org/quake2/windows/quake2-7.02a.zip

If that doesn't fix your problem, please try replacing the SDL2.dll with the one from http://libsdl.org/release/SDL2-2.0.7-win32-x86.zip - if there is a mouse problem on Windows, it's most probably an issue that would have to be fixed in SDL, maybe they have already done so before the latest release (2.0.7 is from October 22nd, and thus newer than the version we're bundling).

If it still doesn't work then, you'll need to tell us more about your system (which version of windows are you using, do you know which specific update broke the mouse input etc?)

LordReven666:
Just tried both of the first suggestions and still not working.

I have a 64-bit Windows 10 Home edition PC with Intel i7 6700, Geforce 980, 16gb RAM, SoundBlaster Zx, and my mouse is an Mionix Avior(recently updated the firmware also) 

I believe the name of the update was the "Windows 10 Fall Creators update".

ericwa:
AFAIK, we didn't fix anything in Quakespasm, just changed the default exe to be the SDL2 one.

My understanding is, SDL1.2's mouse input stopped working as of the Windows 10 Fall Creators update. SDL2's still worked (it uses the Windows raw input API by default), except if you set the `SDL_HINT_MOUSE_RELATIVE_MODE_WARP` hint to `1` (not the default).

This is the related SDL bug: https://bugzilla.libsdl.org/show_bug.cgi?id=3931

I haven't tried yquake2 but I can try reproducing this on Win 10 Fall Creators.

ericwa:
I couldn't reproduce any mouse glitches with http://deponie.yamagi.org/quake2/windows/quake2-7.02a.zip
Tried both the included SDL2.dll and http://libsdl.org/release/SDL2-2.0.7-win32-x86.zip

This is Windows 10 Fall Creators Update (1709 build 16299.64)

btw, what exactly happens for you when the mouse input is broken @LordReven666 ?

LordReven666:
Interesting, well when I am in game i cannot move my mouse, it is as if its stuck or something. Works fine in every other game I have now, except for yquake2.

[...]

PR #266, opened 2017-12-17
quake2.exe => yquake2.exe; quake2.exe wrapper

DanielGibson:
Apparently something (possibly nvidia's driver) on some windows installations has some stupid application profile for quake2.exe that breaks mouse input if the console has been opened, see #257 
    
The workaround is to rename quake2.exe to yquake2.exe and provide a wrapper quake2.exe that just calls the real one for backwards compatibility.


DanielGibson:
(Note that I only adjusted the Makefile, not cmake, and I'm still waiting for someone who is affected by the bug to test this workaround to make sure it actually works - after all "messing with input of a process  spawned by quake2.exe" or "messing with all executables with quake2 in their name" wouldn't be that much weirder than "messing with the input just because the process's executable is called quake2.exe" because in a sane world, this kind of thing should never happen at all)

PR #268, opened 2018-01-01
bus error fix when quitting. spotted on openbsd while

devnexen:
working ok on Linux and others.

### DIFF
--- a/src/common/cvar.c
+++ b/src/common/cvar.c
@@ -619,11 +619,11 @@ Cvar_Fini(void)
 
 	for (var = cvar_vars; var;)
 	{
-		cvar_t *c = var;
+		cvar_t *c = var->next;
 		Z_Free(var->string);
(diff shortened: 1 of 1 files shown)


DanielGibson:
looks good, thanks!

(And this could only have worked by chance, ASan or valgrind would have complained as well. Yamagi says this code has only been merged recently and we missed this in the review)

devnexen:
No problem I ve just noticed the indentation discrepancy sorry for that.

DanielGibson:
just noticed, you fixed your own bug here, it was introduced by your earlier PR: https://github.com/yquake2/yquake2/pull/232 =)

devnexen:
Very true :-) sorry for that I will be more careful if I ever do a new PR ;-)

DanielGibson:
no problem, bugs happen, just thought it was funny :)

PR #273, opened 2018-01-20
Soft render clean up.

0lvin:
Removed unused function and variables and added 'static' to local variables and functions.

### DIFF
--- a/src/client/refresh/gl1/gl1_light.c
+++ b/src/client/refresh/gl1/gl1_light.c
@@ -46,7 +46,7 @@ R_RenderDlight(dlight_t *light)
 	GLfloat vtx[3*18];
 	GLfloat clr[4*18];
 
-	unsigned int index_vtx = 4;
+	unsigned int index_vtx = 3;
 	unsigned int index_clr = 0;
 
 	glEnableClientState( GL_VERTEX_ARRAY );
--- a/src/client/cl_screen.c
+++ b/src/client/cl_screen.c
@@ -1207,7 +1207,7 @@ SCR_ExecuteLayoutString(char *s)
 
 			else
 			{
-				DrawAltStringScaled(x, y, block, scale);
+				DrawStringScaled(x, y, block, scale);
 			}
 
 			continue;
--- a/src/client/refresh/gl1/gl1_model.c
+++ b/src/client/refresh/gl1/gl1_model.c
@@ -39,7 +39,6 @@ byte *mod_base;
 void LoadSP2(model_t *mod, void *buffer);
 void Mod_LoadBrushModel(model_t *mod, void *buffer);
 void LoadMD2(model_t *mod, void *buffer);
-model_t *Mod_LoadModel(model_t *mod, qboolean crash);
 void LM_BuildPolygonFromSurface(msurface_t *fa);
 void LM_CreateSurfaceLightmap(msurface_t *surf);
 void LM_EndBuildingLightmaps(void);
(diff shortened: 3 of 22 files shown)


src/client/refresh/gl1/gl1_light.c
```diff
@@ -46,7 +46,7 @@ R_RenderDlight(dlight_t *light)
 	GLfloat vtx[3*18];
 	GLfloat clr[4*18];
 
-	unsigned int index_vtx = 4;
+	unsigned int index_vtx = 3;
```
DanielGibson: This change looks like it makes sense, after all the `for ( i = 0; i < 3; i++ )` loop sets a value to `vtx[0]`, `vtx[1]` and `vtx[2]`, so it makes sense that index_vtx starts at 3, so other loop first sets `vtx[3]` and the following ones - the old code apparently left `vtx[3]` uninitialized *and* wrote past the end of the array.

Would be helpful to add this information to the commit message though, because the compiler error itself isn't very specific on what's going wrong there.

0lvin:
@DanielGibson Thank you, i have added litle more information about fix.

0lvin:
@Yamagi Could you please review?

Yamagi:
Hi,
I've been busy with my day job... Sorry. At a first glace the change looks okay. I'll have a deeper look at the weekend.

0lvin:
@Yamagi thank you

Yamagi:
And merged! Thank you :)

0lvin:
Thank you)

0lvin:
@Yamagi @DanielGibson What do you think about replace Sys_Milliseconds by SDL_GetTicks?  

Like 1ea7e4dfd88614e23cb889902ea08e9af658bc4b (I have replaced in client related codes, for q2ded need to link SDL.)

Yamagi:
Already tried that once and it doesn't work. We need to derive the millisecond counter from the microsecond counter, otherwise they diverge and the player sees subtile timing problems. And at my tests SDL_GetTicks() wasn't precise enough. I didn't look at the code back than but it seemed like SDL caches the value for some time and doesn't redetermine it at every call.

One well ment tip: Do yourself a favour and leave everything timing related alone. Quake IIs timing code is a bitch (sorry), even small changes can break it in very subtile and hard to notice ways. Those bugs are often even harder to debug. After literally tens of hours debugging and rethinking the code we're now in the comfortable situation that it works in all situations. Synchronous and asyncronos client, the network code against new and old server, the renderer with vsync enabled and disabled, on too slow CPUs or GPUs, broken GPU drivers (Hello AMD!), broken compositors, between all platforms, etc. It's just a big mess...

DanielGibson:
As much as I love SDL, I'm against making the whole code depend on it heavily.
Right now, we (mostly?) only use at at the "edges" to the system (window handing, sound, ...), very few source files make any SDL calls - even the OpenGL backends only have one source file with SDL stuff each.

For the SW renderer this is different, and that's OK because SDL **is** its graphics API, but let's keep the generic code generic.

0lvin:
I see, thank you.

iButane:
I noticed the particles increase in size as you select a higher resolution.  Shouldn't this be reversed? Or is it intentional? I took 3 screenshots--  320x240, 640x480 and 1920x1080.

maraakate:
That's an original q2 bug and I have a hack fix for it in q2dos but there's always a better way...

0lvin:
@iButane Could please you check changes in #285 ?

PR #275, opened 2018-02-06
Use  SDL_HAPTIC_CARTESIAN as effect direction.

0lvin:


0lvin:
@Yamagi @DanielGibson Could you check/review changes with you gamecontrollers/joystics? 

I have updated code to generate effects to bith types of sounds with/without fixed origin, and used distance as volume for sound with fixed origin.

I am no sure that ``` VectorSubtract(listener_forward, listener_up, orientation); ``` as good/correct way to get orientation. Openal code based on [it](https://github.com/yquake2/yquake2/blob/master/src/client/sound/openal.c#L624-L628) and i reused in someway.

Yamagi:
The code looks good. I don't have a gamepad or joystick with haptic feedback. @DanielGibson, can you give it a try?

0lvin:
Maybe @Pickle @xDShot like to test? :-) They showed interest in joystick support in previous #215

xDShot:
I don't have controller with force feedback, unfortunately.

0lvin:
@Yamagi Could you pleace cherrypick 80616a418eeca176c07980be343316a82a0099ca to master? 

It must fix issue after 832f3f6497cd6a42166eb82d0cf2cbb9c37da20b that required to init vid_displayrefreshrate before calculate fps and soft render does not do such.

Yamagi:
I've lent myself a XBox 360 gamepad. The rumbling feels good. As always: Thank you for the good work. 

I'm planing to do the next release (version 7.20) at the first weekend in march. This will be - of course - part of it.

0lvin:
Thank you.

PR #278, opened 2018-02-12
Don't forget entries of Multiplayer "Address Book"

DanielGibson:
By making sure the corresponding cvars ("adr0", "adr1", ..., "adr8") are always initialized.


iButane:
I use this port for MP and it's great minus the slow downloads of skins, maps etc.
Would be nice to have faster http downloads when connecting to servers.

PR #279, opened 2018-02-18
Fixing unzip build under FreeBSD/OpenBSD

devnexen:


Yamagi:
FreeBSD already has some special handling in the Minizip code. However not by me... And we likely need something similar for DragonFly, NetBSD, the Illumos family and so on. What about a more general approach: Define `IOAPI_NO_64` unconditionally on all platforms except Linux and Windows? Even if the platform needs fopen64() for large file support it's very unlikely that someone really has a >2GB ZIP file with assets.

krytarowski:
BSD switched to 64-bit offsets over 20 years ago (in BSD 4.4 or so).

Yamagi:
My suggestion can be found in d6f9cf6. It works on FreeBSD and Linux, of course. Would be nice if anyone could test it on OpenBSD.

devnexen:
Works fine.

Yamagi:
And merged to master. Thank you! :)

PR #285, opened 2018-03-25
Fixes scale particles in soft render

0lvin:
* window height as multiple for particles size.
*  Add custom particle shape by sw_custom_particles.
   0 - use rectangle (default shape)
   1 - use hexagon

### DIFF
--- a/src/client/refresh/files/wal.c
+++ b/src/client/refresh/files/wal.c
@@ -34,11 +34,18 @@ void
 GetWalInfo(char *name, int *width, int *height)
 {
 	miptex_t *mt;
+	int size;
 
-	ri.FS_LoadFile(name, (void **)&mt);
+	size = ri.FS_LoadFile(name, (void **)&mt);
 
 	if (!mt)
 	{
+		return;
+	}
+
+	if (size < sizeof(miptex_t))
+	{
+		ri.FS_FreeFile((void *)mt);
 		return;
 	}
 
--- a/src/client/refresh/gl1/gl1_image.c
+++ b/src/client/refresh/gl1/gl1_image.c
@@ -1013,10 +1013,10 @@ R_LoadPic(char *name, byte *pic, int width, int realwidth,
 }
 
 static image_t *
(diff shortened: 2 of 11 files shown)


DanielGibson:
What does the "Add size checks for WAL load" commit do?
Or, what problem does it solve? I can't remember any bugs with invalid .wal files

0lvin:
@DanielGibson 
> What does the "Add size checks for WAL load" commit do?
It is only small validation code, can be usable for broken *.wal files, same checks as we have in LoadPCX when if image is too short we return error.

>   Apparently the particles always have the same size (like "4x4 px" or whatever), regardless of resolution.
Good idea, need to check.

maraakate:
@DanielGibson: if you implement the better scaling code and it works well let me know.

0lvin:
@DanielGibson @maraakate Could you check? What do you think about such implementation?

maraakate:
ill try and implement the changes to q2dos over the next couple of days.

0lvin:
@maraakate Also if you like - code in https://github.com/0lvin/yquake2/commits/heretic2 support load maps from heretic2 demo.  Without game logic, but enough for static objects on map.

maraakate:
No, we're trying to keep Q2DOS relatively clean.  Superflous extensions like this won't benefit it.  Thanks though.

DanielGibson:
The implementation looks good!
Two things however:
* I think the particles could be a bit bigger - IMO, `pix = (izi * d_pix_mul) >> 7;` looks better than shifting by 8. But this is mostly a matter of taste I guess?
* I personally like the new particle shape better - but I prefer the OpenGL renderer with filtered textures anyway. I guess the people who prefer a softrenderer also prefer the oldschool square particles. I guess if you add rounder particles, that should be configurable via cvar?

0lvin:
@DanielGibson  I have checked ">> 8" to ">>7" and added "sw_custom_particles" for use default particles shape.

Yamagi:
And to me too :)

PR #289, opened 2018-03-29
fix compilation on musl libc.

maxice8:
based on https://github.com/voidlinux/void-packages/pull/13012/files


Yamagi:
Merged. Thank you. :)

Issue #292, opened 2018-04-22
Underwater rendering in soft renderer broken in non 4:3 resolutions


DanielGibson:
![q2_0073](https://user-images.githubusercontent.com/1216493/39095569-f0c9353e-4642-11e8-9df9-f9b5597025c0.png)
![q2_0074](https://user-images.githubusercontent.com/1216493/39095570-f0dfda46-4642-11e8-9dcc-30ade4c8898b.png)
The bars at top/bottom of the screen are not supposed to be there (at least they are not in 4:3 resolutions)

@0lvin can you take a look at this? :-)

iButane:
I'm guessing this was his way of fixing (bypassing),the underwater crash at higher resolutions. If you lower the res the bars disappear. I think it's one of those gamebreaking bugs that is impossible to fix.

0lvin:
#293 Possible fix for warp. I am going also to fix "train" disappear bug #281, but have not finished yet.

Yamagi:
This should be solves with the merge of pull request #293.

PR #293, opened 2018-04-27
#292 Warp fixes

0lvin:
* Add warp fixes;
* add texture filtering **sw_texture_filtering** based on Unreal from [Fabien Sanglard Article](http://fabiensanglard.net/quake2/index.php);
* replace all x/y/z 16 bit shift's to use constants SHIFT16XYZ;
* fix waring about string length.


0lvin:
@maraakate I have checked with: 1366x768, 1024x768 and 3840x2160. I am not sure that it will work with any resolutions and everywhere, but underwater on first level(outer base) looks good.

maraakate:
If you can survive that then it's probably OK.

maraakate:
Here is the comment @sezero  left about MAXWIDTH in r_local.h for q2dos not sure if you fixed this problem already:

/* be careful if you ever want to change MAXWIDTH: 12.20 fixed
 * point math used in R_ScanEdges() overflows at width 2048 !! */
#define MAXWIDTH		2040

0lvin:
@maraakate Fixed by use dynamic shift: https://github.com/yquake2/yquake2/blob/master/src/client/refresh/soft/sw_main.c#L1912

It has illogical thing - theoretically need to have bigger shift because size_of(int) == 8 on x86_64, but when i tried to shift more left (as we have bigger space) i saw  strange glitches.

If you like you can port "shift_size" logic to q2dos, it only replace all "shift 20 left" to "shift 18" if resolution bigger that hd.

maraakate:
it depends if it involves busting ASM can't do it.  Need ASM for DOS.  That's why the crappy particle size hack remains.  We're not supporting super high res in ref_soft anyways.

DanielGibson:
sizeof(int) is 4 basically everywhere, including x86_64

sizeo(size_t) and sizeof(void*) is 8 on x86_64

sizeof(long) is 4 on x86 and windows x86_64 (x64) and 8 on x86_64 on Linux, *BSD, OSX etc

sezero:
> because size_of(integer) == 8 on x86_64

NO, it is not. sizeof(int) is 4 on ILP32 systems, i.e. practically everywhere.

However sizeof(long) is 8 on LP64 systems, i.e. all unix, etc, (except for
win64 who obviously feel some sort of urge to be weird.)

DanielGibson:
Just tested it, the water works well now at 1280x720 and 1680x1050 :+1: 

`[-pedantic] fix compilation waring about static array initialization`
who cares, it works on all relevant compilers (and remember many compilers are not supported because we reequire C99 VLAs anyway)..  fixing code to work with `-Wpedantic` is a waste of time, IMHO

maraakate:
I can tell you many stories about pedantic warnings and pedantic bugs.

0lvin:
> sizeof(int) is 4 basically everywhere, including x86_64
yep, my mistake. I have also replaced use "integer" to shift20_t - so type can be changed in one place.

"-pedantic" fixes is useful in one case - we can search typos in code.

I also added "Unreal kernel filtering" with separate flag, from http://fabiensanglard.net/quake2/quake2_software_renderer.php

0lvin:
@DanielGibson @Yamagi Could you please review changes?

Yamagi:
Hi,
I finally found the time to have a look. This looks good. Do you think that's ready to merge?

0lvin:
Hi, thank you. As from me - yes, I think amount of changes in pull request is already really big and warp issue looks as fixed.

Yamagi:
And merged, thank you. :)

PR #303, opened 2018-06-12
Make uses of pkg-config cross-compilation-friendly

smcv:
One of the de facto standard interfaces used for cross-compiling is to override PKG_CONFIG at the same time as overriding CC, for example to `arm-linux-gnueabihf-pkg-config` and `arm-linux-gnueabihf-cc`. In particular, this is done when Debian packages are cross-compiled.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,6 +133,10 @@ else
 COMPILER := unknown
 endif
 
+# Used to detect libraries. Override to foobar-linux-gnu-pkg-config when
+# cross-compiling.
+PKG_CONFIG ?= pkg-config
+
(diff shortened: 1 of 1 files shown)


Yamagi:
That looks good. Thank you. But don't put too much effort in the X11GAMMA stuff. It'll go away when we remove SDL 1.2 support somewhere in the near future.

smcv:
I replaced pkg-config with $(PKG_CONFIG) without really looking at the modules we were checking for - it would have been more effort (and counter-productive) to make an exception for X11GAMMA :-)

PR #309, opened 2018-07-15
Fix segfault in OGG_PlayTrack when ogg_enable is set to zero

### LINKED ISSUE #308: Crash when starting a new game
When starting a new game, the program crashes segfaulting at src/client/sound/ogg.c line 314 because cd_shuffle is NULL and trackNo is nonzero. I believe this is due to ogg_enable being set to zero, which makes OGG_Init return early without initializing cd_shuffle.

rg3:
Fixes #308.

### DIFF
--- a/src/client/sound/ogg.c
+++ b/src/client/sound/ogg.c
@@ -555,17 +555,17 @@ OGG_Cmd(void)
 void
 OGG_Init(void)
 {
+	// Cvars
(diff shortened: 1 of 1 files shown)


Yamagi:
*hrmpf* It's the same as always. You do some rather big changes and of course you miss some edge cases during testing. And while everything's okay in the standard cases, one of them has an ugly bug...

Nevertheless: Thank you for the report and the fix! :)

Issue #281, opened 2018-03-04
Software Render Glitching


iButane:
Been testing out the software render... So far I've only noticed this 1 issue where the train model/texture glitches but only on higher resolutions.  
![20180303230925_1](https://user-images.githubusercontent.com/20784457/36942505-aa0bbd2e-1f41-11e8-89f1-bd340012786f.jpg)
![20180303230930_1](https://user-images.githubusercontent.com/20784457/36942506-ab53aaa2-1f41-11e8-8f92-61fb94b24efc.jpg)
After the train moves the issue is no longer there regardless of resolution.

Short video clips..

Before train moves: https://youtu.be/fA9_KGhoeTQ
After: https://youtu.be/6DEraxagW3U

DanielGibson:
@0lvin can you take a look at this as well? :)

0lvin:
I will look.

Yamagi:
Is there any progress in fixing the missing surfaces?

0lvin:
@Yamagi I have solutions for train, i am working on clean up changes. I am going to create pull request in week.

Yamagi:
Cool! That's very good to hear!

maraakate:
@0lvin I'd imagine, but didn't look too hard that raising the surf/edges limits fixed it?  I've ran into this problem before on large maps and high resolutions.

0lvin:
@maraakate We already allocate buffers on render init, so code will rellocate buffers in case "too small" and we will always will have enough space for any map size or count of vert's.

maraakate:
OK.  But was that related to the problem with the disappearing surfaces?  I saw you redid some code related to it.  I totally forgot about that being such a thing in the original ref_soft (it was a CVAR originally) and would cause this type of problem.

0lvin:
Yes, you are fully right, surfaces disappeared because no space for it in buffers. 

Now you can run ```noclip``` and see full map at once. Also it fixed issues in other games where map have big free space levels.

maraakate:
ref_gl has a similar problem with a max drawing distance if I remember correctly, but it's only relevant to huge "boomed" ports and probably not worth investigating for yq2.

Yamagi:
This was fixed with version 7.30.

PR #313, opened 2018-07-28
Scale edge/surfaces/verts on outof....

0lvin:
* Scale up allocated edge/surfaces/verts based on value in outof...
* Add check bounds for verts.
* Code style cleanups
* Fixed issue with disappeared train #281 
* Use common Mod_DecompressVis for all renders.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -454,6 +454,7 @@ set(GL1-Source
 	${REF_SRC_DIR}/files/pcx.c
 	${REF_SRC_DIR}/files/stb.c
 	${REF_SRC_DIR}/files/wal.c
+	${REF_SRC_DIR}/files/pvs.c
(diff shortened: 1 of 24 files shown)


DanielGibson:
I briefly tested it and it seems to work much better!
I only noticed one small glitch: In ware1 (same level as the train) right at the beginning there is this explodable crate, and from some view points it has a "hole" (but not from others):
q2_0075
q2_0077

Yamagi:
Very good work. Thank you! If no one has objections we could merge ist.

0lvin:
@DanielGibson Issue should be fixed in cf4887c. Could you check?

DanielGibson:
Yep, the glitch is fixed, great work!

Yamagi:
Aaaand merged! :)

PR #314, opened 2018-08-08
Render speedup

0lvin:


0lvin:
@Yamagi @DanielGibson Could you look? 

I have mainly changed logic in sw_scan.c for corner case when we have vertical/horizontal spans or span is in parallel to screen (z value does not changed between span end's).

Also in several places i have added const to unchangeable values or move static variables to function params for receive more "functional" style and have more clean code for check where variables changed or real inputs for functions.

Yamagi:
This looks very, very good. I'm impressed how you care for code quality. I've always been the pedantic type of developer and even after more than 10 years I'm still disgusted by Quake2s code... I'm totally fine with merging this if @DanielGibson has no objections.

For the bigger plan: In the last weeks I've removed SDL 1.2 and refactored the client <-> renderer interaction. The soft renderer is still special but that shouldn't be a problem as long there's only one renderer using the SDL Renderer interface. And I don't thing that we'll ever add another soft renderer. Yesterday I've finally replaced the libvorbis dependencies with the stb_vorbis single header library. I'm currently pondering if I could and should do the same with zlib, to get rid of all statically linked dependencies. After that I'll wait 2 or 3 weeks for bug reports and then do the next release, 7.30. Likely at the end of august or in early september.

0lvin:
@Yamagi 

* Do we need to call  SWimp_Shutdown in render on change soft->gl? If remove this func i see two windows, in other case render change failed with error `Error: Couldn't load libGL: Video subsystem has not been initialized!`.

* What do you think about rename `sw_texture_filtering` and `gl_anisotropic` -> `r_anisotropic` for aniso filtering in menu?

Yamagi:
* Hmm... soft -> gl and gl -> soft are working on my desktop with FreeBSD and Nvidia. But I see the same problems as you on the laptop with Linux and Intel.  
One problem is that ``SWimp_Shutdown()`` calls SDL_Quit*(). That's clearly wrong, SDL is now initialized and deinitialized by vid.c. The problem with the duplicate window is somewhat more tricky. I've missed a call to ``ShutdownGraphics()`` in ``GLimp_Shutdown() but that's not enough. We need  to destroy the window at renderer shutdown by calling ``ShutdownGraphics()`` from ``VID_ShutdownRenderer()``. With that the window handling works but the softrenderer doesn't show anything. So we likely need to reset SDLs internal GL state when the GL renderers are shutting down. And *tada* it work.  
The code (more or less untested!) can be found here: https://github.com/yquake2/yquake2/tree/renderswitch This is of course still hacky. In theory we would need a swimp_sdl.c at client side and vid.c needs to track if GL or soft renderer is running. But that's a lot of work. Maybe for cold winter evenings. :)
* Renaming both `sw_texture_filtering` and `gl_anisotropic` to `r_anisotropic` is a good idea. You need to add them to the replacement list in cvar.c, otherwise older config will break.

0lvin:
Cool, i will check. 

Also about "another soft renderer." :-) We also can have full color render (i suppose it can share 50% of code from current soft render) and vulkan render. Only because we can....

Yamagi:
If we want another soft renderer 'just because we can' we'll need a swimp_sdl.c to coordinate between them. 

In the old SDL 1.2 world there was no real distinction between the window and the GL state. When SDL 2.0 arrived we hacked the code to support the old SDL 1.2 way and the newly introduced window <-> state seperation. In the last weeks I've removed SDL 1.2 support, put the window handling in the client and the GL state handling in the renderer. Having the state handled by the renderer saves a lot calls between client and renderer. Additionally I've cleaned up some historical dept like the SDL reinitialization at vid_restart.

This is somewhat complicated by the missuse of vid_fullscreen. The client and all renderers set the vid_fullscreen cvar to 'modified' to communicate changes to the renderer configuration, triggering a vid_restart.

TL;DR: In theory the soft renderer needs to follow the changes made to the GL renderer. Because it worked / works I haven't done it.

0lvin:
@Yamagi Rebase to be9942d5160750b91d25d3fa4b80680bd15c3ed1 has fixed render changes.

Also i have added code for rename from `gl_anisotropic` to `r_anisotropic`.

I will continue my experiments with render speed up in other branch. And code is ready for review.

Yamagi:
I'm still fine with this. Let's @DanielGibson have a look. I'll merge it if he has no objections.

stuff/cvarlist.md
```diff
@@ -81,7 +81,7 @@ Audio:
 * **ogg_enable**: Enable Ogg/Vorbis music playback.
 
 * **ogg_ignoretrack0**: Normally Quake II disabled the background music
-  if a major objective has been archived by setting the music track to 0.
```
DanielGibson: Those double spaces at the line endings have a meaning: They make markdown insert at <br>, i.e. new line but not new paragraph. please leave them in.
(The single spaces at end of lines are indeed superfluous)

Maybe your editor stripped spaces at line ends automatically; Generally I recommend using git gui (or similar) to create commits, and only commit lines you actively changed (=> leave whitespace changes from your editor in other parts of the file(s) out of the commit).

src/client/refresh/soft/sw_main.c
```diff
@@ -1570,21 +1570,20 @@ GetRefAPI(refimport_t imp)
...
-static SDL_Surface *surface = NULL;
-static SDL_Texture *texture = NULL;
-static SDL_Renderer *renderer = NULL;
-static qboolean X11_active = false;
+static SDL_Window	*window = NULL;
+static SDL_Surface	*surface = NULL;
+static SDL_Texture	*texture = NULL;
+static SDL_Renderer	*renderer = NULL;
 
 /*
```
DanielGibson: why is this icon stuff here, this should already be done in glimp_sdl.c, called by `GLimp_InitGraphics()`

Yamagi: The softrenderer doesn't use the APIs provided by glimp_sdl.c, it does it's own stuff.

DanielGibson: Would it be possible to refactor it so it does use the glimp_sdl.c stuff?
I don't think it's tied to OpenGL in any way.

stuff/cvarlist.md
0lvin: > Those double spaces at the line endings have a meaning...

Good to know. I restored whitespaces in dc3da331f5dc081513c20c60b40b2c06936a4cfc.

src/client/refresh/soft/sw_main.c
0lvin: should be, maybe by add additional flag for create window without setting GL flags.

DanielGibson: This is already handled: `re.PrepareForWindow()` returns the `SDL_WINDOW_OPENGL` flag which is then used to create the window. Other rendering backends can just return 0 or whatever flag they need (`-1` means "an error happened", but that's not a valid SDL window flag)

Yamagi:
Merged. Thank you. My window handling fixes will follow shortly.

0lvin:
Thank you, @Yamagi i have pushed several commits 1c1bf1e100308644cca5d5d9f9605b81c9bc4830 [render_speedup_review](/0lvin/yquake2/tree/render_speedup_review) for reuse GLimp_InitGraphics in SoftRender, its not fully tested and maybe has some issues in render change, but soft->gl3->gl1->gl3->soft and change resolution are worked by first look.

0lvin:
@Yamagi @DanielGibson I have created separate #317 for this changes, also i have removed use of `SDLSurface` as we does not need such and changed texture format back as previous format rendered twice faster.

PR #315, opened 2018-08-14
Xcode project for macOS

tomkidd:
Hey, I added an Xcode project for Mac IDE development. Not sure if you're interested in it but if you are here it is.


DanielGibson:
Does this have any advantages over generating an xcode project with cmake?
If so, would it be possible to adapt our cmake project so it's as good?

src/server/sv_main.c
```diff
@@ -35,7 +35,7 @@ client_t *sv_client; /* current client */
 cvar_t *sv_paused;
 cvar_t *sv_timedemo;
 cvar_t *sv_enforcetime;
-cvar_t *timeout; /* seconds without any message */
+cvar_t *q2timeout; /* seconds without any message */
```
DanielGibson: why?

tomkidd:
@DanielGibson I'm not sure, it's been so long since I worked with makefiles and the like that I forgot about cmake, so I'll look into it. If cmake's result can be improved I'll do that.

src/server/sv_main.c
tomkidd: I forgot about that - yeah the name "timeout" now conflicts with a variable in the curses.h file that the macOS 10.13 SDK includes. I'm not sure the path it's being included in, or if a makefile-based build has the same issue (probably not) so between this and the cmake thing you probably want to decline this pull request (however that works, not sure)

Yamagi:
We aren't going to merge this. We've got a CMakeLists.txt to support foreign IDEs and build systems. I have never tested it  on OS X, if it isn't working it's better to fix it then putting effort in another build system that won't be maintained by anyone and will stop working within weeks.

tomkidd:
I agree. Thanks.

PR #316, opened 2018-08-26
Use CMake 'GREATER' instead of 'GREATER_EQUAL'.

makson96:
Hi,
'GREATER_EQUAL' was introduced in CMake 3.7:
https://cmake.org/cmake/help/v3.7/release/3.7.html#commands
CMakeList.txt states that minimum required version of CMake is 3.0. So please use 'GREATER' instead of 'GREATER_EQUAL'.
Please also note, that Ubuntu 16.04 has CMake 3.5, so using 'GREATER_EQUAL' will break build process for this platform.
Best regards,


Yamagi:
Hmm... I must admit that I didn't test my CMake changes with older versions. Thank you!

PR #317, opened 2018-08-28
Cleanup code in soft render

0lvin:


src/client/refresh/soft/sw_main.c
```diff
@@ -1680,17 +1632,26 @@ CreateSDLWindow(int flags, int w, int h)
...
+	/* Clear the entire screen to our selected color. */
+	SDL_RenderClear(renderer);
+
+	/* Up until now everything was drawn behind the scenes.
+	   This will show the new, black contents of the window. */
+	SDL_RenderPresent(renderer);
 
 	texture = SDL_CreateTexture(renderer,
-				    SDL_PIXELFORMAT_ARGB8888,
+				    SDL_PIXELFORMAT_BGRA32,
```
DanielGibson: SDL_PIXELFORMAT_BGRA32 seems to be a relatively new addition to SDL2.

I'd suggest using something like:
```c
int pixelformat;
#if SDL_BYTEORDER == SDL_BIG_ENDIAN
    pixelformat = SDL_PIXELFORMAT_BGRA8888;
#else
    pixelformat = SDL_PIXELFORMAT_ARGB8888;
#endif
```
and then pass the `pixelformat` variable to `SDL_CreateTexture()`

It's a bit uglier, but at least it works with older SDL2 versions, including the one used by Ubuntu 16.04, Linux Mint 18.x and other slightly old Linux distros.

DanielGibson:
While testing I just noticed something else (sorry, I should have brought that up earlier when you did that change): I find it confusing that `r_anisotropic` enables this texture filtering/dithering in the software renderer - this is very different to what it does in the GL renderers (where you can use it even with otherwise unfiltered textures). I also guess that many people who use the software render want that pixely look, so they'd be surprised by this change as well and wouldn't know how to disable it, as it's non-obvious.
Another thing about the dithering I noticed is that it's only applied to world geometries, but not models - is that intended?

0lvin:
@DanielGibson SDL pixel format fixed. Thank you.

>...it's only applied to world geometries, but not models....

Yes for now. Models have used fully different way of rendering and dithering is not applied on it. What about such note af5c029655640fa3abdb402dde9639ad4eb199d1 ?

0lvin:
@Yamagi @DanielGibson I have reverted texture filtering and removed commit related to add const to shared headers.

Yamagi:
Cool, thank you!

DanielGibson:
Thanks for the changes!

Unfortunately, when I just tested your branch again, I noticed some glitches of walls clipping into models that are in front of them, see screenshots
q2_0078
q2_0083

DanielGibson:
Seems to work now, thank you very much for the quick fix! :-)

Yamagi:
Thank you!

0lvin:
Thank you! 

Small facts for history :-). 
Timedemo values (`timedemo 1;`):
* [QUAKE2_7_30](a6b85eaa):
  `map demo1.dm2` = 27.1 fps (+12% improve)
  `map demo2.dm2` = 25.5 fps (+13% improve)

* [QUAKE2_7_21](75300795):
  `map demo1.dm2` = 24.2 fps  (+0% not changed)
  `map demo2.dm2` = 22.5 fps  (+0% not changed)

* [QUAKE2_7_20](c0291746) :
  `map demo1.dm2` = 24.2 fps (+1% improve)
  `map demo2.dm2` = 22.5 fps (+1% improve)

* [QUAKE2_7_10 + soft render](b993e9d3):
  `map demo1.dm2` = 23.8 fps
  `map demo2.dm2` = 22.0 fps

Issue #263, opened 2017-11-30
Should MP via network work on OSX?


snrkl:
Hi - I have been playing with this and have managed to get it all working single player..
(instructions were great - thanks for that!)

Quick question: Should multiplayer on OSX work? Client cannot see the server in the server browser.  (Tried with three macs  (2x Mavericks 10.11.6 and 1x Sierra 10.12.6). Doesn't work in any direction...

No firewalls in the way or networking problems.. machines are both on the same Layer2 subnet (same network LAN segment) and can ping each other just fine..

outright `connect <ip.add.re.ss>` just times out:

Client gives a "`NET_Socket: bind: Address already in use`" error,  then just says:
```
Connecting to <ip.add.re.ss>... 
Connecting to <ip.add.re.ss>... 
Connecting to <ip.add.re.ss>... 
```

A tcpdump on the server shows the client broadcasting trying to find the server.. If I use a 'connect' command, you can just see the client connecting and no response from the server:

```
23:21:21.124292 IP 10.10.10.100.64726 > 255.255.255.255.27910: UDP, length 11
23:21:26.113305 IP 10.10.10.100.64726 > 255.255.255.255.27910: UDP, length 11
23:21:45.672761 IP 10.10.10.100.60290 > 10.10.10.103.27910: UDP, length 17
23:21:48.677614 IP 10.10.10.100.60290 > 10.10.10.103.27910: UDP, length 17
23:21:51.686695 IP 10.10.10.100.60290 > 10.10.10.103.27910: UDP, length 17
23:21:54.695611 IP 10.10.10.100.60290 > 10.10.10.103.27910: UDP, length 17
```
etc...

I have tried this in both directions, and the same behaviour is seen either way (ie: Machine A server, Machine B client.. retest with Machine A client, Machine B server...)

Yamagi:
It should, of course. But I don't thing that anyone tested it in the last years. And I don't have a Mac to do it... :/

Yamagi:
Can anyone tell if this still is a problem with the current HEAD?

snrkl:
Are there precompiled OSX binaries for the version you want me to test? A look at https://deponie.yamagi.org/quake2/osx/ shows me a 7.20 binary download, but not a 7.30 binary..

If you can let me know in a little more detail what you want me to test, I am happy to do it..

Yamagi:
[...]

PR #322, opened 2018-09-11
Fix IPV6 networking

Mathyn:
When hosting or joining a server via a IPV6 address no connection can be made. This is because of the ipv6 address '::' which should be '::1'. This PR fixes this mistake.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -360,10 +360,10 @@ client:
 	$(MAKE) release/quake2
 
 ifeq ($(YQ2_OSTYPE), Darwin)
-build/client/%.o : %.m
+build/client/%.o : %.c
 	@echo "===> CC $<"
 	${Q}mkdir -p $(@D)
-	${Q}$(CC) $(OSX_ARCH) -x objective-c -c $< -o $@
+	${Q}$(CC) $(OSX_ARCH) -x objective-c -c $(CFLAGS) $(SDLCFLAGS) $(ZIPCFLAGS) $(INCLUDE)  $< -o $@
 else
 build/client/%.o: %.c
 	@echo "===> CC $<"
--- a/src/backends/unix/network.c
+++ b/src/backends/unix/network.c
@@ -829,7 +829,7 @@ NET_Socket(char *net_interface, int port, netsrc_t type, int family)
 	if (!net_interface || !net_interface[0] ||
 		!Q_stricmp(net_interface, "localhost"))
 	{
-		Host = (family == AF_INET6) ? "::1" : "0.0.0.0";
+		Host = (family == AF_INET6) ? "::/0" : "0.0.0.0";
 	}
 	else
 	{
--- a/src/backends/unix/network.c
+++ b/src/backends/unix/network.c
@@ -829,7 +829,7 @@ NET_Socket(char *net_interface, int port, netsrc_t type, int family)
 	if (!net_interface || !net_interface[0] ||
 		!Q_stricmp(net_interface, "localhost"))
 	{
-		Host = (family == AF_INET6) ? "::" : "0.0.0.0";
+		Host = (family == AF_INET6) ? "::1" : "0.0.0.0";
 	}
 	else
 	{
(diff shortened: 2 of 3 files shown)


DanielGibson:
I don't think this is correct.

`::1` is the equivalent of `127.0.0.1`, so it's a loopback-address, only reachable from the same machine, see https://en.wikipedia.org/wiki/Localhost

`::` is the equivalent of `0.0.0.0`, so it's the unspecified address, and listening on that address means "accept connections on all network interfaces and all their addresses", see https://en.wikipedia.org/wiki/IPv6_address#Unspecified_address

So it should be correct to listen on `::` (and not `::1`) here.

Mathyn:
You are correct! I changed it to '::/0'. For some reason this address does work and simply doing '::' caused 'address in use' errors on the dedicated server and simply finding no servers on the client.

Mathyn:
Just to clarify: '::' does not work on MacOS. I have not tested on Windows systems which might understand '::' to mean '::/0'. 

Maybe this is what caused this issue: #263 ?

DanielGibson:
I don't think this caused #263, after all that one was about IPv4?

Really weird that MacOS apparently needs `::/0` instead of `::`

Mathyn:
I'm not sure why it doesn't work. '::' should work but every time we use '::' no servers can be found and the dedicated server claims the address is already in use.

So far I've been able to verify this behavior on three Macbooks.

DanielGibson:
What are the exact error messages you get from the game?

And if I understand https://en.wikipedia.org/wiki/IPv6_address#Special_addresses correctly, it should probably be `::/128`, not `::/0`? Does that work as well?

Mathyn:
When using address '::' the following error is shown when trying to start a dedicated server:

```
NET_Socket: bind: Address already in use
```

When trying to join a server from the client this error is shown twice. No other errors are displayed as far as I can see.

Changing the address to '::/128' also works and looking at the table you posted appears to be the correct address. I'll update the PR.

This is really strange. I'm assuming (but have not tested) using address '::' works just fine on Windows and Linux?

Just to be complete. This is the terminal output when starting a dedicated server with the address set to '::'.

```
Yamagi Quake II v7.31pre
========================
Platform: Darwin
Architecture: x86_64
Byte ordering: little endian

Added packfile './baseq2/pak0.pak' (3308 files).
Added packfile './baseq2/pak1.pak' (279 files).
Added packfile './baseq2/pak2.pak' (2 files).
Added packfile '/Users/mathyn/Games/quake2/.//baseq2/pak0.pak' (3308 files).
Added packfile '/Users/mathyn/Games/quake2/.//baseq2/pak1.pak' (279 files).
Added packfile '/Users/mathyn/Games/quake2/.//baseq2/pak2.pak' (2 files).
Using '/Users/mathyn/.yq2//baseq2' for writing.
execing default.cfg
couldn't exec yq2.cfg
execing config.cfg
couldn't exec autoexec.cfg
NET_Socket: bind: Address already in use
-------- game initialization -------
LoadLibrary("game.dylib")
Game is starting up.
Game is baseq2 built on Sep 11 2018.

------- server initialization ------
50 entities inhibited.
1 teams with 2 entities.

==== Yamagi Quake II Initialized ====

*************************************
```

And this is the output shown in the terminal when entering the 'join server' menu in the client:

```
NET_Socket: bind: Address already in use
NET_Socket: bind: Address already in use
pinging broadcast...
pinging multicast...
NET_SendPacket: IPv6 multicast destination but +set multicast not specified: ff12::666
pinging 192.168.3.51..
```

When trying to host a game from the client I get the following output:

```
maxclients will be changed for next game.
deathmatch will be changed for next game.
==== ShutdownGame ====
NET_Socket: bind: Address already in use
NET_Socket: bind: Address already in use
-------- game initialization -------
LoadLibrary("game.dylib")
Game is starting up.
Game is baseq2 built on Sep 11 2018.

------- server initialization ------
0 entities inhibited.
0 teams with 0 entities.

Server restarted
Server disconnected, reconnecting
Connecting to localhost...
[0.0.0.0]:0: challenge
Mathyn connected
[0.0.0.0]:0: client_connect


-===================================-

 The Edge
```

Mathyn:
Any updates on this? Let me know if I need to investigate more or make any changes :)

Yamagi:
And merged, thank you :)

PR #329, opened 2018-09-23
Added .ent file support (from KMQuake2, it works!)

NeonKnightOA:
.ent file support for external .ent files (drop them in the maps/ directory) for trivial entity manipulation without having to modify the map. **It finally works!** Ported from KMQuake2. Open to changes, though.

### DIFF
--- a/src/common/collision.c
+++ b/src/common/collision.c
@@ -1610,16 +1610,60 @@ CMod_LoadVisibility(lump_t *l)
 }
 
 void
-CMod_LoadEntityString(lump_t *l)
+CMod_LoadEntityString(lump_t *l, char *name)
 {
+	// Knightmare- .ent file support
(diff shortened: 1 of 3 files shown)


Yamagi:
Hi,
that look good. But can we please stick to one coding style (remove the unnecessary whitespaces and indent with tabs) and remove the nonsense comments? After that it could be merged, I think.

DanielGibson:
You're aware that you can commit more than one file at once, though? :-P

NeonKnightOA:
I had to work with the online editor, as the disc where I held my fork just went belly-up. :/ That's why the number of commits is high.

Yamagi:
I've just merged this by hand (squished all commits into one) as d036868600056a7a29076cccb12ca3c66f05e29a. Thank you! :)

PR #332, opened 2018-09-26
Softrender cleanup

0lvin:
* Cleanup soft render little more
* Support for big bsp with 32k+ marksurfaces.
* allocate hunks with relation to original file size.
* save hunk size to size_t typed variable for unix systems.

### DIFF
--- a/src/backends/unix/shared/hunk.c
+++ b/src/backends/unix/shared/hunk.c
@@ -47,14 +47,16 @@
 #endif
 
 byte *membase;
-int maxhunksize;
-int curhunksize;
+size_t maxhunksize;
+size_t curhunksize;
 
 void *
 Hunk_Begin(int maxsize)
 {
+
 	/* reserve a huge chunk of memory, but don't commit any yet */
-	maxhunksize = maxsize + sizeof(size_t);
+	/* plus 32 bytes for cacheline */
+	maxhunksize = maxsize + sizeof(size_t) + 32;
 	curhunksize = 0;
 
 	membase = mmap(0, maxhunksize, PROT_READ | PROT_WRITE,
@@ -65,9 +67,9 @@ Hunk_Begin(int maxsize)
 		Sys_Error("unable to virtual allocate %d bytes", maxsize);
 	}
 
-	*((int *)membase) = curhunksize;
+	*((size_t *)membase) = curhunksize;
 
-	return membase + sizeof(int);
+	return membase + sizeof(size_t);
 }
 
 void *
@@ -76,14 +78,14 @@ Hunk_Alloc(int size)
 	byte *buf;
 
 	/* round to cacheline */
-	size = (size + 63) & ~63;
+	size = (size + 31) & ~31;
 
 	if (curhunksize + size > maxhunksize)
 	{
 		Sys_Error("Hunk_Alloc overflow");
 	}
 
-	buf = membase + sizeof(int) + curhunksize;
+	buf = membase + sizeof(size_t) + curhunksize;
 	curhunksize += size;
 	return buf;
 }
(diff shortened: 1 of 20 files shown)


Yamagi:
It would be nice to get at least a little bit of test exposure for the memory management changes. They should be okay, but you never know... A set of Windows binaries build from this branch can be found here: https://deponie.yamagi.org/quake2/misc/quake2-0lvin-fabda9b.zip

0lvin:
@Yamagi Thank you. 

I have [added logic](199c56e4555b9046da712b71dca76426629b7b68) for allocate hunk with size equal to filesize in case simple models(MD2/SP2) and multiplication to (sizeof(void*)  / 2 ) for brush model(that has pointers to parts).

0lvin:
I have rebased changes to last master.

Yamagi:
Oh yeah. I knew why I never looked at hunk.c. ;) The change is good but triggers a hidden bug on FreeBSD and likely all other systems besides Linux... The logic for the non Linux cases in `Hunk_End()` is crap:

* There's no need to distinguish between FreeBSD and other non Linux platforms.
* The third case, `new_size == `old_size` is missing. I guess we never hit that by pure luck. Since `n` is never reset to something else than 0 the `n != membase` check down below triggers and the client is quit.

A possible fix is here:

```
diff --git a/src/backends/unix/shared/hunk.c b/src/backends/unix/shared/hunk.c
index 26954b2a..dd19986e 100644
--- a/src/backends/unix/shared/hunk.c
+++ b/src/backends/unix/shared/hunk.c
@@ -97,55 +97,28 @@ Hunk_End(void)
 
 #if defined(__linux__)
        n = (byte *)mremap(membase, maxhunksize, curhunksize + sizeof(size_t), 0);
-#elif defined(__FreeBSD__)
-       size_t old_size = maxhunksize;
-       size_t new_size = curhunksize + sizeof(size_t);
-       void *unmap_base;
-       size_t unmap_len;
+#else
+       #ifndef round_page
+        #define round_page(x) ((((size_t)(x)) + sysconf(_SC_PAGESIZE)) & ~(_SC_PAGESIZE))
+       #endif
 
-       new_size = round_page(new_size);
-       old_size = round_page(old_size);
+       size_t old_size = round_page(maxhunksize);
+       size_t new_size = round_page(curhunksize + sizeof(size_t));
 
        if (new_size > old_size)
        {
-               n = 0; /* error */
+               /* Can never happen. If it happens something's very wrong. */
+               n = 0;
        }
        else if (new_size < old_size)
        {
-               unmap_base = (caddr_t)(membase + new_size);
-               unmap_len = old_size - new_size;
-               n = munmap(unmap_base, unmap_len) + membase;
-       }
-#else
- #ifndef round_page
- #define round_page(x) (((size_t)(x) + (page_size - 1)) / page_size) * \
-            page_size
- #endif
-
-       size_t old_size = maxhunksize;
-       size_t new_size = curhunksize + sizeof(size_t);
-       void *unmap_base;
-       size_t unmap_len;
-       long page_size;
-
-       page_size = sysconf(_SC_PAGESIZE);
-       if (page_size == -1)
-       {
-               Sys_Error("Hunk_End: sysconf _SC_PAGESIZE failed (%d)", errno);
-       }
-
-       new_size = round_page(new_size);
-       old_size = round_page(old_size);
-
-       if (new_size > old_size)
-       {
-               n = 0; /* error */
+               /* Hunk is to big, we need to shrink it. */
+               n = munmap(membase + new_size, old_size - new_size) + membase;
        }
-       else if (new_size < old_size)
+       else
        {
-               unmap_base = (caddr_t)(membase + new_size);
-               unmap_len = old_size - new_size;
-               n = munmap(unmap_base, unmap_len) + membase;
+               /* No change necessary. */
+               n = membase;
        }
 #endif
```

With this change (and some testing on OS X if someone steps up) I'm fine with this PR. :)

0lvin:
@Yamagi Thank you. Applied and rebased.

Yamagi:
Merged. Thank you. :)

PR #340, opened 2018-11-01
Minor bug-fixes: Case-insensitive file loading & Rogue dmflags display

earth-metal:
I noticed these issues while making some changes for CTF in the "Start Server" menu for myself:

1. Case-insensitivity isn't working for directly loading files when YQ2 is located in a directory tree with capitalization (when it is also executed from a different directory than where it resides).
2. It looks like there's a typo in the code that would prevent the Rogue dmflags from displaying in the menu.  (I don't have the mission pack to verify that, though.)

Also, let me know if you'd be interested in adding the CTF-specific game options to the "Start Server" menu and "dmflags" menu, and I'll send those through.


Yamagi:
Merged. Thank you. :)

We're always interested in improvements to the menu. But options specific to the ctf addon must be limited to it and not displayed if another addon or baseq2 itself is running, of course.

PR #342, opened 2018-11-02
Add support for CTF options to "start network server" menu.

earth-metal:
Allow numeric input of "capturelimit" cvar and toggling of CTF dmflags via menu system as follows:

1. "start network server" menu
- Remove "rules" selection from "start network server" menu.  (No coop support in CTF; defaults to deathmatch.)
- Add input box for "capturelimit."
2. "deathmatch flags" menu
- Remove flags ignored in CTF: "teamplay" & "friendly fire."
- Add toggling of "force join," "armor protect," and "no techs" dmflags.

Existing var names & dmflag definitions for Rogue support were re-used since they're otherwise idle in CTF.


Yamagi:
And merged. Thank you!

PR #344, opened 2018-11-07
Fix some typos in documentation

Calinou:


doc/01_index.md
```diff
@@ -8,8 +8,8 @@ Some general hints:
 
 * No matter what's the current keyboard layout or on which platform the
   game is running, the console is always opened and closed by pressing
-  **Left-Shift + Escape**.
-* Yamagi Quake II saves it's configuration, savegames and screenshots
+  <kbd>Left Shift + Escape</kbd>.
```
DanielGibson: I don't think using `<kbd>` is really an improvement here (and it looks worse when viewing the file as a plain text file)

doc/06_contributing.md
```diff
@@ -12,20 +12,20 @@ into the idea.
 
 Some rules to follow:
 
-* Use Github! Sign up for an account, fork our repository and send a
+* Use GitHub! Sign up for an account, fork our repository and send a
   pull request after you're satisfied with your work. We won't accept
-  patches send by mail or - even worse - as pastebin links.
+  patches sent by email or - even worse - as pastebin links.
```
DanielGibson: email? shouldn't it be e-mail or E-Mail or something?

DanielGibson:
Apart from my minor nitpicks this looks great, thanks for proof-reading! :)

doc/06_contributing.md
Calinou: [English seems to use email more often than e-mail](https://grammarist.com/style/e-mail-email/), so I opted for that form.

doc/01_index.md
Calinou: I reverted to bold instead of using `<kbd>`.

DanielGibson:
Thanks!

PR #349, opened 2018-12-15
Added parenesis for REGEX REPLACE command to fix this error:

Pickle:
CMake Error at CMakeLists.txt:25 (string):
  string sub-command REGEX, mode REPLACE needs at least 6 arguments total to
  command.

Added librt to link list (was required during cross compile)


Yamagi:
No progress for 6 weeks -> closed.

PR #352, opened 2018-12-30
added feature to delete savegames

JoBergeron:
Working on Unix-likes, would need to be ported to Windows if it's a feature we would like.

Simply press delete on the the selected savegame from the save or load menu and select "yes" to confirm deletion.

### DIFF
--- a/src/backends/unix/system.c
+++ b/src/backends/unix/system.c
@@ -490,6 +490,7 @@ Sys_Remove(const char *path)
 void
 Sys_RemoveDir(const char *path)
 {
+	char filepath[MAX_OSPATH];
 	DIR *directory = opendir(path);
 	struct dirent *file;
 
@@ -497,11 +498,11 @@ Sys_RemoveDir(const char *path)
 	{
 		while ((file = readdir(directory)) != NULL)
 		{
-			Sys_Remove(va("%s%s", path, file->d_name));
+			sprintf(filepath, "%s/%s", path, file->d_name);
+			Sys_Remove(filepath);
 		}
 
 		closedir(directory);
-
 		Sys_Remove(path);
 	}
 }
--- a/src/backends/unix/system.c
+++ b/src/backends/unix/system.c
@@ -487,6 +487,25 @@ Sys_Remove(const char *path)
 	remove(path);
 }
 
+void
+Sys_RemoveDir(const char *path)
+{
+	DIR *directory = opendir(path);
+	struct dirent *file;
+
+	if (Sys_IsDir(path))
+	{
+		while ((file = readdir(directory)) != NULL)
+		{
+			Sys_Remove(va("%s%s", path, file->d_name));
+		}
+
+		closedir(directory);
+
+		Sys_Remove(path);
+	}
+}
+
 /* ================================================================ */
 
 void *
--- a/src/backends/unix/system.c
+++ b/src/backends/unix/system.c
@@ -498,7 +498,7 @@ Sys_RemoveDir(const char *path)
 	{
 		while ((file = readdir(directory)) != NULL)
 		{
-			sprintf(filepath, "%s/%s", path, file->d_name);
+			snprintf(filepath, MAX_OSPATH, "%s/%s", path, file->d_name);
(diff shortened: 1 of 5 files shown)


src/client/menu/menu.c
```diff
@@ -2322,6 +2406,19 @@ LoadGame_MenuKey(int key)
         LoadGame_MenuInit();
         return menu_move_sound;
 
+    case K_DEL:
+
+		if ((item = Menu_ItemAtCursor(m)) != 0)
```
DanielGibson: I'd use NULL instead of 0

src/client/menu/menu.c
```diff
@@ -2439,6 +2537,18 @@ SaveGame_MenuKey(int key)
         SaveGame_MenuInit();
         return menu_move_sound;
 
+    case K_DEL:
+
+		if ((item = Menu_ItemAtCursor(m)) != 0)
```
DanielGibson: same here

src/common/filesystem.c
```diff
@@ -1411,7 +1411,7 @@ void FS_BuildGenericSearchPath(void) {
 	fsRawPath_t *search = fs_rawPath;
 
 	while (search != NULL) {
-		Com_sprintf(path, sizeof(path), "%s/%s", search->path, BASEDIRNAME);
+		Com_sprintf(path, sizeof(path), "%s%s", search->path, BASEDIRNAME);
```
DanielGibson: the slash must stay: two slashes are harmless, but we could run into the case of having no slash at the end of the path if it has been specified by the user via cvar

src/backends/unix/system.c
```diff
@@ -487,6 +487,25 @@ Sys_Remove(const char *path)
...
+Sys_RemoveDir(const char *path)
+{
+	DIR *directory = opendir(path);
+	struct dirent *file;
+
+	if (Sys_IsDir(path))
+	{
+		while ((file = readdir(directory)) != NULL)
+		{
+			Sys_Remove(va("%s%s", path, file->d_name));
```
DanielGibson: va() is not good here - we support paths with up to 4096 chars (`MAX_OSPATH`), but va() uses a 1024 char buffer.

DanielGibson:
looks mostly good, useful, apart from the things I mentioned and of course we need to also implement windows support (but I have some code that could be used as a base for this at https://github.com/DanielGibson/Snippets/blob/filehandling/DG_filehandling.h#L341 and following - it might not be well tested though - OTOH, there might be better/easier ways to just delete a directory recursively on Windows)

src/backends/unix/system.c
```diff
@@ -490,18 +490,19 @@ Sys_Remove(const char *path)
...
+	char filepath[MAX_OSPATH];
 	DIR *directory = opendir(path);
 	struct dirent *file;
 
 	if (Sys_IsDir(path))
 	{
 		while ((file = readdir(directory)) != NULL)
 		{
-			Sys_Remove(va("%s%s", path, file->d_name));
+			sprintf(filepath, "%s/%s", path, file->d_name);
```
DanielGibson: `snprintf()` would be better/safer :)

src/client/menu/menu.c
```diff
@@ -2322,6 +2406,19 @@ LoadGame_MenuKey(int key)
         LoadGame_MenuInit();
         return menu_move_sound;
 
+    case K_DEL:
```
0lvin: Do we can also attach some button for joystick? What do you think? @Yamagi @DanielGibson

DanielGibson: dunno..
Can the existing savegame menu functionality be used via joystick?
Is there an "obvious" joystick button for this? It would suck if people accidentally deleted their savegames if pressing the wrong button on the joystick.. (Yes, there's still the confirmation dialogue, but that would need to be joystick-navigable as well, so I fear that with one fatfingered press onto the controller one could accidentally delete and confirm)

DanielGibson:
Looks good to me!

DanielGibson:
(rebased and) merged, thanks! :)

PR #354, opened 2019-01-08
Fix case error for linux -> win32 cross compile

nrich:
When cross compiling for Windows on a Linux environment case sensitivity matters for file names.


DanielGibson:
thanks a lot!

PR #361, opened 2019-01-22
Enable automatic Windows builds with AppVeyor

EwoutH:
Add continuous integration with AppVeyor, which enables automatic builds of yquake2 for x86-64 Windows. Example can be found [here](https://ci.appveyor.com/project/EwoutH/yquake2/history).

AppVeyor only needs to be [enabled](https://github.com/marketplace/appveyor) on the GitHub Marketplace.

At the moment it produces this error, is this expected?
```
CMake Error at C:/Program Files (x86)/CMake/share/cmake-3.12/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find SDL2 (missing: SDL2_LIBRARY SDL2_INCLUDE_DIR)
Call Stack (most recent call first):
  C:/Program Files (x86)/CMake/share/cmake-3.12/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  stuff/cmake/modules/FindSDL2.cmake:173 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:96 (find_package)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Yamagi Quake II
+# Yamagi Quake II  [![AppVeyor Build Status](https://ci.appveyor.com/api/projects/status/github/yquake2/yquake2?branch=master&svg=true)](https://ci.appveyor.com/project/yquake2/yquake2)
 
 This is the Yamagi Quake II Client, an enhanced version of id Software's Quake
 II with focus on offline and coop gameplay. Both the gameplay and the graphics
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,6 +7,8 @@ install:
   - mkdir cmake_build
   - cd cmake_build
   - cmake --version
+  - find_package(SDL2 REQUIRED)
+  - include_directories(${SDL2_INCLUDE_DIR})
   - cmake c:\projects\yquake2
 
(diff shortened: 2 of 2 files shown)


Yamagi:
The general problem with Yamagi Quake II itself and all 3 addons  on Windows is, that the code does not compile with MSVC. At least not without ugly hacks. It uses several smaller C99 features and parts of the code (the old GL1 renderer) make excessive use of VLAs. The release builds are currently build with gcc in an MSYS2 environment. It may be possible to use a native clang instead, but I never had the patients to figure that out.

Regarding cmake and SDL2: I really don't know. The CMakeLists.txt works on FreeBSD, Linux and MacOS, so it should also work on Windows. But I've never tried it and I have no clue what the SDL2 module does to find the library. It look like there aren't search pathes for Windows in it - https://github.com/yquake2/yquake2/blob/master/stuff/cmake/modules/FindSDL2.cmake

Btw., q2vkpt is very nice. Keep up the good work! :)

Yamagi:
As said above this can't work and there's no visible progress in making it work. The account of the submitter looks like he's trying to earn "internet points" and has no real interest in YQ2. So I'm closing here.

Issue #355, opened 2019-01-13
[Feature suggestion] add detect native screen resolutions


johndoe71rus:
some game, example https://github.com/wolfetplayer/RealRTCW has option r_mode "-2". detect and set desktop resolution. Can you add this?

Yamagi:
@0lvin implemented this and I just merged his pull request. It'll be part of the next release.

PR #382, opened 2019-03-20
Add auto detect resolution

0lvin:
* Add auto detect resolution #355
* code cleanups for make cppcheck little more happy
* additional sanitize checks 
* move code patches authors from code comments to about menu

### DIFF
--- a/src/client/menu/menu.c
+++ b/src/client/menu/menu.c
@@ -1559,6 +1559,11 @@ static const char *idcredits[] = {
 	"Marty Stratton",
 	"Henk Hartong",
 	"",
+	"+Patches authors:"
+	"PGM",
+	"PMM",
+	"eliasm",
+	"",
 	"+YAMAGI QUAKE II BY",
 	"Yamagi Burmeister",
 	"Daniel Gibson",
--- a/src/client/menu/menu.c
+++ b/src/client/menu/menu.c
@@ -1559,7 +1559,7 @@ static const char *idcredits[] = {
 	"Marty Stratton",
 	"Henk Hartong",
 	"",
-	"+Patches authors:",
+	"+PATCHES AUTHORS",
 	"PGM",
 	"PMM",
 	"eliasm",
--- a/src/client/menu/menu.c
+++ b/src/client/menu/menu.c
@@ -1559,7 +1559,7 @@ static const char *idcredits[] = {
 	"Marty Stratton",
 	"Henk Hartong",
 	"",
-	"+Patches authors:"
+	"+Patches authors:",
 	"PGM",
 	"PMM",
 	"eliasm",
--- a/src/client/menu/menu.c
+++ b/src/client/menu/menu.c
@@ -1560,8 +1560,6 @@ static const char *idcredits[] = {
 	"Henk Hartong",
 	"",
 	"+PATCHES AUTHORS",
-	"PGM",
-	"PMM",
 	"eliasm",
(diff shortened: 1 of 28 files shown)


src/client/refresh/soft/sw_main.c
```diff
@@ -1469,6 +1470,9 @@ static int RE_PrepareForWindow(void)
 	return flags;
 }
 
+// Declared in vid/header/ref.h
+refexport_t	re;
```
DanielGibson: This is not supposed to be a global variable in the ref DLLs.
It *is* a global variable in the executable (see vid.c) (and is set from the return value of GetRefAPI() from the ref DLL)

src/client/refresh/gl3/gl3_sdl.c
```diff
@@ -341,3 +341,27 @@ void GL3_ShutdownContext()
...
+ * Detect default desktop mode
+ */
+qboolean
+GL3_GetDesktopMode(int *pwidth, int *pheight)
+{
+	// Declare display mode structure to be filled in.
+	SDL_DisplayMode current;
+
+	// We can't get desktop where we start, so use first desktop
+	if(SDL_GetDesktopDisplayMode(0, &current) != 0)
```
DanielGibson: can't you use https://wiki.libsdl.org/SDL_GetWindowDisplayIndex to get the desktop the window is currently on?

0lvin: Window is already destroyed when code run GetDesktopMode, I have moved detection code to client/vid/glimp_sdl.c as more reliable place for get window (f0dc7bd).

src/client/menu/menu.c
```diff
@@ -1559,6 +1559,11 @@ static const char *idcredits[] = {
 	"Marty Stratton",
 	"Henk Hartong",
 	"",
+	"+Patches authors:"
+	"PGM",
+	"PMM",
```
Yamagi: Do we really need to add PGM and PMM here? That are Peter Mack and Patrick Magruder, already credited in the Ground Zero sections. That just did some changes to the client and the renderes to add support for various new effects used in Ground Zero, mostly the weapons and the new power ups.

src/client/refresh/gl3/gl3_sdl.c
0lvin: @DanielGibson I have also added code for save window position, in other case window will always recreated/moved to first desktop.

src/client/menu/menu.c
0lvin: @Yamagi I tried to save authors of patches, and if they are in list already, i have removed it from changes.

src/client/refresh/soft/sw_main.c
```diff
@@ -1469,6 +1470,10 @@ static int RE_PrepareForWindow(void)
 	return flags;
 }
 
+// struct for save refexport callbacks, copy of re struct from main file
+// used different variable name for prevent confusion
+static refexport_t	refexport;
```
DanielGibson: What's the point of this change?
refexport is only used in GetRefAPI() and returned by value, I don't understand why it should be a global/static variable

0lvin: @DanielGibson Oops, thank you, i will fix.

Yamagi:
And now I'm back to the Quake II business. Merged. Thank you and keep up the good work. :)

PR #384, opened 2019-03-29
hunk api mremap supported by NetBSD as well.

devnexen:


Yamagi:
And merged. Thank you. :)

As an unrelated note, in case someone of the power who are stumbles over it: It would be really nice if we could get a generalized version of https://reviews.freebsd.org/D14835 into FreeBSD.

PR #388, opened 2019-04-05
Weapon selector mechanism for controllers

blargelJim:
In order to switch weapons in a convenient way on a controller one needs a better mechanism in Quake 2 to do so because of the limited number of buttons a controller.
This change consists of 2 changes:
1. Add a couple of bindings in the config.cfg (HAT_LEFT...HAT_DOWN for first 4 weapons to adress, one binding for the "selector key" (JOY6 on the PS4 controller on a retropie, the top right shoulder button).
2. Support in Quake 2 to let the user press "selector key" + HAT_LEFT...HAT_DOWN to reach 4 more weapons. The selector key should also be rebindable from the "control settings" in the GUI.

My solution now also introduces 4 new cvars, "weaponselector_weapon1"..."weaponselector_weapon4" for the weapons that can be reached through "selector key" + HAT_LEFT...HAT_DOWN. This will make it possible to use this for mods as well as the cvars could be setup in the config.cfg of the mod if you have weapons with new names.

### DIFF
--- a/src/client/cl_input.c
+++ b/src/client/cl_input.c
@@ -60,6 +60,7 @@ kbutton_t in_left, in_right, in_forward, in_back;
 kbutton_t in_lookup, in_lookdown, in_moveleft, in_moveright;
 kbutton_t in_strafe, in_speed, in_use, in_attack;
 kbutton_t in_up, in_down;
+kbutton_t in_weaponselector;
 
 int in_impulse;
 
@@ -192,6 +193,18 @@ IN_KLookDown(void)
 
 void
 IN_KLookUp(void)
+{
+	KeyUp(&in_klook);
+}
+
+void
+IN_WeaponSelectorDown(void)
+{
+	KeyDown(&in_weaponselector);
+}
+
+void
+IN_WeaponSelectoUp(void)
 {
 	KeyUp(&in_klook);
 }
@@ -627,6 +640,8 @@ CL_InitInput(void)
 	Cmd_AddCommand("impulse", IN_Impulse);
 	Cmd_AddCommand("+klook", IN_KLookDown);
 	Cmd_AddCommand("-klook", IN_KLookUp);
+	Cmd_AddCommand("+weaponselector", IN_WeaponSelectorDown);
+	Cmd_AddCommand("-weaponselector", IN_WeaponSelectoUp);
 
 	cl_nodelta = Cvar_Get("cl_nodelta", "0", 0);
 }
--- a/src/client/cl_input.c
+++ b/src/client/cl_input.c
@@ -198,15 +198,9 @@ IN_KLookUp(void)
 }
 
 void
-IN_WeaponSelectorDown(void)
+IN_WeaponSelector(void)
 {
-	KeyDown(&in_weaponselector);
-}
-
-void
-IN_WeaponSelectoUp(void)
-{
-	KeyUp(&in_klook);
+	
 }
 
 void
@@ -608,6 +602,7 @@ CL_InitInput(void)
 {
 	Cmd_AddCommand("centerview", IN_CenterView);
 	Cmd_AddCommand("force_centerview", IN_ForceCenterView);
+	Cmd_AddCommand("weaponselector", IN_WeaponSelector);
 
 	Cmd_AddCommand("+moveup", IN_UpDown);
 	Cmd_AddCommand("-moveup", IN_UpUp);
@@ -640,8 +635,6 @@ CL_InitInput(void)
 	Cmd_AddCommand("impulse", IN_Impulse);
 	Cmd_AddCommand("+klook", IN_KLookDown);
 	Cmd_AddCommand("-klook", IN_KLookUp);
-	Cmd_AddCommand("+weaponselector", IN_WeaponSelectorDown);
-	Cmd_AddCommand("-weaponselector", IN_WeaponSelectoUp);
 
 	cl_nodelta = Cvar_Get("cl_nodelta", "0", 0);
 }
--- a/src/client/cl_input.c
+++ b/src/client/cl_input.c
@@ -197,7 +197,7 @@ IN_KLookUp(void)
 }
 
 void
-IN_WeaponSelector(void)
+IN_AltSelector(void)
 {
 	
 }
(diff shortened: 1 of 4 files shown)


DanielGibson:
I haven't gotten around to look at the code more closely, but the general approach of `+weaponselector` sounds reasonable and being able to bind those combinations (instead of hardcoding the weapon names) is a good idea as well!

I wonder if this could be generalized a bit further so "weaponselector button" + "any known joystick button/hat direction" can behave like a logical key/button that can be bound directly? So e.g. in addition to `K_HAT_UP` we'd have `K_HAT_UP_WEAPSEL` (or whatever name), same for K_JOY*
Just a random idea, as I said I haven't looked closely at the code yet and don't know if it's feasible.

blargelJim:
Hi Daniel (Guessing this is your name :)

I like that idea, I think the Id team for sure would have made it like that if there were people playing with controllers back in those days. It would be cool to be able to say  'bind K_HAT_UP + K_JOY6 "use Grenade Launcher"'. It probably is a bit more work though cause I guess the parser would need to be able to understand it and some generalized form of it supported in cl_keyboard.c
I might try and have a look at it, but I think this solution is still probably a good leg-up for most players out there using example the retropie/retroarch + controller and a dual-stick controller like the ps4 controller or xbox controller. It is pretty standard in FPS-like game to choose weapon using these keys on a controller at least on the PS4.

Trying to dive into your doom3 port as well, would be awesome to bring that to the say the retropie. But might prove to be a bit to big of a challenge since Im guessing it will need a opengl es 2.0 port and Im just one guy already pushing 40h at my pay-for-the-food-and-rent job :)

DanielGibson:
I thought more about `bind K_JOY6 weaponselect` and then `bind K_HAT_UP_WEAPONSELECT "use Shotgun"` (or maybe `K_HAT_UP_ALT` or sth like that) - so ideally you could press the weaponselect key and the other key in the keybinding menu to set up that binding.
Not sure how hard that would be..

Would be cool to see dhewm3 on retropie - there is the http://www.continuation-labs.com/projects/d3wasm/ which is based on dhewm3 and uses OpenGL ES 2.0 (at least it used to, no idea if it needs 3.0 now), and an iOS port based on that: https://github.com/tomkidd/DOOM3-iOS which seems to use ES 2.0 as well.
dhewm3 doesn't have joystick/controller support yet, but I'd like to have that, just haven't gotten around to implement it myself so far.

DanielGibson:
By the way, how does Yamagi Quake2 work on retropie?
It doesn't support OpenGL ES 2.0 either - does t he OpenGL 1.x renderer work? Or has someone adapted the OpenGL3 renderer?

blargelJim:
Ahh, yeah that might work to.
Was going to experiment a bit more with it today but managed to catch a cold on that trip I did, so it will have to be on pause for a while till Im feeling a bit better.

Checked out the opengl es 2 web port someone did on your doom3 port. Pretty neat stuff. Tricky part would probably be getting enough speed on the retropie but sure would be fun trying it out. Would love to run doom 3 on it too!

Yamagi quake 2 works very well on the retropie i would say (got the 3 b+ model). It would need an opengl es solution to work with hw-accelerted gfx. Might be someone who has already done some work on that which could be brought into the yamagi quake 2 engine actually.
But the software render works nice and the game runs very smooth as it is today.

DanielGibson:
I fear that it might indeed be (much) too slow for dhewm3, as neither the CPU nor the GPU are very fast.. would be interesting to see how much fps you get though (if it works at all)

xDShot:
Maybe make a weapon selector, something like in DOOM 2016?
image

blargelJim:
Hi XD,

Personally, I was never a fan of this weapon wheel selector. To often I ended up choosing the wrong weapon and it was annoying to have to shift my focus from trying to aim at the bad guys to select the weapon I wanted. I prefer a solution where it is mapped to one button or one button + some other button. It could be cool to show which weapon is assigned to which button on the HUD though, especially if you press down a button, so you can easily see which button maps to which key.

Yamagi:
> It could be cool to show which weapon is assigned to which button on the HUD though, especially if you press down a button, so you can easily see which button maps to which key.

That may not be impossible, but hard to implement. The input stuff is split between the renderer and the client, the HUD is generated in the game and send through the server (over the network) to the client. For such display the HUD calls must be intercepted and altered in the client, without breaking mods and the like.

blargelJim:
Hi,

Thought I would try and introduce a new menu option, after "customize controls" - "customize controls alternative function". That way we would have a whole page where mappings to keys for a second use could be added. I would keep the "weaponselector" binding in the "customize controls". Then whatever you press together with "weaponselector" would be what happens. I would change "weaponselector" to "alternative key function" or something.
Any thoughts on it? No gurante I will finish it but I think it would not be way to much work.
Perhaps it might present like a "unfair advantage" in multiplayer contexts if someone could bind some advanced move to a "button X + Y". Is that something I would need to worry about?

blargelJim:
Hi,

I've re-worked the solution a bit more. I chosed to go with a cvar solution to be able to map the alternative keys and add an additional menu "customize controls alt function" (maybe the name should change - anyone got a better one?) where you can map the extra keys. The reason I choose to store it cvars is simply because it seemed like to much of a mess to try and hitch on to the bind logic.

The menu sends a "bind KEY "do this thing" using Com_sprintf in Keys_MenuKey. The "bind" command in turn is tied to char *keybindings (cl_keyboard.c). So to support using bind, I would need to either expand the keybindings entry and re-write all functions that look into that (which is quite
intricate if one just freetext search keybindings and see all the code that peeks and pokes at this data).
I just think it introduces to much of a risk for bugs/unexpected behaviour to grant updating the "bind" code. If not updating the "bind" code I would need to introduce a new command and use that especially for this thing, which did not feel right either. Hence, I choose to use cvars and use the same infrastructure that the menu uses for binding keys (Com_sprintf, but sending a "set weaponblahblah_key_X "SOME KEY") and then I use the same cvars in cl_keyboard.c to figure out what button combination equals what action. This way, it can also be set in .cfg file on startup directly.

Any thoughts on this one? I want to clean it up a bit. Perhaps change name from weaponselector to "alternate" or something like that and try to add support for 3 more things I think might be nice to reach using the "alternate" key - inventory (previous, next, use).

I've already checked in this change so you can have a look at it and see if you think this is a sufficent solution though. And weigh in on the added menu, if you like that idea or not.
I did this all under a bit of a fever and headache but I just could'nt keep away - let me know if something looks weird or is just not proper.

Cheers
/Jimmie

DanielGibson:
I was wondering if it would get very complicated to implement this by generating different virtual keys if the altselector is pressed, which should make binding those combinations more straightforward overall, so I tried it out.
Turns out, it was simple enough, see https://github.com/yquake2/yquake2/pull/392

It doesn't have menu entries to bind the weapons, though (but this seems impossible to do in a generic way that doesn't hardcode the baseq2 weapon names in the game executable itself)..

blargelJim:
Pull request closed, accepted solution in in https://github.com/yquake2/yquake2/pull/392

Issue #393, opened 2019-04-27
Incorrect weapon lighting.


m-x-d:
SURF_DRAWSKY flag is never assigned in GL1/GL3 renderers, yet the code in RecursiveLightPoint and LM_CreateSurfaceLightmap checks for it.

In some setups this can result in incorrect weapon lighting. For example, you can get this ([close to the ground](https://i.imgur.com/N6MM4jg.jpg), [hovering slightly higher](https://i.imgur.com/LN0sqY8.jpg)) at the beginning of the first map of Q2MP2 (rmine1). Looks like Rogue devs used brushes with "light", "nodraw" and "sky" flags to light the maps, and the code in RecursiveLightPoint can pick surfaces form such brushes.

Yamagi:
That's a very nice find. This may explain some other lighting issues in the Rogue levels. We'll look into it.

Yamagi:
I just had a deeper look. You're right, the SURF_DRAWSKY checks can be removed. The flag is never set, so it can never be true. But removing them doesn't change anything, of course. The lighting problems you're seeing are likely caused by badly placed dlights, but I'm not 100% sure.

m-x-d:
The lighting problems are likely related to this:
1. LM_CreateSurfaceLightmap() call is skipped in Mod_LoadFaces(), because the logic checks surf->texinfo->flags for SURF_SKY flag (which is present).
2. Fullbright lightmap is created in R_BuildLightMap / GL3_BuildLightMap.
3. The resulting surface is not skipped by RecursiveLightPoint(), because it checks surf->flags for SURF_DRAWSKY (which is never set) and lightmap presence (which is present).

Yamagi:
Ah, I thought the wrong way. I think we'll need something in the lines of this: c765cbf

Without the fix:
![q2_0001](https://user-images.githubusercontent.com/1224576/57031159-36d64280-6c47-11e9-8cbd-9b407464b471.jpg)

And with the fix:
![q2_0000](https://user-images.githubusercontent.com/1224576/57031183-4b1a3f80-6c47-11e9-864e-b6363af29f35.jpg)

@DanielGibson, you're okay with it?

DanielGibson:
I mean it looks more correct, but maybe that effect was intended?
I can't imagine the addon level designers using the SW renderer to check the looks of their maps (because it has no colored lights) - I guess we should check what it looked like in the vanilla client with its old GL renderer, because that's probably how the creators intended things to look like.

[...]

PR #395, opened 2019-04-28
Make gl3 fast with AMDs Windows drivers

DanielGibson:
Seems like AMDs Windows driver doesn't like it when we call `glBufferData()` *a lot* (other drivers, incl. Intels and, IIRC, the AMD legacy driver used for the Radeon HD 6850 don't seem to care as much).
Even on an i7-4771 with a Radeon RX 580 I couldn't get stable 60fps on Windows without this workaround (the open source Linux driver is ok).

The be enabled/disabled with the gl3_usebigvbo CVar; by default it's -1 which means "enable if AMD driver is detected".

Enabling it when using a nvidia GPU with their proprietary drivers reduces the performance to 1/3 of the fps we get without it, so it indeed needs to be conditional...
So yeah, this is our first GPU-vendor specific workaround :-/

I'm not 100% sure about the code yet, I especially wonder if I should add some padding between the data in the VBO of consecutive calls to `GL3_BufferAndDraw3D()`..

### DIFF
--- a/src/client/refresh/gl3/gl3_main.c
+++ b/src/client/refresh/gl3/gl3_main.c
@@ -121,6 +121,7 @@ cvar_t *r_modulate;
 cvar_t *gl_lightmap;
 cvar_t *gl_shadows;
 cvar_t *gl3_debugcontext;
+cvar_t *gl3_usebigvbo;
 
 // Yaw-Pitch-Roll
 // equivalent to R_z * R_y * R_x where R_x is the trans matrix for rotating around X axis for aroundXdeg
@@ -205,6 +206,11 @@ GL3_Register(void)
 	gl3_particle_size = ri.Cvar_Get("gl3_particle_size", "40", CVAR_ARCHIVE);
 	gl3_particle_fade_factor = ri.Cvar_Get("gl3_particle_fade_factor", "1.2", CVAR_ARCHIVE);
 	gl3_particle_square = ri.Cvar_Get("gl3_particle_square", "0", CVAR_ARCHIVE);
+
+	//  0: use lots of calls to glBufferData()
+	//  1: reduce calls to glBufferData() with one big VBO (see GL3_BufferAndDraw3D())
+	// -1: auto (let yq2 choose to enable/disable this based on detected driver)
+	gl3_usebigvbo = ri.Cvar_Get("gl3_usebigvbo", "-1", CVAR_ARCHIVE);
 
 	r_norefresh = ri.Cvar_Get("r_norefresh", "0", 0);
 	r_drawentities = ri.Cvar_Get("r_drawentities", "1", 0);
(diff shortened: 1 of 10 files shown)


src/client/refresh/gl3/gl3_main.c
```diff
@@ -589,10 +621,72 @@ GL3_Shutdown(void)
...
+		 *
+		 * BTW, we couldn't observe this kind of problem with any other driver:
+		 * Neither nvidias driver, nor AMDs or Intels Open Source Linux drivers,
+		 * not even Intels Windows driver seem to care that much about the
+		 * glBufferData() calls.. However, at least nvidias driver doesn't like
+		 * this workaround (with glMapBufferRange()), the framerate dropped
+		 * significantly - that's why both methods are available and
+		 * selectable at runtime.
+		 */
+#if 0
```
DanielGibson: I should test and (most probably) delete this

src/client/refresh/gl3/gl3_main.c
```diff
@@ -589,10 +621,72 @@ GL3_Shutdown(void)
...
+		// as we make sure to use a previously unused part of the buffer,
+		// doing it unsynchronized should be safe..
+		GLbitfield accessBits = GL_MAP_WRITE_BIT | GL_MAP_INVALIDATE_RANGE_BIT | GL_MAP_UNSYNCHRONIZED_BIT;
+		void* data = glMapBufferRange(GL_ARRAY_BUFFER, curOffset, neededSize, accessBits);
+		memcpy(data, verts, neededSize);
+		glUnmapBuffer(GL_ARRAY_BUFFER);
+
+		glDrawArrays(drawMode, curOffset/sizeof(gl3_3D_vtx_t), numVerts);
+
+		gl3state.vbo3DcurOffset = curOffset + neededSize; // TODO: padding or sth needed?
```
DanielGibson: yeah, not sure about this (would it make sense to pad to some GPU-internal pagesize or sth?)

src/client/refresh/gl3/gl3_main.c
```diff
@@ -530,6 +536,32 @@ GL3_Init(void)
...
+	if(gl3_usebigvbo->value == 1.0f)
+	{
+		R_Printf(PRINT_ALL, "Enabling useBigVBO workaround because gl3_usebigvbo = 1\n");
+		gl3config.useBigVBO = true;
+	}
+	else if(gl3_usebigvbo->value == -1.0f)
+	{
+		// enable for AMDs proprietary Windows and Linux drivers
+		// TODO: should we match a version number? does the workaround
+		//       slow down legacy drivers that work fine without it?
```
DanielGibson: I need to test this with the AMD legacy driver on my Radeon HD 6850.
I think the GL3 renderer worked fine there without this workaround, but the real question is if it gets slower with the workaround enabled (which would mean we'd have to check the driver version number)

DanielGibson: .. of course it turns out that the legacy driver doesn't like this code and runs much smoother without the workaround :-/

(For the record, it's really a Radeon HD 6**9**50 and it's  the "Radeon Software Crimson Edition Beta" driver version 16.2.1 Beta, on Win7 x64. This is the latest driver AMD offers for that GPU.)

DanielGibson:
also includes the fixes for #393 now

src/client/refresh/gl3/gl3_main.c
DanielGibson: matching the version now..

src/client/refresh/gl3/gl3_main.c
DanielGibson: deleted

src/client/refresh/gl3/gl3_main.c
DanielGibson: still no idea, can be fixed in a later commit

DanielGibson:
updated binaries that contain this fix (and also the one for #393): https://deponie.yamagi.org/quake2/misc/quake2-d84bc47-20190504.zip

j4reporting:
could you please check the archive? Looks like some of the files like game.dlls  and docs are from the older 7.30 release package.

DanielGibson:
hmm right, the game dlls in ctf/ xatrix/ and rogue/ are old
feel free to replace them with the ones from the latest release - the main point of this snapshot is the updated yquake2.exe and rendering dlls

Yamagi:
A new set of Windows binaries (let's call it 7.41-RC1) with this fix included can be found here: https://deponie.yamagi.org/quake2/misc/quake2-298cb05-20190513.zip

PR #399, opened 2019-05-07
CMake fixes for Windows

maxcrofts:
This resolves a couple issues, allowing all targets to be successfully built for Windows using CMake.

I'm assuming that SDL is linked with q2ded on Windows for Unicode command line support. Would it be worth using `wmain` directly in order to remove that dependency?


Yamagi:
And merged, thank you. :)

SDL_main is mainly used for UTF-8 command line arguments, yes. And the dedicated server needs to be able to parse command line arguments with non-ascii chars in it. Besides that we try to minimize the differences between the full client and the dedicated server. So let's keep things as they are.

PR #400, opened 2019-05-08
quake2.exe wrapper puts Q2 window in foreground

DanielGibson:
because it sucks when you start quake2.exe and the game starts in the background

### DIFF
--- a/src/win-wrapper/wrapper.c
+++ b/src/win-wrapper/wrapper.c
@@ -31,6 +31,26 @@
 
 static const WCHAR WRAPPED_EXE[] = L"yquake2.exe";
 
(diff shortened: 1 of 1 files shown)


DanielGibson:
if anybody wants to test it, I attached a zip with the new quake2.exe wrapper
[yq2-wrapper.zip](https://github.com/yquake2/yquake2/files/3155029/yq2-wrapper.zip)

Yamagi:
A new set of Windows binaries (let's call it 7.41-RC1) with this fix included can be found here: https://deponie.yamagi.org/quake2/misc/quake2-298cb05-20190513.zip

PR #406, opened 2019-06-04
Allow scrolling with mouse in console

maxcrofts:
Scroll functionality for both the mouse wheel and buttons 4 & 5 was already implemented in `Key_Console()`. However, those keycodes never reached that function as they were not declared in `consolekeys`.

Once that issue was fixed another appeared. If those keycodes were unbound, warning messages would then appear in instances where the console was technically open but `cls.key_dest` was not set to `key_console` (e.g. fullscreen console).

In order to fix this unbound key warnings now only print when in-game. This is the only time when the warnings are actually useful as `Con_DrawNotify()` is only called during gameplay.

### DIFF
--- a/src/client/cl_keyboard.c
+++ b/src/client/cl_keyboard.c
@@ -1016,6 +1016,10 @@ Key_Init(void)
 	consolekeys[K_KP_PLUS] = true;
 	consolekeys[K_KP_MINUS] = true;
 	consolekeys[K_KP_5] = true;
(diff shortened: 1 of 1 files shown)


Yamagi:
I like this change. We should have restricted the unbound warning to `ca_active` long ago, they don't make sense in menus and the console and are annoying.

@DanielGibson: You okay with this, too?

Yamagi:
And merged. Thank you :)

PR #412, opened 2019-06-16
Multi-Monitor Management

Spirrwell:
This pull request adds an option to the "VIDEO" menu that allows you to specify which display/monitor you want the game to run on, and can be seen in the screenshot.

If this option is not desired, please feel free to disregard this pull request.

This adds a new archived Cvar "vid_displayindex" which the video menu sets and signals a restart with the "restart" qboolean in ApplyChanges()

Extra Notes:

The last_position_x and last_position_y variables don't actually seem to be used, though the window position is still stored in them.

I have a habit of adding extra spaces around parenthesis, and I notice I still accidentally did that, but if you prefer I conform to the no-spaces coding style, I can throw that in an extra commit.

If you have any other questions/comments/concerns, please let me know.

### DIFF
--- a/src/client/vid/glimp_sdl.c
+++ b/src/client/vid/glimp_sdl.c
@@ -64,6 +64,16 @@ GLimp_GetNumVideoDisplays(void)
 	return num_displays;
 }
 
+/*
+ * Resets the display index Cvar if out of bounds
+ */
+static void
+ClampDisplayIndexCvar(void)
+{
+	if (vid_displayindex->value < 0 || vid_displayindex->value >= num_displays)
+		Cvar_SetValue("vid_displayindex", 0);
+}
+
 static void
 ClearDisplayIndices(void)
 {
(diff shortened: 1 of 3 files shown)


src/client/vid/glimp_sdl.c
```diff
@@ -42,14 +43,68 @@ static int last_position_x = SDL_WINDOWPOS_UNDEFINED;
...
 // --------
 
 static qboolean
 CreateSDLWindow(int flags, int w, int h)
 {
+	// Reset display index Cvar if out of bounds
+	if ( vid_displayindex->value < 0 || vid_displayindex->value >= num_displays)
+		Cvar_SetValue( "vid_displayindex", 0 );
+
+	// last_position_x and last_position_y aren't really used...
```
0lvin: `last_position_x` and `last_postion_y` is used save a window position between render backend restarts.  
Before create a window - a window is positioned to center of the default screen. And values changed to real values after create window and before destroy window. And if we only change render backend, window will be recreated on the same screen, that prevent issues with the move window to unexpected place or create to big/small window if screens have different resolution.

As for me better change code to something like and save `vid_displayindex` before window close:
```c
if (last_position_x == SDL_WINDOWPOS_UNDEFINED || last_position_y == SDL_WINDOWPOS_UNDEFINED) {
      last_position_x = last_position_y = SDL_WINDOWPOS_UNDEFINED_DISPLAY( (int)vid_displayindex->value );
}
```

Spirrwell: Alright, I've updated it to work more like that, though I used SDL_WINDOWPOS_ISUNDEFINED() for clarity as SDL_WINDOWPOS_UNDEFINED is the same thing as SDL_WINDOWPOS_UNDEFINED_DISPLAY(0)

Other than that, I also made it so that the "display index" option in the VIDEO menu will be hidden if the user only has a single display so it's not confusing for people that only have one monitor.

Yamagi:
This is a cool idea. At a first look the code looks good. I'll have a deeper look in the next few days. @DanielGibson what do you think?

src/client/vid/glimp_sdl.c
```diff
@@ -144,12 +206,23 @@ void GLimp_GrabInput(qboolean grab);
...
 		/* save current display as default */
 		last_display = SDL_GetWindowDisplayIndex(window);
+
+		/* or if current display isn't the desired default */
+		if (last_display != vid_displayindex->value) {
+			last_position_x = last_position_y = SDL_WINDOWPOS_UNDEFINED;
+			last_display = vid_displayindex->value;
+		}
+		else {
 		SDL_GetWindowPosition(window,
```
DanielGibson: please fix the indentation of this (and the next) line to respect the else {} block

DanielGibson:
apart from the indentation thing, this looks good, thank you! :)

src/client/vid/glimp_sdl.c
Spirrwell: Alright, should be good.

Yamagi:
Aaaaand merged.... Thank you :)

PR #415, opened 2019-06-24
Soft render speed up

0lvin:
* fix mips select in Sonic Mayhem
* provide copy rectangle in SDL_RenderCopy
* skip frames without changes
* clean up sw_bsp.c

### DIFF
--- a/src/client/refresh/soft/header/local.h
+++ b/src/client/refresh/soft/header/local.h
@@ -454,14 +454,12 @@ extern msurface_t	*r_alpha_surfaces;
 //
 // current entity info
 //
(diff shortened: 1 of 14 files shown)


0lvin:
Before patch:
sonic-before
After patch:
sonic-after

Fps with cl_showfps=0:
before: demo1.dm2 ~ 1378 frames, 41.0 seconds: 33.6 fps
after: demo1.dm2 ~ 1378 frames, 40.0 seconds: 34.4 fps

Yamagi:
Cool, good work. :)

PR #416, opened 2019-06-26
Fix overrun in RE_BufferDifferenceStart

0lvin:
* Fixed incorrect pointer arithmetic in RE_BufferDifferenceStart, we can compare integers instead pixel_t, but should use pixel_t pointers in base.
```
back_buffer = (int*)swap_frames[0] + vmin * vid.width;
```
vs
```
back_buffer = (int*)(swap_frames[0] + vmin * vid.width);
```
* Added framebuffer damage based logic for get posible start/end diffeences from sprite render code.

### DIFF
--- a/src/client/refresh/soft/header/local.h
+++ b/src/client/refresh/soft/header/local.h
@@ -582,6 +582,9 @@ void	R_FreeUnusedImages(void);
 void R_InitSkyBox(void);
 void R_IMFlatShadedQuad( vec3_t a, vec3_t b, vec3_t c, vec3_t d, int color, float alpha );
 
+// VID Buffer damage
+void VID_DamageBuffer(int u, int v);
+
 /*
 ====================================================================
 
--- a/src/client/refresh/soft/sw_draw.c
+++ b/src/client/refresh/soft/sw_draw.c
@@ -119,6 +119,9 @@ RE_Draw_CharScaled(int x, int y, int num, float scale)
 		drawline = (vid.height - y) / iscale;
 	}
 
+	VID_DamageBuffer(x, y);
+	VID_DamageBuffer(x + (iscale << 3), y + (iscale * drawline));
+
 	while (drawline--)
 	{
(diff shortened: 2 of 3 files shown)


Yamagi:
Hi,
this breaks the menu. Not if it's opened while in game, but right after the game is started. The whole menu flickers and shows 'interlacing' artifacts.

DanielGibson:
the bug looks like this: image

0lvin:
Thank you.

@DanielGibson I can't reproduce issue in the stable way. I have seen such only one time. I have reverted last commit related to 2d textures damage. Last commit should improve check in menus and not critical to game mode. 

Could you recheck? Revert should fix issue.

I will recheck damage code and maybe create new PR or send as part of other PR.

Yamagi:
I just gave it a try and I'm still seeing the issue.

Yamagi:
Even with clean config. It seems to get worse when increasing the resolution. With 640x480 it's hard to notice, with 2560x1440 the menu is completely broken.

0lvin:
Issue looks as video driver specific behaviour logic related to partial textures update. I don't have other ideas for now.

@Yamagi Could you check now? I have reverted logic for search buffer difference end and use forced full buffer update be default.

Yamagi:
I can confirm that 50f1526 is working for me :)

Yamagi:
Can I merge it or do you want to do some other changes?

0lvin:
I have found reason of artifacts, it was related to partial update render from texture. As result fps change is not so impressive something like 27.1->29.2..30.7 instead 33+. Artifacts related to places where detected no change in current frame. By some reason artifact never appear in top of screen, but for sure I have deleted whole part of partial update(```SDL_RenderCopy(renderer, texture, &copy_rect, &copy_rect);```). 

@Yamagi @DanielGibson If you see no artifacts with new code, code can be merged.

Yamagi:
And merged. Thank you. :)

PR #431, opened 2019-08-01
Silence deprecation warning with OpenGL.

devnexen:
on Darwin, Mojave has deprecated it in favor of Metal,
here we silent the lengthy build warning.


Yamagi:
And finally merged. Thank you!

PR #442, opened 2019-09-05
Add zBuffer damage logic

0lvin:
* Save current player position and compare with a position in the previous
frame, if position is same refresh only changed rectangle in zbuffer.
zBuffer can be damaged by models, particles and static transparent spans.
zBuffer damage is not useful for the screen refresh minimize for now,
as result of explosion can change texture light value.
* Add logic for dump current position in world.
* remove screen flush after after a palette change, fixes issue with black screen on new game after show the introduction video


src/client/cl_main.c
```diff
@@ -596,6 +603,8 @@ CL_InitLocal(void)
 
 	Cmd_AddCommand("currentmap", CL_CurrentMap_f);
 
+	Cmd_AddCommand("currentview", CL_CurrentView_f);
```
Yamagi: I think it would be better to enhance the `viewpos' command instead of introducing a new one. `viewpos` already prints the view origin and yaw. Adding the angles to it would be easy.

Yamagi:
Thank you for the patch. I've added a comment to the `currentview` command. Otherwise I'm fine with this.

src/client/cl_main.c
0lvin: Updated

0lvin:
@Yamagi Could you review? I have moved position dump to `viewpos`.

PR #455, opened 2019-09-15
Fix for insane marines not always playing pain sounds

BjossiAlfreds:
This was a little thing partially caused by the change in sound prioritization. Insane marines will no longer scream or moan for 1 second after taking damage. To my knowledge the insane marines are the only enemy type in the game who need this special handling of sound priority. This also fixes the sound issue on nightmare difficulty, since they don't go into pain frames they constantly moan and scream which frequently would cut off the pain sound. I will also add pull requests for xatrix and rogue to address this there.

### DIFF
--- a/src/game/monster/insane/insane.c
+++ b/src/game/monster/insane/insane.c
@@ -27,6 +27,8 @@
 #include "../../header/local.h"
 #include "insane.h"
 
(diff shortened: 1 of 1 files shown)


Yamagi:
Merged. Thank you.

Issue #460, opened 2019-09-25
[Bug] Monsters run aimlessly if their enemy goes away (NULL or !inuse)


BjossiAlfreds:
For once this is actually a YQ2 bug. I first saw this happen in rogue, when a monster destroys a tesla it continues to run aimlessly. It affects all 3 games so I'm putting the report here.

My first thought was that ai_checkattack() was at fault and indeed it was. At the very top of the function it has a check for !self->enemy || !self->enemy->inuse and returns immediately if so. This is incorrect because ai_checkattack() actually expects the possibility of NULL or !inuse enemy as you see further down in the function, where it will update the monster's state accordingly. This is very similar to the ai_run() bug I reported a while ago.

I will provide a pull request that fixes this problem.

PR #461, opened 2019-09-25
Fix monsters not changing state after enemy goes away

BjossiAlfreds:
This pull request fixes bug https://github.com/yquake2/yquake2/issues/460. I will provide ones for xatrix and rogue as well.


Yamagi:
Merged. Thank you.

Issue #456, opened 2019-09-15
[Original bug] Idle monsters that duck down become disoriented and angry


BjossiAlfreds:
To reproduce:
1. Go to map lab
2. Set notarget on
2. Find brains
3. Shoot at him with the blaster or hyperblaster to make him duck down (this is random so may need a few tries)
4. Even though he gets angry and starts moving after he finishes ducking, you will notice he is having a hard time pathing to you, appearing disoriented

Without notarget, this bug will mostly occur if you shoot projectiles near corners, where the projectile can reach the monster without it seeing you directly.

(This also happens to other ducking monsters, it is just most obvious with brains because he has no ranged attack so you can observe his movement more closely)

What is happening is that in monster ducking functions, the enemy is set to the attacker, but there is no call to FoundTarget(), which causes the monster's state to be incomplete for chasing its enemy. I believe the disorientation comes from the monster's goalentity being different from its enemy (worldspawn?).

Rogue fixed this in their rewrite of the ducking code, but this still affects baseq2 and xatrix.

PR #468, opened 2019-09-27
Fix monsters being disoriented after ducking

BjossiAlfreds:
This pull request fixes bug https://github.com/yquake2/yquake2/issues/456 by adding a call to FoundTarget() after assigning the enemy in the dodge functions.


Yamagi:
And another one. Merged.

Issue #471, opened 2019-09-29
[Suggestion] Generalize xatrix weapon switching into a command


BjossiAlfreds:
To explain what I mean, in xatrix you can cycle between the hyperblaster and ionripper, and between the railgun and phalanx, by repeatedly pressing the respective use key ('8' and '9' by default).

This weapon switching style is hard-coded by xatrix for these specific weapons. You can find the code for it in g_cmds under the use command function and then the rest of it in player/weapon.

I find this type of weapon switching really nice so I extended it to rogue in my own mod. First I hard-coded it the same way as xatrix did, but later on I removed all that ugly hard-coding completely and wrote a command, cycleweap, that generalizes the idea.

The generalized idea is pretty similar to how you switch weapons in the Half-Life games. You group a set of weapons into categories, each category bound to a key, and by repeatedly pressing the same key for a particular category, you cycle through the weapons in that group.

Here is an example of how you would implement xatrix' default weapon cycling with the command:

bind 8 "cycleweap weapon_hyperblaster weapon_boomer"
bind 9 "cycleweap weapon_railgun weapon_phalanx"
(I use classnames rather than item names to avoid whitespaces and use the command parser's built-in argument splitting)

Is this something you'd be interested to see in YQ2? I already have the code and could easily migrate it over. Personally I never looked back once I started using this style of weapon switching in rogue, quickly cycling between chaingun and ETF rifle, hyperblaster and plasmabeam, etc.

Yamagi:
Sounds cool. If that command is optional I would merge it.

BjossiAlfreds:
It is completely independent from the use command and other weapon switching commands so entirely optional.

For xatrix you have two options though: leave the hard-coded system in, or supply an autoexec.cfg with the release that explains why it's there and binds 8 and 9 to cycleweap so that the user experience is still the same as before (well unless they re-bound those keys in their own file). That code wouldn't interfere with the command so if you don't mind ugly hard-coding and duplicate functionality it wouldn't hurt to keep it. This is entirely up to you.

PR #472, opened 2019-09-29
Added cycleweap command for generalized xatrix-like weapon switching

BjossiAlfreds:
This pull request adds the command cycleweap, suggested in https://github.com/yquake2/yquake2/issues/471.

Example use: bind 2 cycleweap weapon_shotgun weapon_supershotgun


Yamagi:
Merged. Thank you. I'll add he `cycleweap` to the list of autocomplete commands. And I need to add a section listing all added commands to the docs...

Issue #470, opened 2019-09-29
[Original bug(?)] Hand grenades don't play quad sound when thrown with quad damage


BjossiAlfreds:
To reproduce:
1. Open map jail2 and set "notarget" and "give all"
2. Go to the tank by the blue key door
3. Activate a quad damage powerup
4. Switch to the grenades
5. Throw two grenades at the tank
6. Notice that quad damage sound was never played even though quad damage was applied to the grenade damage

This one would be simple enough to fix by adding a gi.sound() call inside the player grenade fire function, but the question is, should this be considered a bug that we want to fix? What are your thoughts? Personally I definitely see it as a bug, a simple oversight at the time.

Yamagi:
The question is were to draw the line. All fixes you send so far were some corner cases that most players won't notice anyway, made the game more robust, helped with custom maps, etc. I think it's okay to make such chances. But things like playing the quad damage sound when throwing a grenate are changes that are easily noticed, exactly that kind of gameplay changes we try to avoid. So I would say that we should leave things as they are. @DanielGibson, what do you think?

BjossiAlfreds:
Just to be clear, I don't mean to play the sound all the time, just when the quad damage powerup is active like other weapons do.

Players expect to hear the quad damage sound when they fire the current weapon while the quad damage powerup is active. In fact I noticed this during a normal runthrough exactly because I expected to hear the sound but did not. I wasn't even sure if the quad damage was being applied to the grenade damage at all so I had to throw it at a durable enemy like a tank to be sure. I would say this is an anomaly, especially in the context of the code since all weapons except grenades use Weapon_Generic() which takes care of playing the sound, so it seems likely to me they just forgot add the sound to Weapon_Grenade().

Yamagi:
> Just to be clear, I don't mean to play the sound all the time, just when the quad damage powerup is active like other weapons do. I've added reproduction steps to the top comment that should demonstrate exactly what I have in mind.

I realized that. :) I just played a little and think that you're right. The sound should be added for consistency. Can you send a PR?

DanielGibson:
Throwing itself makes hardly a sound (just a slight "whoosh" from moving fastly through the air), maybe they thought it'd sound weird to add the quad damage effect to that (as the sound isn't even made by the weapon but your arm)?
[...]

PR #474, opened 2019-10-01
Quad sound for quadded hand grenades

BjossiAlfreds:
This pull request addresses the inconsistency reported in https://github.com/yquake2/yquake2/issues/470.


Yamagi:
Merged. Thank you.

PR #475, opened 2019-10-02
Copy drawn lines if picture scaled

0lvin:
If we have picture scaled and we can reuse same picture row in 1+ lines, code scale first one horizontally and copy by memcpy next one's verticaly. 

This code part have used in the intro videos.


Yamagi:
Coo. Thank you for your effort!

0lvin:
Thank you!

PR #476, opened 2019-10-08
Add quick save to server save game.

apartfromtime:
Overwrite "save1" slot with quick save.


Yamagi:
This isn't a good idea, because it's a huge change in behavior. Just overwriting savesgames may break some users workflow. I've added the quicksave to the menu, that must be enough.

PR #478, opened 2019-10-12
Added some text to cycleweap doc and added name to credits

BjossiAlfreds:
I was only going to add my name to the credits for this PR but I saw some important information missing from the documentation for my cycleweap command so I provided some extra text there as well.

### DIFF
--- a/doc/050_commands.md
+++ b/doc/050_commands.md
@@ -5,7 +5,10 @@ original clients (Vanilla Quake II) commands are still in place.
 
 
 * **cycleweap <weapons>**: Cycles through the given weapons. Can be used
-  to bind several weapons on one key.
+  to bind several weapons on one key. The list is provided as a list of
+  weapon classnames separated by whitespaces. A weapon in the list is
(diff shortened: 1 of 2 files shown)


Yamagi:
Welcome to the credits. :) And thank you for the small correction.

BjossiAlfreds:
Thanks! By the way, would you like me to add an entry for coop_baseq2 to the cvar docs? I noticed it is absent there. If I were to write docs for it, where in the file and under what category name would you like it to be?

Yamagi:
Back in Quake II land. Yeah. :)

A documentation entry for `coop_baseq2` would be very welcome. We try to document all cvars that we changed or added. Not a problem here, but good to know: We've got an alias mechanism for renamed cvars. The old and new name can be added to `replacements` array in cvar.c and it just works.

I think we'll do a release sometime in the next few weeks. The last one was in May and we've accumulated a lot changes.

BjossiAlfreds:
Let me know if you need any help with a release changelog, since I'm responsible for a lot of bugfixes.

PR #479, opened 2019-10-18
Added gl1_particle_use_pointparams to avoid the point parameters rendering path

ahluka:
ref_gl1 renders points weirdly on my machine. Way too big, and square.

After trying to work out why for an hour or so I just forced it to use R_DrawParticles2() regardless of point parameters support, which renders them correctly. I added gl1_particle_use_pointparms to make it slightly cleaner, but it's still quite hacky...

Someone with more experience can likely work out why R_DrawParticles() wasn't behaving, and point me to a real fix, which would be appreciated.

### DIFF
--- a/src/client/refresh/gl1/gl1_main.c
+++ b/src/client/refresh/gl1/gl1_main.c
@@ -88,6 +88,7 @@ cvar_t *gl1_particle_size;
 cvar_t *gl1_particle_att_a;
 cvar_t *gl1_particle_att_b;
 cvar_t *gl1_particle_att_c;
+cvar_t *gl1_particle_use_pointparams;
 
 cvar_t *gl1_palettedtexture;
 cvar_t *gl1_pointparameters;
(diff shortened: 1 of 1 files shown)


Yamagi:
> ref_gl1 renders points weirdly on my machine. Way too big, and square.

Ah, the broken point parameters extension again... Isn't it enough to set `gl1_pointparameters 0` instead of adding another work around?

ahluka:
>Isn't it enough to set `gl1_pointparameters 0` instead of adding another work around?

Just tried and it doesn't seem to be, although that would make sense and I obviously missed it!

I commented out the branch I added in `R_DrawParticles()` and deleted my config.cfg to be sure. Still rendering particles incorrectly with `gl1_pointparameters 0`.

ahluka:
My mistake, changing `gl1_pointparameters` to `0` does seem to work. I wasn't reloading the game between changes...

PR #482, opened 2019-11-05
Add force redraw on change gamma

0lvin:
Fixes bug with update only horizontal line with brightness slider in menu.


Yamagi:
And merged. Thank you! :)

PR #492, opened 2019-12-05
Added square particles cvar for GL 1.4 renderer

DenisBelmondo:
I would like to add a GL 1.4 equivalent of the GL 3.2 CVar "gl3_particles_square". Not sure why the diffs are... like that.

PR #493, opened 2019-12-05
Added square particles CVar for GL 1.4

DenisBelmondo:
This is the proper, fixed version of #492. Thanks for alerting me as to the reason why the diffs were like that, still trying to find my way around git (or source control utils in general lol).

### DIFF
--- a/src/client/refresh/gl1/gl1_main.c
+++ b/src/client/refresh/gl1/gl1_main.c
@@ -1671,22 +1671,8 @@ RI_BeginFrame(float camera_separation)
 
 	if (gl1_particle_square->modified)
 	{
+		R_InitParticleTexture();
 		gl1_particle_square->modified = false;
-
-		/* yamagi: GL_POINT_SMOOTH is not implemented by some
-		   OpenGL drivers, especially the crappy Mesa3D backends
-		   like i915.so. That the points are squares and not
-		   circles is not a problem by Quake II! */
-
-		switch ((int)gl1_particle_square->value)
-		{
-			default:
-				glDisable(GL_POINT_SMOOTH);
-				break;
-			case 0:
-				glEnable(GL_POINT_SMOOTH);
-				break;
-		}
 	}
 
 	/* draw buffer stuff */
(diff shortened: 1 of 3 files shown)


src/client/refresh/gl1/gl1_main.c
```diff
@@ -1667,6 +1669,26 @@ RI_BeginFrame(float camera_separation)
...
+		switch ((int)gl1_particle_square->value)
+		{
+			default:
+				glDisable(GL_POINT_SMOOTH);
+				break;
+			case 0:
+				glEnable(GL_POINT_SMOOTH);
+				break;
+		}
+	}
```
Yamagi: This block must be wrapped with `if (gl_config.pointparameters)`. Without it GPUs that don't support the pointparameter extension may run into problem. At `gl1_pointparameter 0` might be broken.

Yamagi:
Hi,
this is a good idea, but it doesn't work for me. No matter if I set `gl1_particle_square` to `1` or `0`, the particles are always rendered as dots and not squares. This is a Geforce 750TI with driver 440.31.

And if you're already on it... ;) In b7bf822e6d985c1fe1c88ae3355ea646f5561e4e we've changed the particles for the `gl1_pointparameter` case from squares to octagons. Maybe it would be a good idea to implement `gl1_particle_square` there as well?

src/client/refresh/gl1/gl1_main.c
DanielGibson: also, why a switch for just two cases (instead of if/else)?

DenisBelmondo:
Interesting, I'll definitely investigate today and also update my branch to include the if statement you mentioned earlier. Thanks for letting me know.

DenisBelmondo:
Hi all, mistakes were made along the way, but I think I got it figured out. I am hoping that the cards that don't support point parameters will enjoy the little square pixel buffer I added right below the dot one :). Unfortunately with my current setup, I cannot verify if that fallback works, but I am very sure the logic checks out.

Yamagi:
Hi,
thank you for the corrections. The normal case with `gl1_pointparameters 1` looks good and I'm fine with it. But there's still something wrong with `gl1_pointparameters 0` (requires `vid_restart`). With a very first sight on the code I don't see what could be the problem.

DenisBelmondo:
Interesting. I'll see if I can do some more investigation soon. Thanks for letting me know.

Yamagi:
I've merged this by hand. And added a commit that fixes the `gl1_pointparameters 0` case.

DenisBelmondo:
Awesome, thanks! Got bogged down with work and holidays recently so I didn't get a chance to look at it very much. Glad to hear it's in now :)

Issue #491, opened 2019-12-04
HUD elements ignore GL_TEXTUREMODE when using OpenGL 3.2


DenisBelmondo:
HUD elements don't seem to respect the gl_texturemode cvar when using OpenGL 3.2. May be related to #489?

**Reproduction steps:**
Open the video menu, set the renderer to Open GL 3.2 and apply. Then, open the console and set gl_texturemode to "GL_NEAREST", exit the game and start it up again (since #489 says that vid_restart may not work), and observe the HUD elements as they are still drawn with linear filtering.

DanielGibson:
I guess this is a feature, already exists in the original GL1 code (or at least our GL1 code).
2D elements and skies are ignored when changing gl_texturemode cvar - no idea why.
(As you have probably observed, changing the cvar changes the filtering of world and model textures immediately, without vid_restart or anything)

However, there definitely is a bug with the default filter modes not being set from the cvar when starting the GL3 renderer, I'll have to fix that.

No idea about 2D elements ("pics") and skies, we'll have to do some digging and try to guess why they're explicitly ignored in R_TextureMode()/GL3_TextureMode().

DanielGibson:
Looks like this behavior (pic and skies being ignored for texturemode) was indeed already present in vanilla Q2: https://github.com/id-Software/Quake-2/blob/master/ref_gl/gl_image.c#L226

"// change all the existing mipmap texture objects" sounds like it's only supposed to done for textures with mipmaps - maybe they assumed the others wouldn't get filtered anyway?
And for GUI textures it was probably true: **We** added menu/HUD/... scaling, so before they would've been renderered in original size and needed no filter
But the sky.. while it's always rendered at the same distance from the player, so IMO it would make sense to set a filter?

HOWEVER, I just tried removing the check for `it_sky` and `it_pic` and then, when I change the filtering mode to one with "_MIPMAP_" in the name, the HUD and sky textures turn black. just GL_LINEAR and GL_NEAREST work though.
However, checking if a texture is pic/sky and in that case using `gl_filter_max` for `GL_TEXTURE_MIN_FILTER` (like it's also done ~100 lines below in `GL3_Upload32()`) fixes that.

So anyway, I'll clean up my fixed code later and commit it.

DanielGibson:
suggested fix in #494

DenisBelmondo:
[...]

PR #494, opened 2019-12-08
Use gl_texturemode for UI and skies as well, fixes #489 and #491

### LINKED ISSUE #489: GL_TEXTUREMODE ignored on vid_restart
If GL_TEXTUREMODE is set for instance to GL_NEAREST_MIPMAP_NEAREST and you execute vid_restart, textures are filtered as if GL_LINEAR_MIPMAP_LINEAR was the value set.

If you close the client and run it again, the correct option is used again.

DanielGibson:
and make sure that after vid_restart (or starting the game) it's used
correctly in GL3.

While at it, made sure that it's *not* applied to textures from
gl_nolerp_list, because they're supposed to always use GL_NEAREST
independent of this setting (used so console font and crosshairs don't
look blurry)

### DIFF
--- a/src/client/refresh/gl1/gl1_draw.c
+++ b/src/client/refresh/gl1/gl1_draw.c
@@ -501,8 +501,10 @@ RDraw_StretchRaw(int x, int y, int w, int h, int cols, int rows, byte *data)
 				0, GL_COLOR_INDEX, GL_UNSIGNED_BYTE, image8);
 	}
 
-	glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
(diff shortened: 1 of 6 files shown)


Yamagi:
Windows binaries can be found here: https://deponie.yamagi.org/quake2/windows/testbuilds/quake2-20191209-1077b75-texturemode_fix.zip

Yamagi:
Can we merge this or are we still waiting for feedback??

DanielGibson:
there seems to be minor issues with it (when switching from GL3->GL1->GL3), also it's not used for cinematics like the end screen: https://github.com/yquake2/yquake2/issues/491#issuecomment-563940141

I'll have a look at it when I find some time

Issue #495, opened 2019-12-29
Script issue in Torture Chambers


some0nebad:
Well, I think this is a script issue.

In this room, after death of this marine, a key should fall out. But in the latest version (7.42), it does not appear if you kill him by yourself. Key will appear only after the strogg presses the button and you do not interfere. Version 7.30 does not have this issue.

Windows 7 x64
[![yq2](https://user-images.githubusercontent.com/52467651/71558518-220fb000-2a5d-11ea-93fc-e32bcbc6b4c7.jpg)
](url)

some0nebad:
And you can press a button and wait for his death, and key will appear. But if you shoot him it will not

Yamagi:
Hi,
thank you for the report. I'll look into it, but it'll take a few days.

DanielGibson:
@BjossiAlfreds: We haven't investigated this, but as most recent changes to the game code have been yours, it might make sense if you look at it (if you have some time)?

BjossiAlfreds:
Sure, I'll help you guys out and look into it as well.

BjossiAlfreds:
I looked into the jail4.bsp entity string. The misc_insane guy has a target t130, which means upon death he should fire target t130 and trigger all entities with targetname t130, including a target_spawner which spawns key_pass.

I have tested in-game and I can confirm there is something wrong and targets aren't being fired like they should. I'll take a closer look at the code and try to see what is causing this, there is a strong probability one of my many bugfixes broke this...

DanielGibson:
Thanks for debugging this! :)

I guess we can try to find another struct member and try to make double-sure it's not otherwise used by the insanes - otherwise we'll have to live with the pain sounds getting cut off.

BjossiAlfreds:
I went shopping at local.h and I have a candidate - fly_sound_debounce_time. I did a full reference check and it is only used in g_trigger by trigger_push. I would of course add a comment to make sure we don't forget about this hack.

BjossiAlfreds:
I have made PRs for yq2, rogue and xatrix that fix this bug. Since it can be game-breaking it may be a good idea to update the currently released 7.42 binaries.

Yamagi:
I just merged the PRs. I'll build a new set of Windows binaries tonight or tomorrow.

PR #496, opened 2020-01-02
Fix for insane marines not firing deathtarget right away

BjossiAlfreds:
Fixes bug https://github.com/yquake2/yquake2/issues/495. As detailed in the comments, this was caused by using the delay struct member for storing data specific to the insane-marines, when in fact it is a generally used value by G_UseTargets.

I fixed it by using a different struct member, fly_sound_debounce_time, this time making absolutely sure there won't be any interference. This struct member is only stored and referenced in one place in the code and only by clients, so interference by the insane marines code is impossible. I added a comment in local.h to explicitly state this new use case for the struct member.

### DIFF
--- a/src/game/header/local.h
+++ b/src/game/header/local.h
@@ -1015,7 +1015,7 @@ struct edict_s
 	float touch_debounce_time;
 	float pain_debounce_time;
 	float damage_debounce_time;
-	float fly_sound_debounce_time;
+	float fly_sound_debounce_time;	/* now also used by insane marines to store pain sound timeout */
 	float last_move_time;
 
 	int health;
--- a/src/game/monster/insane/insane.c
+++ b/src/game/monster/insane/insane.c
@@ -65,7 +65,7 @@ insane_moan(edict_t *self)
 	}
 
 	/* suppress screaming so pain sound can play */
-	if (self->delay > level.time)
+	if (self->fly_sound_debounce_time > level.time)
 	{
 		return;
 	}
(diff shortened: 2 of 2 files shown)


Yamagi:
I think that this is correct and doesn't break other cases. @DanielGibson, what do you think?

src/game/header/local.h
```diff
@@ -1015,7 +1015,7 @@ struct edict_s
 	float touch_debounce_time;
 	float pain_debounce_time;
 	float damage_debounce_time;
-	float fly_sound_debounce_time;
+	float fly_sound_debounce_time;	/* now used by insane marines to store pain timeout */
```
DanielGibson: maybe "now *also* used..." so it's clear that the originally intended usage also still exists?

BjossiAlfreds: Good call, I'll amend that.

BjossiAlfreds: I've updated the comment for all 3 PRs now (yq2, rogue and xatrix).

DanielGibson:
I think this is a good solution - and even if it causes some kind of conflict after all (maybe if an insane is pushed and has pain at the same time or sth?) at least both usages are just sounds so it shouldn't break the gameplay.

BjossiAlfreds:
Only players use the fly sound variable when pushed: "if (other->client) ...", so the two use cases should be mutually exclusive, if I'm reading the code correctly at least.

PR #498, opened 2020-01-05
It is now possible to build a docker dedicated server.

Opvolger:
Yes I know, first ask if the code is welcome. But I wanted to run it on my NAS (with docker). So otherwise I would have made it too.

This allows a docker image to be created

this can auto build in docker hub:
https://hub.docker.com/r/opvolger/yquake2

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+FROM ubuntu:18.04 AS build
+
+WORKDIR /yquake2
+RUN apt-get update
+RUN apt-get install libcurl4-openssl-dev build-essential -y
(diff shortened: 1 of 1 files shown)


Yamagi:
What's the difference to `make server game`?

Opvolger:
I am not so familiar with makefiles yet. But I now see what you mean. I adjusted it

Yamagi:
I've put the Dockerfile in stuff/. I know, Dockerfiles are put under the root dir, but that's already cluttered enough and Docker isn't a feature that users use everyday.

Issue #373, opened 2019-02-13
vkQuake2 Vulkan render support?


shoober420:
I just came across this vulkan source port of Quake 2.

https://github.com/kondrak/vkQuake2

Is there a way that the Vulkan render from this port can be used with Yamagi?

kondrak:
I'm not really familiar with Yamagi port but I'll try to add my 2 cents here. :)
According to one of the devs on Twitter, it should be "fairly easy" to add the new renderer - there are some differences in the initialization process but overall, the entire backend would likely remain untouched. The "Vulkan part" of vkQuake2 is agnostic enough so it can be pulled out rather easily with some minor changes (I introduced some dependencies on the internal Quake 2 types but nothing really critical). 
With that said - there's really no added value introduced by Vulkan rendered in this case, other than being able to say that Yamagi supports Vulkan now - performance gains will be insignificant in comparison.

Once I clean up all the rough edges in vkQuake2, I might be considering helping out on this one since Yamagi seems to be one of the best Quake 2 source ports out there. Can't really give any specific timeline for this though, with it being just a side-project and all.

Yamagi:
Some years ago when we've added the GL3 renderer we've designed a completely new renderer API. That renderer API is much cleaner than the original one and should be - now that we've gone through 2 or 3 iterations of improvements - fairly stable. So it shouldn't be a be a big problem to support "inofficial" renderers. 

The client side entry point is here: https://github.com/yquake2/yquake2/blob/master/src/client/vid/vid.c#L332

That said, I'm always happy merging new features. :) But only if there's someone volunteering as a maintainer, at least until the feature as working as expected and doesn't have too many bugs. A good example is the softrenderer which was contributed by @0lvin. He did the initial work, solved many bugs and problems and is now getting to the point that the code is some kind of maintenance mode where it's just working.

That said,

DanielGibson:
Thanks for chiming in, Krzysztof! :-)

The DLL interface (structs passed between client and renderer) is at https://github.com/yquake2/yquake2/blob/master/src/client/vid/header/ref.h#L124

> He did the initial work, solved many bugs and problems and is now getting to the point that the code is some kind of maintenance mode where it's just working.

But still occasionally bugs turn up and *he* fixes them. If he stopped doing that and there would still be too many issues we'd probably remove the soft renderer again, as we don't want to spend much time on maintaining it ourselves.

[...]

PR #500, opened 2020-01-11
Add Vulkan render based on vkQuake2

0lvin:
* #373 Add Vulkan code based on [vkQuake2](https://github.com/kondrak/vkQuake2/releases/tag/1.4.3) commit  9c85ad15a73320be6641c840697644d27e851fd4
* have made really small changes for sync codebase with other renders
* checked only on linux and disabled by default


0lvin:
Big thanks to @kondrak.

kondrak:
Oh wow, I did not expect that coming! Excellent work! 👍

kondrak:
Regarding R_EndWorldRenderpass - in vkQuake2, each time a popup appeared in the menu (when pinging servers, trying to refresh lists etc.) it caused the OpenGL path to exit prematurely - this didn't work with Vulkan, since the exit path did not submit all the data correctly, hence that function was introduced.

kondrak:
As for Vulkan memory allocation, passing `-std=c++11` to gcc flags should be enough to make it work!

kondrak:
As for Vulkan memory allocation, passing `-std=c++11` to gcc flags should be enough to make it work!

Yamagi:
Good morning,
it's nice to wake up and find something like this in the mailbox. Of course I haven't had a chance to take a deeper look, I just played through base1. First let me make clear, that I'm **not** against merging this. But before we can consider that some things needs to be ensured:

* You've already linked to #373. @DanielGibson already wrote there, that neither he nor me have any knowledge about Vulkan. While it would be a nice project to read into it, hack a little bit around, etc. we currently don't have the time to do that. We aren't students anymore, we've got daytime jobs, families and other projects. Therefore it's inevitable that someone with Vulkan knowledge is willing to maintain this renderer. Work through bug report, analyse and fix the bugs, chase changes in GPU driver and the Vulkan SDK if necessary... More or less the same as we did for the software renderer.
* There're likely some things that needs to be ruled out before we can merge it. It needs documentation, compiling the code with gcc9 gave some warnings, the gamma scale is the wrong way round (in the OpenGL renderers higher values are brighter, with the Vulkan renderer higher values are darker), typing `disconnect` into the console freezes the renderer... That are all fixable problems, but they need to be solved.

On a bigger scale the big question is if Vulkan is really the way to go or if it would be better to use something like https://github.com/bkaradzic/bgfx. With that we would get Vulkan, DirectX, Metal and WebGL by doing all the work involved with adding and maintaining a renderer only once.

0lvin:
I can try fix errors and some simple bugs, but i am not Vulkan expert

kondrak:
Now that the renderer is basically "done" in here, I think I could spare some time to help out with Vulkan bugs if they surface. The renderer in vkQuake2 is pretty much complete, not much has changed since the initial release (not counting bugfixes), so the behavior in yQuake2 should not differ much from the original.

turol:
I know Vulkan but I don't have much time. If someone wants to ask questions then I can answer them but probably can't do much hacking on yquake2 myself.

Yamagi:
So many helping hands. That's nice to see. :) What about something like this:

1. Fix the gamma (likely something like `vid_gamma->value * -1` missing somewhere), fix the renderer hanging after disconnect (start a game with `map base1`, disconnect from it with `disconnect` and the renderer freezes) and update this pull request with the changes.
2. We merge the renderer into a feature branch and provide official test build. The test builds will be linked on the homepage. Fix upcoming bugs (I don't think that there will be many since the major part code was already tested in vkQ2), enhance the renderer API as needed (there aren't any out of tree consumers anyways) and write documentation.
3. After 3 to 6 month, when it's ready, it's merged into master and released as YQ2 8.0.

kondrak:
Sounds good to me!

Pickle:
Are there any performance improvements with this renderer?

Yamagi:
I think that'll depend on the GPU driver. GPUs with crappy OpenGL drivers - for example the Windows driver for GCN Radeons - will likely see an improvement. GPUs with good OpenGL drivers won't, because they were already running at the hardware limit or YQ2s internal 999 FPS cap.

kondrak:
Like Yamagi said, performance gains are impossible to measure here due to original 1000 FPS cap. In vkQuake2, the difference between OpenGL and Vulkan was substantial (though impossible to notice by the player on modern hardware), in yquake2 I expect there to be no difference at all.

kondrak:
btw, @0lvin are you going to address the first two issues (gamma and renderer hang) yourself?

0lvin:
@kondrak I can look only over the weekend. If you could look to it, i will merge you changes.

0lvin:
@kondrak  Also if you have time - could you look to retexturing code? It has used scaled images instead original size - that make it useless in case when user provided hd images as it will change only colors of image instead use high definition version of images.

kondrak:
I think it's best to focus on getting this to feature branch, so 1 and 2 first and then we can start collecting more specific issues to fix. This way it will be easier for me personally to allocate time to fix/improve things.

0lvin:
I have fixed `gamma` and `disconnect` command.

Yamagi:
Cool, thank you. It looks much better than before. As promised I have merged this into a feature branch named `ref_vk`: https://github.com/yquake2/yquake2/tree/ref_vk

I'm closing here. An issue to discuss and coordinate this effort can be found here: https://github.com/yquake2/yquake2/issues/505 And I've added a new Label *vulkan* to mark ref_vk related issues.

Please open new pull requests against *rev_vk*, not *master*. :)

0lvin:
Thank you

PR #503, opened 2020-01-19
Easier handling of skill levels by using defines instead of numbers.

NeonKnightOA:


Yamagi:
This is a very good idea and looks perfect. I'm going to merge it. But (no kind words without a but ;)) can you do the same for the rogue and xatrix addons? It shouldn't be too hard and we try to keep the addons in sync with baseq2 when doing cleanups and bugfixes.

NeonKnightOA:
There. I did adaptations for rogue, xatrix and zaero.

Yamagi:
Cool, that's very good work!

PR #510, opened 2020-01-26
Rearrange code in ref_vk

0lvin:
* add cvar documentation
* update stb_* codes
* add source for shaders
* little bit code cleanup


Yamagi:
Aaand merged :)

It might be a good idea to merge stuff that isn't specific to the Vulkan renderer into master and not inot the feature branch. Otherwise they're stuck in the feature branch until we merge it.

PR #511, opened 2020-01-30
Small fixes

0lvin:
* port not specific to Vulkan changes from ref_vk branch
* fix segmentation fault on device without sound card, ogg_shuffle is not initialized.
* update stb_* codes
* use snprintf with buffers


Yamagi:
As always: Thank you! :)

 It even solves some of the compiler warnings in the stb headers.

PR #515, opened 2020-02-08
Hunk API, using FreeBSD super page for large requests.

devnexen:


Yamagi:
I don't think that using superpages will make any performance difference, but nevertheless: Thank you!

PR #519, opened 2020-02-16
Fix open multiplayer

0lvin:
* gamma of 3d surfaces changed without restart render #505
* make more functions static
* fix popup menu

### DIFF
--- a/src/client/cl_screen.c
+++ b/src/client/cl_screen.c
@@ -1519,7 +1519,7 @@ SCR_UpdateScreen(void)
 		numframes = 2;
 		separation[0] = -gl1_stereo_separation->value / 2;
 		separation[1] = +gl1_stereo_separation->value / 2;
-	}		
+	}
 	else
 	{
 		separation[0] = 0;
@@ -1571,6 +1571,7 @@ SCR_UpdateScreen(void)
 					R_SetPalette(NULL);
 					cl.cinematicpalette_active = false;
 				}
+
 				R_EndWorldRenderpass();
 				SCR_DrawConsole();
 			}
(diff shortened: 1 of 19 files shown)


Yamagi:
Hi,
thank you for your continued efforts. Let's start with some general comments:

* Reviewing is much easier if the pull request / review request has only the relevant commits and not unrelated stuff. It would be really helpfull if rebasing / merging commits from master into the feature branch and actual changes can be kept separate.
* From my experience it ss a good idea to avoid unnecessary changes to the upstream code, it just makes merging future changes harder.
* Does the Vulkan renderer really need the texture resizing stuff? Can't the textures just uploaded as they are? If I remember right the resizing code was only to support very early GPUs that required power of two textures. And is the STB resize code suited for Quake IIs low res textured, doesn't it change the look on such old GPUs if the textures are resized with a better algorithm? Doesn't a better algorithm require more CPU power?

Regarding the menu pop-ups: It's working now, but is it really a good idea to enter the renderer and abort the frame by returning early? Who not just adding a new function to the render API. There aren't any 3rd party renderers anyways, as far as I know.

0lvin:
Thank you!

> Reviewing is much easier if the pull request / review request has only the relevant commits and not unrelated stuff. It would be really helpfull if rebasing / merging commits from master into the feature branch and actual changes can be kept separate.

I have remove `master` branch merge.

> From my experience it is a good idea to avoid unnecessary changes to the upstream code, it just makes merging future changes harder.

I have moved changes non vulkan changes to #521

Yamagi:
> Yes, its limitation current code from vkQuake2.

I don't know nothing about Vulkan but it looks like something worth investigating.

Nevertheless, I'm merging this now. And I'm going to merge master into ref_vk in a few minutes.

0lvin:
thank you

PR #520, opened 2020-02-20
Allow loading ZIP files with a `.pkz` extension

Calinou:
This makes it possible to distribute add-ons that are compatible with both Yamagi Quake 2 and Q2PRO. (Q2PRO only supports `.pkz` as a file extension for ZIP files.)

See https://github.com/Calinou/quake2-neural-upscale/issues/2 :slightly_smiling_face:


Yamagi:
Merged by hand in 9160af7. Thank you!

PR #521, opened 2020-02-23
Small cleanup

0lvin:
* small cleanup in soft render for make cppcheck happy
* reuse resize code in gl1 render

### DIFF
--- a/src/client/refresh/gl1/gl1_image.c
+++ b/src/client/refresh/gl1/gl1_image.c
@@ -409,53 +409,6 @@ R_FloodFillSkin(byte *skin, int skinwidth, int skinheight)
 	}
 }
 
-void
-R_ResampleTexture(unsigned *in, int inwidth, int inheight,
-		unsigned *out, int outwidth, int outheight)
-{
-	int i, j;
-	unsigned *inrow, *inrow2;
-	unsigned frac, fracstep;
-	unsigned p1[1024], p2[1024];
-	byte *pix1, *pix2, *pix3, *pix4;
-
-	fracstep = inwidth * 0x10000 / outwidth;
-
-	frac = fracstep >> 2;
-
-	for (i = 0; i < outwidth; i++)
-	{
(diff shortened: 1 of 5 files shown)


Yamagi:
Was the ResizeSTB() stuff tested?

0lvin:
Yes, it used in soft render already.

I have checked it with vulkan and gl1 render - and have not found any visible issues.

0lvin:
@Yamagi What do you think about process all new code by [indent](https://www.gnu.org/software/indent/manual/html_node/indent_4.html#SEC4) to have common codestyle?

DanielGibson:
> What do you think about process all new code by indent to have common codestyle?

No. That'd screw up `git blame` (you'd get the reformat commit instead of the commit introducing the "real" code change)

0lvin:
> No. That'd screw up git blame (you'd get the reformat commit instead of the commit introducing the "real" code change)

I mean only new code like vulkan render, where we have only several contributors and real author/commit of code lines before merge to master(stabilize) does not matter.

0lvin:
@Yamagi @DanielGibson Do you have some preferences for code style for **_ fully new code_**? Or style like **_linux kernel code_** from ident will be fine?

Yamagi:
When I started YQ2 I put all code through uncrustify, the config is still here: https://github.com/yquake2/yquake2/blob/master/stuff/misc/uncrustify.cfg But in retroperspective I wouldn't do it again, I would leave the code indention and the file layout as they were. Merging is much, much easier when there aren't as little divergences as possible. I was young, inexperienced and didn't know that.

0lvin:
thank you

Yamagi:
As style for new code: I'm more or less following FreeBSDs code style, https://www.freebsd.org/cgi/man.cgi?query=style&apropos=0&sektion=9&manpath=FreeBSD+12.1-RELEASE+and+Ports&arch=default&format=html @DanielGibson does his own things. So something like the Linux kernel style should be okay.

PR #523, opened 2020-02-27
Make build system more distro-friendly

smcv:
* build: Let users pass in their own CFLAGS, CPPFLAGS, LDFLAGS
    
    There is a widely-used convention that CFLAGS, CPPFLAGS and LDFLAGS are
    set by the user (or packaging system) when building software,
    defaulting to empty: in Autotools terminology they are said to be
    "reserved for the user". Linux distributions use this to provide
    globally-used compiler flags, for example enabling enabling various
    hardening features.
    
    These are conventionally placed on the compiler command-line *after* the
    build system's choice of CFLAGS and LDFLAGS, so that they can override
    the build system's defaults. For example, when the Debian packaging
    toolchain is told to build a package without optimization for debugging,
    it includes -O0 in CFLAGS. This would not be effective in previous
    versions of yquake2, which hard-code -O2.
    
    Loosely based on a patch by Fabian Greffrath, which we have been applying
    in Debian for several years.

* build: Add an option to turn off RPATHs
    
    When distributions like Debian build binaries that are going to use the
    standard system-wide libraries in directories that are searched by
    default, it's undesired to have a RPATH that makes the library search
    behaviour less predictable. Continue to set the RPATH by default, for
    use in portable binaries, but provide an off-switch.
    
    Based on a patch by Fabian Greffrath, which we have been applying in
    Debian for several years.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,11 +132,11 @@ endif
 # -fwrapv for defined integer wrapping. MSVC6 did this
 #  and the game code requires it.
 ifeq ($(YQ2_OSTYPE), Darwin)
-CFLAGS := -O2 -fno-strict-aliasing -fomit-frame-pointer \
+YQ2_CFLAGS := -O2 -fno-strict-aliasing -fomit-frame-pointer \
 		  -Wall -pipe -g -fwrapv
-CFLAGS += $(OSX_ARCH)
+YQ2_CFLAGS += $(OSX_ARCH)
 else
-CFLAGS := -std=gnu99 -O2 -fno-strict-aliasing \
+YQ2_CFLAGS := -std=gnu99 -O2 -fno-strict-aliasing \
 		  -Wall -pipe -g -ggdb -MMD -fwrapv
 endif
 
@@ -146,27 +146,27 @@ endif
 ifeq ($(COMPILER), clang)
 	# -Wno-missing-braces because otherwise clang complains
 	#  about totally valid 'vec3_t bla = {0}' constructs.
-	CFLAGS += -Wno-missing-braces
+	YQ2_CFLAGS += -Wno-missing-braces
 else ifeq ($(COMPILER), gcc)
 	# GCC 8.0 or higher.
 	ifeq ($(shell test $(COMPILERVER) -ge 80000; echo $$?),0)
 	    # -Wno-format-truncation and -Wno-format-overflow
 		# because GCC spams about 50 false positives.
-    	CFLAGS += -Wno-format-truncation -Wno-format-overflow
+	YQ2_CFLAGS += -Wno-format-truncation -Wno-format-overflow
 	endif
 endif
 
 # ----------
 
 # Defines the operating system and architecture
-CFLAGS += -DYQ2OSTYPE=\"$(YQ2_OSTYPE)\" -DYQ2ARCH=\"$(YQ2_ARCH)\"
+YQ2_CFLAGS += -DYQ2OSTYPE=\"$(YQ2_OSTYPE)\" -DYQ2ARCH=\"$(YQ2_ARCH)\"
 
 # ----------
 
 # Fore reproduceable builds, look here for details:
 # https://reproducible-builds.org/specs/source-date-epoch/
 ifdef SOURCE_DATE_EPOCH
-CFLAGS += -DBUILD_DATE=\"$(shell date --utc --date="@${SOURCE_DATE_EPOCH}" +"%b %_d %Y" | sed -e 's/ /\\ /g')\"
+YQ2_CFLAGS += -DBUILD_DATE=\"$(shell date --utc --date="@${SOURCE_DATE_EPOCH}" +"%b %_d %Y" | sed -e 's/ /\\ /g')\"
 endif
 
 # ----------
@@ -176,23 +176,23 @@ endif
 # Would be nice if Clang had something comparable.
 ifeq ($(YQ2_ARCH), i386)
 ifeq ($(COMPILER), gcc)
-CFLAGS += -ffloat-store
+YQ2_CFLAGS += -ffloat-store
 endif
 endif
 
 # Force SSE math on x86_64. All sane compilers should do this
 # anyway, just to protect us from broken Linux distros.
 ifeq ($(YQ2_ARCH), x86_64)
-CFLAGS += -mfpmath=sse
+YQ2_CFLAGS += -mfpmath=sse
 endif
 
 # ----------
 
 # Systemwide installation.
 ifeq ($(WITH_SYSTEMWIDE),yes)
-CFLAGS += -DSYSTEMWIDE
+YQ2_CFLAGS += -DSYSTEMWIDE
 ifneq ($(WITH_SYSTEMDIR),"")
-CFLAGS += -DSYSTEMDIR=\"$(WITH_SYSTEMDIR)\"
+YQ2_CFLAGS += -DSYSTEMDIR=\"$(WITH_SYSTEMDIR)\"
 endif
 endif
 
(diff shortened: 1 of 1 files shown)


smcv:
> Loosely based on a patch by Fabian Greffrath, which we have been applying in Debian for several years.

In the previous patch we changed all the `CFLAGS :=` to `CFLAGS +=`, but I think this one is better, because it gets the precedence right: if yquake2's default is `-O2`, and the user or packaging system that is installing it asks for `-O0` or `-Og` because they are trying to debug it, we want the `-O0` to win.

DanielGibson:
I'm a bit afraid of letting the user set (systemwide, default?) compilerflags - if they feel clever and set `-O3` or even `-ffast-math` shit might break.

Also, looking at the changed code, it seems like `CFLAGS` is not used at all but has been replaced with `YQ2_CFLAGS` - am I missing something or does this mean that in fact the users CFLAGS don't have any effect?

smcv:
> Also, looking at the changed code, it seems like CFLAGS is not used at all but has been replaced with YQ2_CFLAGS - am I missing something or does this mean that in fact the users CFLAGS don't have any effect?

I think you didn't spot the bit where `CFLAGS` and `CPPFLAGS` are appended to `YQ2_CFLAGS`, and `CFLAGS` and `LDFLAGS` are appended to `YQ2_LDFLAGS` (see the nearby comment for why it isn't the obvious 1:1 mapping).

> I'm a bit afraid of letting the user set (systemwide, default?) compilerflags - if they feel clever and set -O3 or even -ffast-math shit might break.

If people are feeling clever, they're going to break it anyway. This setup matches how the compiler flags are normally handled in Autotools (always), Meson (always), and CMake (with `-DCMAKE_BUILD_TYPE=none`).

The typical use of this in distributions is for packaging infrastructure (`dpkg-buildflags` or RPM macros) to build everything with debug symbols (`-g`), a conservative optimization level (`-O2` or `-O0` depending whether an unoptimized build was requested), and add security-hardening options like stack-protector and PIE.

Would it help if I wrapped it like this?

```
ifeq ($(WITH_COMPILER_FLAGS), yes)
# Take compiler flags from the environment. If you enable this, any broken
# builds that result are entirely your responsibility.
YQ2_CFLAGS += $(CPPFLAGS) $(CFLAGS)
(etc.)
endif
```

Or I could re-word the commit message and/or comments to phrase it in terms of distro compiler flags?

smcv:
If people use `-ffast-math` and it breaks, that's because `-ffast-math` violates the C standard - which obviously is a bad idea to do in codebases that aren't expecting it. I realise the name of the option is rather a trap, because it sounds like such a good thing ("of *course* I want mathematics to be fast, this is a game engine!") - so perhaps something like this would make sense?

```
#ifdef __FAST_MATH__
#error gcc -ffast-math does not follow IEEE/ISO specifications and is not supported
#endif
```

Yamagi:
Hi,
I've just pushed several commit into the branch *cflags*. They:

* Make the base CFLAGS and LDFLAGS overrideable
* Make sure that required compiler and linker flags are passed
* Some general cleanup, mostly for Windows

They're here: https://github.com/yquake2/yquake2/commits/cflags

Instead of renaming CFLAGS and LDFLAGS and appending the environment to them, both base CFLAGS and LDFLAGS can be overridden by the environment. They're the first flags passed to the compiler and linker, thus they shouldn't be able to override required options. At least clang on FreeBSD is last option serves.

smcv:
Thanks, I think that does most of what I need. I'll try it when I get a chance.

The one thing that's definitely missing compared with this PR is `WITH_RPATH`, so I'll rebase that on top of your `cflags` branch.

> both base CFLAGS and LDFLAGS can be overridden by the environment. They're the first flags passed to the compiler and linker, thus they shouldn't be able to override required options.

I think that's correct on all reasonable platforms, yes.

Yamagi:
I've pushed a commit that adds the `WITH_RPATH` option back. If it's set to *no* the RPATH to $ORIGIN/lib is ommitted and `WITH_SYSTEMWIDE` cannot be set to *yes*.

smcv:
> If it's set to no the RPATH to $ORIGIN/lib is ommitted

Great!

> and WITH_SYSTEMWIDE cannot be set to yes

Is there a reason why not? WITH_SYSTEMWIDE but no RPATH is the configuration we use in Debian, and if the Debian package doesn't count as a system-wide installation, I don't know what is.

We put the `quake2` and `q2ded` executables in `/usr/lib/yamagi-quake2`, the native-code (architecture-dependent) game modules in `/usr/lib/yamagi-quake2/baseq2/game.so` and so on, and the architecture-independent assets (pak files, videos, music) in `/usr/share/games/quake2`. I think that's in line with what you recommend in 070_packaging.md?

Unless I'm missing some secondary function, the RPATH is only needed if you are linking the game executable to shared libraries (SDL, libcurl etc.) in a non-standard location, for example bundled with yquake2. That's likely to be something you'd want to do if you are shipping a generic Linux binary designed to be portable between distributions, but the Debian package only needs to run on Debian, and gets all its shared libraries from the default library search path.

If I understand correctly, `game.so` is searched in the game search path (which includes the directory the executable is in) and passed to dlopen as an absolute path, and the refresher/renderer modules like `ref_gl3.so` are searched in the executable's directory and passed to dlopen as an absolute path, so it's unnecessary for the directory that contains them to appear in the RPATH (unless it also contains bundled libraries, which, again, distro packages don't).

smcv:
> WITH_SYSTEMWIDE but no RPATH is the configuration we use in Debian

More specifically, WITH_SYSTEMWIDE has two effects: it adds the RPATH ($(WITH_SYSTEMDIR)/lib or /usr/share/games/quake2/lib) for shared libraries, and it defines -DSYSTEMWIDE, which makes the engine look in SYSTEMDIR or /usr/share/games/quake2 for assets.

In the Debian package we want to get the assets from their system-wide location, but the RPATH is unnecessary and doesn't really make sense (we wouldn't install architecture-dependent native-code under /usr/share anyway).

Yamagi:
You're right. We don't need the RPATH for the systemwide installation. I don't even remember why we ever set it... I've force-pushed an update that restores WITH_SYSTEMWIDE when WITH_RPATH is set to *no* and removes the systemwide RPATH altogether.

smcv:
I've tried applying the `cflags` branch to the yquake2 Debian package.

The general technique works fine for the pattern where you build like this:

```
CFLAGS="-Og -g" LDFLAGS="-Wl,--as-needed" make
```

or equivalently

```
export CFLAGS="-Og -g" LDFLAGS="-Wl,--as-needed"
make
```

However, it doesn't work for the case where you pass these variables as arguments to `make`, which is also a fairly common approach:

```
make CFLAGS="-Og -g" LDFLAGS="-Wl,--as-needed"
```

because in that case, the variables are completely overridden, ignoring any `+=` (at least, that's the case with GNU make - I haven't tried *BSD make).

If you want that last pattern to work, I think the only alternative to using separate variables would be to use the syntax `override CFLAGS += -fwrapv` for the things you need to force, but I think that might be a feature that only exists in GNU make. Is the Makefile intended to be portable to other make implementations, or is it OK to use GNU-specific features? 020_installation.md describes it as a "GNU Makefile", but nothing actually seems to say that GNU make is required.

smcv:
Another issue with the `cflags` branch is that it adds `-std=c99`. In gcc and glibc, this is interpreted as a request to act as a very strict C99 compiler, disabling many GNU extensions - in particular, `readlink()` isn't declared, and inline assembly in `asm()` isn't recognised (although I think `__asm__()`, which is in the namespace for system-specific extensions, still is).

One way to deal with this would be to use `-std=gnu99` (which is C99 with GNU extensions), or to not use a `-std` option at all (because the default in any vaguely recent gcc version is `gnu99` or newer). This is the approach I'd be inclined to use.

Another would be to get the `readlink` declaration by using feature test macros (Linux readlink(2) says defining `_XOPEN_SOURCE` to a value >= 500, or `_POSIX_C_SOURCE` to a value >= 200112L, should work; or you could define `_GNU_SOURCE` which enables all the extensions), and to use `__asm__` for inline assembly. That effectively means you're requesting a very strict C99 environment, and then making targeted exceptions for specific extensions beyond C99.

smcv:
> One way to deal with this would be to use -std=gnu99 (which is C99 with GNU extensions)

It used to be `-std=gnu99` before the `cflags` branch.

It does seem weird to be telling the compiler in the Makefile "please implement C99", and then ask it in `src/common/header/shared.h` whether it's implementing C11...

Just removing `-std` doesn't seem to work with my compiler (gcc 9.2.1): it defaults to C11 mode, but in C11 mode, in `src/common/header/shared.h`, because `__STDC_VERSION__` is defined to `201112L` or later, we define `YQ2_ATTR_NORETURN` as `_Noreturn`. Unfortunately, gcc doesn't seem to interpret `_Noreturn` as being valid to attach to function pointers (unlike its `__attribute__((__noreturn__))` extension), but there are a couple of function pointers tagged with `YQ2_ATTR_NORETURN`, causing the build to fail in C11 mode.

Yamagi:
> It used to be -std=gnu99 before the cflags branch.

Yeah, that was my fault. I'll force push the change back to gnu99 in a few minutes. On FreeBSD there's no difference in libc between c99 and gnu99, therefor I didn't notice it.

> Is the Makefile intended to be portable to other make implementations, or is it OK to use GNU-specific features? 020_installation.md describes it as a "GNU Makefile", but nothing actually seems to say that GNU make is required.

We've never targeted other make implementations besides GNU make and I don't think that the Makefile will work with anything else. I just gave it a try with FreeBSDs bmake (backward compatible to pmake) and there're 265 errors while parsing the Makefile... I've marked all global additions to the CFLAGS and LDFLAGS as `override`. And added INCLUDE to the list of overideable variables.

> It does seem weird to be telling the compiler in the Makefile "please implement C99", and then ask it in src/common/header/shared.h whether it's implementing C11...

Not really. :) We're building in C99 mode, because we want stay compatible with older compilers. And maybe one day we'll port the code MSVC, every feature from newer C standards will make that harder... On the other hand we're adding C11 pathes for everything that isn't covered in older standards, it'll make ports to other C11 compliant compilers with no or only a few GNU extensions easier

smcv:
Thanks, I'll try that when I have a chance.

> Just removing -std doesn't seem to work with my compiler (gcc 9.2.1): it defaults to C11 mode, but in C11 mode, in src/common/header/shared.h, because __STDC_VERSION__ is defined to 201112L or later, we define YQ2_ATTR_NORETURN as _Noreturn. Unfortunately, gcc doesn't seem to interpret _Noreturn as being valid to attach to function pointers (unlike its __attribute__((__noreturn__)) extension), but there are a couple of function pointers tagged with YQ2_ATTR_NORETURN, causing the build to fail in C11 mode.

I have a patch to separate `YQ2_ATTR_NORETURN` into `YQ2_ATTR_NORETURN` and `YQ2_ATTR_NORETURN_FUNCPTR` (with the latter only having an effect on gcc/clang, like the GLib change proposed on <https://gitlab.gnome.org/GNOME/glib/merge_requests/1078>) which I'll turn into a MR when I've rebased it onto your latest `cflags` work.

smcv:
The latest `cflags` branch seems to work well.

smcv:
I'm closing this, please merge `cflags` instead.

PR #528, opened 2020-03-08
Little ARM case update.

devnexen:
giving hint to cpu it can takes a fast coffee break
to cool down.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -165,9 +165,12 @@ if ("${ARCH}" STREQUAL "x86_64")
 	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -mfpmath=sse")
 endif()
 
-if (YQ2ARCHNATIVE AND "${ARCH}" STREQUAL "arm")
-	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -march=native")
-	add_definitions(-DYQ2ARCHNATIVE=1)
+if ("${ARCH}" STREQUAL "arm")
+    if (CMAKE_SIZEOF_VOID_P EQUAL 4)
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -march=armv6k")
+    else()
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -march=armv8.0-a")
+    endif()
 endif()
 
 set(Backends-Generic-Source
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -165,6 +165,10 @@ if ("${ARCH}" STREQUAL "x86_64")
 	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -mfpmath=sse")
 endif()
 
+if ("${ARCH}" STREQUAL "arm")
+	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -march=native")
+endif()
```
DanielGibson: Are you sure this is a good idea?
Can't this break binaries build for redistribution (e.g. in debian packages)?

Yamagi:
What about the Makefile build?

Yamagi:
Or another question: Is `-march=native` necessary to get the opcode, because gcc defaults to some crappy old ARM revision or is ist to get some more optimizations? In the later case we could leave the CMakeLists.txt and Makefile alone, after we've merged either my *cflags* branch or #523 you do some something like `CFLAGS=-march=native make`.

devnexen:
> Or another question: Is `-march=native` necessary to get the opcode, because gcc defaults to some crappy old ARM revision or is ist to get some more optimizations? In the later case we could leave the CMakeLists.txt and Makefile alone, after we've merged either my _cflags_ branch or #523 you do some something like `CFLAGS=-march=native make`.

Absolutely not optimisation but on ARM mode it can generate this asm instruction otherwise. will update the Makefile

Yamagi:
Hi,
I had a discussion about this with some people that know a lot more than me about ARM. The general consent is: 

* For 32-Bit ARM it should be okay to force `-march=armv6k`. It would cover everything starting from the first RPi generation and older cores are too slow for YQ2 anyway.
* For 64-Bit ARM we could force `-march=armv8.0-a`, it would cover all aarch64 compatible cores.

Forcing the architecture to a sane minimum would simplify the code, because you don't need to communicate into the code if the build supports the `yield` opcode.

Yamagi:
I've just merged a squashed version of these commits into master. Thank you for looking into this, it was a good idea and really good work. :) On my RPi 3 the `yield` opcode saves about 7°C with the dedicated server and 3°C with the gl1 renderer.

PR #531, opened 2020-03-18
Document async configuration correction.

ConHuevosGuey:
Document stated setting 'async 1' disabled asynchronization.  This has been corrected.


Yamagi:
Thank you. You were faster than me. :)

PR #532, opened 2020-03-20
fix march armv8

0lvin:
Gcc does not support "armv8.0-a", correct value for gcc should be [armv8.0-a](https://gcc.gnu.org/onlinedocs/gcc-9.1.0/gcc/AArch64-Options.html).

```
Error: unrecognized option -march=armv8.0-a
cc1: error: unknown value ‘armv8.0-a’ for ‘-march’
cc1: note: valid arguments are: armv8-a armv8.1-a armv8.2-a armv8.3-a armv8.4-a armv8.5-a native; did you mean ‘armv8.1-a’?
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -169,7 +169,7 @@ if ("${ARCH}" STREQUAL "arm")
     if (CMAKE_SIZEOF_VOID_P EQUAL 4)
         set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -march=armv6k")
     else()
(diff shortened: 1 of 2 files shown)


0lvin:
I have tested with `CC="clang-9 " make` and `CC="gcc-9" make`.

Yamagi:
Hrm. Apparently it's `armv8.0-a` on clang and `armv8-a` on gcc. At least my clang also accepts `armv8-a`, so let's see if someone complains. If so we'll need to set arch depending on the compiler.

PR #534, opened 2020-03-21
Add a separate noreturn macro for function pointers

smcv:
C11 _Noreturn is only accepted on function declarations, not on function
pointers, so we can't use it on callbacks like game_import_t.error and
refimport_t.Sys_Error. Use a separate macro for those.

The problematic situation doesn't currently happen because the Makefile
hard-codes -std=gnu99, which disables C11 features; but removing
-std=gnu99 (resulting in the compiler's default, currently gnu11) causes
compilation failures with at least gcc 9.x.

---

As discussed briefly on #523, this isn't a practical problem right now, but will become a practical problem if porting to C11 compilers in future.

### DIFF
--- a/src/client/vid/header/ref.h
+++ b/src/client/vid/header/ref.h
@@ -198,7 +198,7 @@ typedef struct
 
 typedef struct
 {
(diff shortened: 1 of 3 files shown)


Yamagi:
Thank you. :)

PR #535, opened 2020-03-21
WIP: arm: Don't use yield if unsupported

smcv:
(See below for updated description of branch)

~~Building with a -march= argument attempts to override the compiler's
configuration. In distributions that target a higher baseline CPU
than we specify, -march=armv6k will force the compiler to try to build
for ARMv6K specifically, rather than that higher baseline.~~

~~For example, in Debian's "armhf" port (like many 32-bit ARM ports), the
compiler targets ARMv7 with the VFPv3D16 floating-point extensions, and
the port explicitly does not support older CPUs. Asking it to build for
armv6k just fails, which appears to be because armv6k implies an older
VFP (floating point) ABI than anything this compiler supports. Even if
it succeeded, it would produce binaries that unnecessarily do not take
full advantage of the distribution's ARMv7 baseline.~~

~~Conversely, in distributions that target a *lower* baseline CPU than
we specify, building with an explicit -march= option will produce an
executable that cannot run successfully on the distribution's baseline
CPU. For example, Debian's "armel" port (ARMv5TE) and the baseline used in
Raspbian (ARMv6, with no K) are designed to be compatible with older ARM
CPUs, even if the CPU they are actually run on in practice might be newer,
such as the ARMv7 in the Raspberry Pi version 2. Individual packages in
these distributions are not generally allowed to override that baseline.~~

~~Instead, let the distribution or user set their target CPU (distributions
configure their gcc to default to a specified baseline, while users can
put -march=native or another suitable -march= option in their CFLAGS if
they want a binary suitable for their specific system), and conditionally
enable the yield instruction at compile-time if that target CPU has
it. As far as I can find out, yield was added in ARMv6K, a minor revision
of ARMv6, so it should be available in all ARMv7 or higher CPUs (which
includes anything 64-bit), plus ARMv6 CPUs *if* they are actually ARMv6K.~~

~~Using -march=armv8-a on 64-bit ARM (aarch64) was unnecessary in any
case, because it is the oldest form of aarch64: the only effect of that
option would be to disable newer functionality if the compiler defaults
to enabling it (which I suspect no distributions do yet, but they might
consider in a few years' time).~~

---

I believe this is correct, but it might be too conservative: adding `|| defined(__ARM_ARCH_6KZ__)` to the condition might also work (I'm unsure whether ARMv6KZ is strictly a superset of ARMv6K).

This has been successfully compiled (although not tested at runtime) on Debian armhf and armel, with some `#warning` to tell me which code path it's taking. It appears to enable the yield instruction on armhf (ARMv7) but not on armel (ARMv5), which is what should happen here.

cc @devnexen, @0lvin

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -168,8 +168,6 @@ endif()
 if ("${ARCH}" STREQUAL "arm")
     if (CMAKE_SIZEOF_VOID_P EQUAL 4)
         set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -march=armv6k")
-    else()
-        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -march=armv8-a")
     endif()
 endif()
 
--- a/Makefile
+++ b/Makefile
@@ -192,10 +192,6 @@ ifeq ($(YQ2_ARCH), arm)
 CFLAGS += -march=armv6k
 endif
 
-ifeq ($(YQ2_ARCH), aarch64)
-CFLAGS += -march=armv8-a
-endif
-
 # ----------
 
 # Systemwide installation.
--- a/src/common/frame.c
+++ b/src/common/frame.c
@@ -146,7 +146,7 @@ Qcommon_Mainloop(void)
 				   a Kaby Lake laptop. */
 #if defined (__GNUC__) && (__i386 || __x86_64__)
 				asm("pause");
-#elif defined(__arm__) || defined(__aarch64__)
+#elif defined(__aarch64__) || (defined(__ARM_ARCH) && __ARM_ARCH >= 7) || defined(__ARM_ARCH_6K__)
 				asm("yield");
 #endif
 


devnexen:
instruction not generated on my raspberry 4 (32 bits).

smcv:
The Raspbian compiler is configured to default to ARMv6 so that the binaries it produces will run on any Raspberry Pi (including version 1, which was ARMv6). I'd recommend using `-march=armv7` or `-march=native` if you have something newer and you don't need your binaries to be portable to older RPi models.

devnexen:
I know :-) so your change is a matter of responsibility transfert in other words.

smcv:
Yes, I suppose it is. Picking a specific CPU to target isn't really yquake2's job: it helps some distributions and users, but hurts others. The compiler vendor (in your case Raspbian) and the user who is compiling binaries (you) are both better-placed to choose a baseline than the maintainer of an arbitrary piece of software that is being compiled.

Yamagi:
And exactly because of that the `-march` setting is overrideable by the environment.

smcv:
So you want builds for 32-bit ARM to default to requiring ARMv6K or up if a user doesn't specify otherwise? That's doable, although I think ARMv7 would probably be a better default. Most 32-bit ARM distros have ARMv7 as their baseline (that's why Debian chose ARMv7 for armhf), with Raspbian being the notable exception.

I don't think the Raspberry Pi v1 is ARMv6K, just ARMv6, so the practical result for Raspberry Pi users (works on version 2 and up, doesn't work on version 1) would be the same with `-march=armv7` baseline as with `-march=armv6k` (although maybe with a bit better code generated).

On Debian armhf and similar distributions, `-march=armv6k` doesn't compile, whereas `-march=armv7` would.

smcv:
How about this? (Work in progress.)

----

* arm: Only use yield instruction if supported
    
    The default on 32-bit ARM is now -march=armv6k, but it's overridable,
    and must be overridden on some toolchains that have a higher baseline
    and cannot compile for -march=armv6k. This includes Debian armhf
    (ARMv7 with the VFPv3D16 floating-point extensions), where armv6k
    appears to imply an older VFP (floating point) ABI than anything the
    compiler supports.
    
    Conditionally enable the yield instruction at compile-time if the target
    CPU has it. As far as I can find out, yield was added in ARMv6K, a minor
    revision of ARMv6, so it should be available in all ARMv7 or higher CPUs
    (which includes anything 64-bit), plus ARMv6 CPUs *if* they are actually
    ARMv6K.

* Don't force target CPU to armv8-a on aarch64
    
    ARMv8-A is the oldest/most-limited/least-extended version of aarch64,
    analogous to the earliest AMD 64-bit processor in the x86 world.
    As a result, forcing this target CPU can only disable features, and
    isn't going to enable anything we might want. The yield opcode was
    added in ARMv7, and all aarch64 CPUs are ARMv8-A or higher (which is
    a superset of ARMv7), so no special architecture specification is
    necessary to access the yield opcode.

----

I've marked this as WIP because of the following issues:

* The first commit message says it's overridable, but having that actually work requires merging the `cflags` branch first.

* As noted above, if you want the default on 32-bit ARM to be representative of CPUs people actually have, I suspect better options are either trust the compiler vendor to have done the right thing (as in the original version of this branch), or go for ARMv7 (which will work on Raspberry Pi v2 or later but not on the Raspberry Pi v1 or Raspberry Pi Zero, the same as Debian armhf and Fedora ARMv7).

Yamagi:
I've just merged the `cflags` branch. Now there's a conflict in the Makefile, because I moved the `-march=` stuff further up and added a longer comment explaining the rationale behind it and what to do if someone really wants to build for ARM5 or older ARM6 CPUs. Can you resolve it? After that i'll merge this PR and hopefully everybody is happy. :)

Regarding the minimum architecture: Everything up to and including the RPi 3B+ is too slow for playing Quake II. That's mostly caused by the crappy GPU driver and it's dead slow legacy OpenGL implementation. I'm getting about 37 FPS at 640x480... For more performance we would need an OpenGL ES 2.0 renderer. I don't think that we'll ever write one, it's too much work for an already  outdated niche platform. The RPi 4 has an OpenGL ES 3.0 compatible GPU, adding an ES 3.0 path to the GL3 renderer is rather easy. We had an experiment implementation somewhere during the GL3 development, it was abandoned due to the lack of hardware. We'll likely revive it some day.  
So going to ARMv7 as the default minimum shouldn't be a big problem. We could just go for `-march=armv7` and see if anyone complains.

smcv:
I'll rebase this now that `cflags` is in.

Yamagi:
I've committed this in 5b327c0.

PR #536, opened 2020-03-21
Fix MINGW64 windows build

0lvin:


src/client/refresh/vk/vk_rmain.c
```diff
@@ -1546,6 +1546,9 @@ R_IsVsyncActive(void)
 
 static int R_PrepareForWindow(void)
 {
+#if defined(__MINGW32__) || defined(__MINGW64__)
+	SDL_Vulkan_LoadLibrary("libvulkan-1.dll");
```
Yamagi: Hmm... I think the function itself should be fine. But what about at least some minimal error checking, for exampling dropping out if the return code is not 0?

DanielGibson: and why not `#ifdef _WIN32` which should also work if/when we get MSVC to work?

0lvin: As for me, it looks like issue in some way related to `mingw`, by [documentation](https://github.com/KhronosGroup/Vulkan-Loader/blob/master/loader/LoaderAndLayerInterface.md#abi-versioning) on windows platform loader should be `vulkan-1.dll`  (without `lib` prefix).

> The Vulkan loader library file name is vulkan-<ABI version>.dll. For example, for Vulkan version 1.X on Windows the library filename is vulkan-1.dll. And this library file can typically be found in the windows\system32 directory (on 64-bit Windows installs, the 32-bit version of the loader with the same name can be found in the windows\sysWOW64 directory).

DanielGibson: If this is supposed to open a **system** lib, why should it matter if YQ2 is built with mingw or MSVC?
Maybe it should always use `vulkan-1.dll` then (and not bundle a libvulkan-1.dll from mingw)?
(I assume that SDL_Vulkan_LoadLibrary() searches the system paths?)

0lvin: I have tried with ```SDL_Vulkan_LoadLibrary("libvulkan-1.dll");``` and i have seen error with surface creation.

Yamagi: Yeah, using the systemwide lib (installed by the driver?) seems the right way to go.

0lvin: If i correctly understand from [SDL_Vulkan_LoadLibrary](https://wiki.libsdl.org/SDL_Vulkan_LoadLibrary), function should be called only in case when we not linked vulkan library to application itself.
> It is fairly common for Vulkan applications to link with libvulkan instead of explicitly loading it at run time.

0lvin: exact error was:
```
setting mode 4: 640 480
Real display mode: 1920x1080@64 (vid_fullscreen: 0)
Enabled extensions: VK_KHR_surface VK_KHR_win32_surface 
...created Vulkan instance
Vkimp_CreateSurface() SDL_Vulkan_CreateSurface failed: VK_KHR_win32_surface extension is not enabled in the Vulkan instance.R_InitContext() - could not initialize Vulkan!
R_SetMode() - invalid mode
```

DanielGibson: if you get that error with `SDL_Vulkan_LoadLibrary("libvulkan-1.dll");`, why do you put it in the code?

0lvin: I had such error only if code does not have such code(`SDL_Vulkan_LoadLibrary`). (Code in PR works and has minimal set for successful run).

0lvin: I have updated PR with error message for SDL_Vulkan_LoadLibrary.

0lvin:
@Yamagi @DanielGibson  have you had a chance to review? Should i add/fix something in relation to run on windows?

Yamagi:
I finally had and I'm mostly fine with it. Can you change `__MINGW32__` and so on to `_WIN32`? after that I'll merge it.

0lvin:
Done, thank you.

Yamagi:
No, thank you! :)

PR #545, opened 2020-04-02
MacOS softrender fixes #541.

### LINKED ISSUE #541: black areas when using software renderer on macOS
Since version 7.42, black areas are visible when using the software renderer on macOS. This happens both in the menu and in the game:

quake2-menu

quake2-game

Version 7.41 does not have this issue. Based on the changelog, it looks like this happened in the changes related to #415 and/or #416. Commit 60f9af27 still works fine, without any visible black areas, just like in version 7.41.

This is still an issue in the current master 07859967. Tested on macOS 10.14.6 (18G4032). My builds were made with clang version 11.0.0 (clang-1100.0.33.17).

Maybe @0lvin can help with this?

0lvin:
On MacOS texture is cleaned up after render and code have to copy a whole
screen to the texture, other platforms save previous texture content and can
be copied only changed parts.

### DIFF
--- a/src/client/refresh/soft/sw_main.c
+++ b/src/client/refresh/soft/sw_main.c
@@ -146,6 +146,7 @@ static cvar_t	*sw_overbrightbits;
 cvar_t	*sw_custom_particles;
 cvar_t	*sw_texture_filtering;
 cvar_t	*sw_retexturing;
+static cvar_t	*sw_partialrefresh;
 
 cvar_t	*r_drawworld;
 static cvar_t	*r_drawentities;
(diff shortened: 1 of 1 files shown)


0lvin:
Code should help with issue on Apple platform with partial render when part of screen rendered as black. 

Other platform looks as does not have such issue. Added debug flag to disable partial copy.

jboekesteijn:
@0lvin Thanks! This fixes the issue on macOS.

Yamagi:
And merged!

Issue #546, opened 2020-04-03
segmentation fault error


BielBdeLuna:
i got the following error:

```
$ ./quake2 

Yamagi Quake II v7.44pre
========================
Client build options:
 + OpenAL audio
Platform: Linux
Architecture: x86_64
Byte ordering: little endian

Added packfile './baseq2/pak0.pak' (1106 files).
Added packfile './baseq2/pak1.pak' (279 files).
Added packfile './baseq2/pak2.pak' (2 files).
Cmd_RemoveCommand: force_centerview not added
Cmd_RemoveCommand: +mlook not added
Cmd_RemoveCommand: -mlook not added
Shutting down input.

=======================================================

Yamagi Quake II crashed! This should not happen...

Make sure that you're using the last version. It can
be found at http://www.yamagi.org/quake2. If you do,
send a bug report to quake2@yamagi.org and include:

 - This output
 - The conditions that triggered the crash
 - How to reproduce the crash (if known)
 - The following files. None of them contains private
   data. They're necessary to analyze the backtrace:

    - quake2 (the executable / binary)

    - game.so (the game.so of the mod you were playing
      when the game crashed. baseq2/game.so for the
      main game)

 - Any other data which you think might be useful

Thank you very much for your help, making Yamagi Quake
II an even better source port. It's much appreciated.

=======================================================

Product:      Yamagi Quake II
Version:      7.44pre
Platform:     Linux
Architecture: x86_64
Compiler:     9.2.1 20191008
Signal:       11

Backtrace:
  ./quake2(printBacktrace+0x35) [0x55cb2dbcccb6]
  ./quake2(signalhandler+0xf1) [0x55cb2dbcced5]
  /lib/x86_64-linux-gnu/libc.so.6(+0x46470) [0x7f1698728470]
  ./quake2(+0x6fee9) [0x55cb2db7cee9]
  ./quake2(+0x70788) [0x55cb2db7d788]
  ./quake2(GLimp_Shutdown+0xd) [0x55cb2db7d9b4]
  ./quake2(VID_Shutdown+0x12) [0x55cb2db83109]
  ./quake2(CL_Shutdown+0x57) [0x55cb2db41ded]
  ./quake2(Com_Error+0x1a5) [0x55cb2db842ea]
  ./quake2(FS_LoadPK3+0xfe) [0x55cb2db8ce67]
  ./quake2(FS_AddDirToSearchPath+0x44c) [0x55cb2db8ea5d]
[...]

PR #549, opened 2020-04-06
Dynamically allocate memory on PAK directory read 

0lvin:
* Fix for #546
* Dynamically allocate memory on PAK directory read 
* Make MAX_FILES_IN_PACK soft limit for archives.

### DIFF
--- a/src/common/filesystem.c
+++ b/src/common/filesystem.c
@@ -701,7 +701,7 @@ FS_LoadPAK(const char *packPath)
 	fsPackFile_t *files; /* List of files in PAK. */
 	fsPack_t *pack; /* PAK file. */
 	dpackheader_t header; /* PAK file header. */
-	dpackfile_t info[MAX_FILES_IN_PACK]; /* PAK info. */
+	dpackfile_t *info = NULL; /* PAK info. */
 
 	handle = Q_fopen(packPath, "rb");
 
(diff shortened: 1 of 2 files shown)


src/common/filesystem.c
```diff
@@ -789,15 +806,23 @@ FS_LoadPK3(const char *packPath)
 	{
 		unzClose(handle);
 		Com_Error(ERR_FATAL, "FS_LoadPK3: '%s' is not a pack file", packPath);
+		return NULL;
 	}
 
 	numFiles = global.number_entry;
 
-	if ((numFiles > MAX_FILES_IN_PACK) || (numFiles == 0))
+	if (numFiles > MAX_FILES_IN_PACK)
```
DanielGibson, line 814: I think in case of zip this check is superfluous - I guess having it was a bug that came from copy-pasting the .pak code and forgetting to remove if

src/common/filesystem.c
```diff
@@ -715,19 +715,35 @@ FS_LoadPAK(const char *packPath)
 	if (LittleLong(header.ident) != IDPAKHEADER)
 	{
 		fclose(handle);
-		Com_Error(ERR_FATAL, "FS_LoadPAK: '%s' is not a pack file", packPath);
+		Com_Error(ERR_FATAL, "%s: '%s' is not a pack file", __func__, packPath);
+		return NULL;
```
DanielGibson, line 719: why return NULL; ? Com_Error() exits the game

src/common/filesystem.c
0lvin: I have removed warning about MAX_FILES_IN_PAK for pkz.

src/common/filesystem.c
0lvin: It was for cppcheck. I have removed it as not required.

Yamagi:
Thank you. Keep up the good work! :)

Issue #543, opened 2020-04-01
Belgian AZERTY layout has unbindable keys


InFerYes:
When launched in a terminal, it outputs `"<UNKNOWN KEYNUM>" isn't a valid key` when I try to bind "previous item".

In this image it is the key in red:

![https://i.imgur.com/1jbiPt5.png](https://i.imgur.com/1jbiPt5.png)

Yamagi:
This is a known limitation of Q2s input system. Most modern games use scancodes, e.g. you press the key or key combination for `ä` and the game receives the string 'ä'. With scancodes the game doesn't need any understanding of the keyboard and it's layout. But Q2 - like many old games - uses keycodes. You press `ä`, the game receives the keycode '48' for a german QWERTZ keyboard in Linux with X11. So the game needs at least a little understanding of the keyboard and it's layout, YQ2 uses SDL2 for that. Since dead keys don't generate keycodes and key combination ofzen generate codes unknown to SDL2 keys like the one you mentioned can't be bind.

The correct solution would be to switch YQ2 to scancodes, but that's a lot of work. Therefore we've never done. It might be possible to hack some fallback logic, check the scancode if an unknown keycode is received. We'll look into that but I'm afraid that it won't happen soon. :/

InFerYes:
Understandable. Can you point me in the right direction?

Yamagi:
The code is in src/client/input/sdl.c. Every keystroke ends in `IN_Update()` as `SDL_KEYUP` event. There's already a little hack using scancodes for some problems with AZERTY keyboards that I forgot about. It may be enough to enhance that. If the hack doesn't trigger the keysym is passed to `IN_TranslateSDLtoQ2Key()` which translates it to Quake IIs internal key representation. You could:

1. Alter `IN_TranslateSDLtoQ2Key()` in way that something like -1 is returned if an unknown keysym is encountered.
2. Fallback to scancodes if `IN_TranslateSDLtoQ2Key()` returns -1.

InFerYes:
`Key_KeynumToString` in src/client/cl_keyboard.c keeps returning `"<UNKNOWN KEYNUM>"` in STDOUT.

I have tried adding to `IN_TranslateSDLtoQ2Key`:

```C
case SDLK_CARET:
key = K_CARET;
```

based on https://wiki.libsdl.org/SDL_Scancode, which has the ^ character as `SDLK_CARET`, under the section 

> These virtual key values do not have corresponding physical keys

[...]

PR #551, opened 2020-04-08
Scancodes for Input and other keyboard input improvements

DanielGibson:
Fixed several things that have to do with `K_*` constants and add `K_SC_*` for scancodes that are used as fallbacks for otherwise unknown keys.
See commit messages for more details.


src/server/sv_main.c
```diff
@@ -339,6 +339,12 @@ SV_RunGameFrame(void)
 	{
 		time_before_game = Sys_Milliseconds();
 	}
+	/*{
+		static int gametime = 0;
+		int newtime = Sys_Milliseconds();
+		printf("since last gameframe: %d ms\n", newtime - gametime);
+		gametime = newtime;
+	}*/
```
Yamagi, line 347: Shouldn't this be removed? Or commented in and hidden behind something like `developer`?

DanielGibson: oh, yes, I didn't mean to commit that

DanielGibson:
Added another commit, see commit message.

I'm sorry.

DanielGibson:
Now also includes a hack to support the "console key" on almost all keyboard layouts and fixes that allow binding keys with keycodes that are "special" in configs: `;`, `$` and `"` (and, just to be sure, `'`)

DanielGibson:
According to https://github.com/yquake2/yquake2/issues/543#issuecomment-612662094 this now works as intended (or even better), @Yamagi feel free to merge unless you have additional objections :)

Yamagi:
This looks fine and should work as intended. If there're some unwanted site effects we can fix them in master.

Issue #558, opened 2020-04-11
Can't hear firing own weapons in dday mod


InFerYes:
In DDay when I fire **any** weapon in the game, my own gunshot is not heard. I can hear other people fire the same guns.

I have checked with `developer 1` in console but see no missing files or errors.

I have tried compiling [q2pro](https://github.com/skullernet/q2pro) to check if the problem is local (missing files) but after symlinking the same folders (baseq2 and dday) it properly plays the gunshots in q2pro, so I'm suspecting it's a yamagi bug. 

An example: (https://github.com/InFerYes/dday/blob/master/gbr/gbr_guninfo.c)
```C
GunInfo_t gbrguninfo[MAX_TEAM_GUNS]=
{
....
	{//Lee Enfield
		{4},{86}, 16,101, 81,81,

			"gbr/303/unload.wav",
				{58},
		
			"gbr/303/reload.wav",
				{67},
		
		"gbr/303/fire.wav",
		"gbr/303/lastround.wav",

		MOD_RIFLE,
		DAMAGE_Enfield, 0,
		13//17
	}
...
};
```
![https://i.imgur.com/uzMr2br.png](https://i.imgur.com/uzMr2br.png)

unload.wav and reload.wav are played properly, fire.wav and lastround.wav are not.

Yamagi:
This could be another fallout from 16ee007. @bibendovsky, @BjossiAlfreds: What do you think?

BjossiAlfreds:
A simple way to check is to reverse that change and then test the mod. Though that change only affects sound events with same scheduled time (reversing the priority so that newest scheduled sound event plays over old ones), so unless I'm missing something there would have to be another sound event that is overriding the firing sounds.

InFerYes:
I reverted this locally and now the firing sounds are played correctly.

BjossiAlfreds:
The old sorting was that the first scheduled sound overrides others that come after it with the same play time. What I wonder now is what sound event is being scheduled after the firing sound to cause this to happen? It should be relatively easy to trace the source code for that, at least assuming it is on the game API side - Quake 2's muzzleflashes and effect stuff does a lot of hard-coded shenanigans outside the game API.

InFerYes:
I don't know enough about the game to find how the sounds are being played and trace everything. I just fix trivial things as I come across them.

BjossiAlfreds:
[...]

PR #564, opened 2020-04-22
Don't play silenced muzzle flash sounds

bibendovsky:
Experimental approach to fix play sound queue.

The idea is to skip playing muzzle flash sound if it's audio data is empty or contains only silence (or very close to it).

See #558 for details.


Yamagi:
This is a really clever idea. I'm afraid that it'll only solve the problem described in #558 and not all the other endless places were it can go wrong. What about giving the user 3 options, selectable by a cvar:

1. New ordering for baseq2, xatrix and rogue, old ordering for everything else.
2. New ordering for everything.
3. Your approach.

I'm pushing my patch to option 1. (non-selectable at this time) in a few minutes. Your branch will likely conflict after that.

bibendovsky:
Added cvar `s_ps_sorting` to control sorting of play sounds.

Values:
`1` (default) New sorting for mission packs only (including Quake 2 itself) or the old one otherwise.
`2` New sorting.
`3` New sorting and skip muzzle flash sounds if they are silence.

Yamagi:
Thank you :)

Issue #508, opened 2020-01-24
liquid shaders in GL mode


StepanST:
I dont think this is bug, but in OpenGL mode liquid shaders has borders between sectors on one textured place. In software mode there is no borders. In different versions of Quake 1 port (dont remember witch one, QuakeSpasm or DirectQ) was different shaders of liquids, one of them was with borders and another without. Is it possible to make liquid shaders in GL mode like in software or it is more deep thihg?

DanielGibson:
Can you add a screenshot that demonstrates this? Not 100% what you mean

StepanST:
![Без имени-1](https://user-images.githubusercontent.com/43229017/73506884-9d240780-4409-11ea-8231-5bcb86094193.jpg)
This is first map of Groun Zero addon, but in original Q2 it although appears

Yamagi:
This should be fixed by merging #569.

PR #569, opened 2020-05-05
Move liquid transformations into fragment shader

bibendovsky:
Fixes visible primitive edges on liquid surfaces.
See #508 for details.


Yamagi:
This even gives a little performance gain, 215 vs. 220 FPS on an old (and slow) Ivy Bridge IGP. Thank you!

PR #574, opened 2020-05-09
fix tearing with fix palette change(explosion near without move)

0lvin:


Yamagi:
And merged. Thank you! :)

PR #575, opened 2020-05-10
Fixed the Jorg skin bug present since the original Quake 2 release.

kondrak:
I fixed this in my own project and figured it might be worth to move it to Yamagi Quake 2, since it seems to still have that bug present.
Jorg and Makron models need to be swapped in order for the "hurt" skin to trigger correctly - right now, if Jorg has < 50% of health, it's the Makron's skin that's being replaced instead of the Jorg's. This fix has been tested in vkQuake2 but I see that multi-model entities (I think Jorg is the only one) are handled in the same manner here.


Yamagi:
Thank you. :) The multi-model code is (more or less) unchanged from Vanilla Q2 and I can confirm that it works.

PR #585, opened 2020-08-03
backtrace support rework proposal.

devnexen:
Linux/Solaris supports it in the libc however, BSD in general, Haiku
relies on an external library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,11 +103,21 @@ find_package(OpenGL REQUIRED)
 list(APPEND yquake2IncludeDirectories ${OPENGL_INCLUDE_DIR})
 list(APPEND yquake2OpenGLLinkerFlags ${OPENGL_LIBRARIES})
 
-# FreeBSD needs libexecinfo.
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -103,11 +103,21 @@ find_package(OpenGL REQUIRED)
...
+check_function_exists(backtrace HAVE_EXECINFO_SYS)
+IF (NOT HAVE_EXECINFO_SYS)
+	check_library_exists(execinfo backtrace "" HAVE_EXECINFO_LIB)
+    if (HAVE_EXECINFO_LIB)
+        list(APPEND yquake2ClientLinkerFlags execinfo)
+        list(APPEND yquake2ServerLinkerFlags execinfo)
+        add_definitions(-DHAVE_EXECINFO)
+    endif()
+else()
+    add_definitions(-DHAVE_EXECINFO)
```
Yamagi, line 120: Do we need the same change for the Makefile?

devnexen: sorry :) I forgot

Yamagi: Wow, that was fast :)

Yamagi:
Aaand merged!

PR #589, opened 2020-08-19
added listmaps command

JoBergeron:
Added command 'listmaps' to list available maps to load through the console. It will list maps in pak files first followed by maps in the 'maps/' folder from baseq2 or mod or expansion being played.

q2_0000

### DIFF
--- a/doc/050_commands.md
+++ b/doc/050_commands.md
@@ -17,3 +17,7 @@ original clients (Vanilla Quake II) commands are still in place.
   entities.
 
 * **teleport <x y z>**: Teleports the player to the given coordinates.
+
+* **listmaps**: Lists available maps for the player to load. Maps from
+  loaded pak files will be listed first followed by maps placed in 
+  the current game's maps folder.
--- a/src/server/sv_cmd.c
+++ b/src/server/sv_cmd.c
@@ -310,6 +310,40 @@ SV_Map_f(void)
 }
 
 /*
+ * Lists available maps for user to load.
+ */
+SV_ListMaps_f(void)
+{
+	char **userMapNames;
+	int nUserMaps = 0;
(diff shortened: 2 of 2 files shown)


Yamagi:
Hi,
that's a very nice idea. The code looks good an I'm going to merge it. But can you add the command to the documentation in https://github.com/yquake2/yquake2/blob/master/doc/050_commands.md ?

Yamagi:
And what about sorting the list?

JoBergeron:
I'm not sure how we should sort this. At the moment, the the maps that 
come from paks will listed in their order of appearence in their pak. 
Files that are stored in the maps/ folder will be appear in a 
alphabetically sort manner.

In the screenshot provided with my pull request. Maps listed from 
'mine1' to 'boss2' are maps from the baseq2 pak files. 'angron1a' and 
'angron1b' are custom maps which are present in the 'maps/' subdirectory 
in baseq2.

I feel this is fine because users of this command will probably be 
interested in listing the map names from their maps/ folder to know 
which custom map to load. If there's a better idea on the overall 
sorting though, I'd be glad to give it a go on the implementation side.

JoBergeron:
> 
> 
> Hi,
> that's a very nice idea. The code looks good an I'm going to merge it. But can you add the command to the documentation in https://github.com/yquake2/yquake2/blob/master/doc/050_commands.md ?

Yes I will add it there!

Yamagi:
> I feel this is fine because users of this command will probably be interested in listing the map names from their maps/ folder to know which custom map to load.

You're right. Sorting the entries would require a fair amount of copying the strings, etc. Let's leave things as they are.

Yamagi:
And merged. Thank you! :)

PR #595, opened 2020-09-01
Haiku support, code simplification.

devnexen:
using more straightforward approach to get the current binary path.
Actual approach more fit to get other infos (e.g. address mappings and so on).


Yamagi:
Thank you!

PR #598, opened 2020-09-06
map name autocompletion

JoBergeron:
When using the command to load a map "map x", if one character from the map name is provided, it will try to autocomplete it.

There are 3 possible scenarios I anticipated:

1) The name is an exact match. In this case we do nothing.
2) The name is partially filled out and only 1 match was found. In this case we fill out the rest.
3) The name has many matches. In this case we complete the map name with the characters from the matches until one differs. We also provide the list of the current matches.


Yamagi:
That's a nice idea. Merged!

PR #599, opened 2020-09-07
Create "mods" submenu to allow changing "game" cvar via UI

earth-metal:
I tried opening the original PR back up, but since I also squashed my commits it was refused. Using Address Sanitizer was easier than I expected.  In addition to correcting my memory leaks, I've also added verification that any params the mod-finder passes to "FS_ListFiles" are formatted properly and any NULL data it returns gets ignored, so mods in "datadir" are located correctly.
I won't be so quick to close a PR next time.


Yamagi:
Thank you for this second version. Looks like everything is working now :)

PR #603, opened 2020-09-30
Little simplification for NetBSD, /proc/<pid>/exe works just fine too.

devnexen:


krytarowski:
Please use sysctl with KERN_PROC_PATHNAME,

DanielGibson:
Can NetBSD use the same code as FreeBSD?

krytarowski:
Almost yes, with different `name` values.

DanielGibson:
Can you give an example for the name values?
Is this a new addition or has this always been supported? IIRC when we first wrote this (2015) the usual way for NetBSD was what we're doing there.

krytarowski:
https://github.com/qemu/qemu/blob/master/util/oslib-posix.c#L383

`KERN_PROC_PATHNAME` is supported in all worth supporting releases.

devnexen:
@krytarowski does it mean NetBSD plans to retire /proc at some point ?

krytarowski:
It is kept for some compat with linux binaries, but not for native applications.

Yamagi:
And merged. Thank you.

PR #604, opened 2020-10-02
Last bit for proper Haiku support.

devnexen:
Even tough it is a multi user os, due to BeOs heritage and being backwards
compatible with, the desktop runs as root, thus its dog get eaten after all...


Yamagi:
And merged, thank you.

PR #627, opened 2020-11-24
Cast arguments to ctype functions to unsigned char

niacat:
According to the C standard, arguments to the ctype functions must fit into unsigned char (presumably so they can be implemented with simple array access). This causes a build time warning on NetBSD, and may function incorrectly if any UTF-8 strings are used.

### DIFF
--- a/src/client/curl/download.c
+++ b/src/client/curl/download.c
@@ -127,7 +127,7 @@ static void CL_EscapeHTTPPath(const char *filePath, char *escaped)
 
 	for (int i = 0; i < len; i++)
 	{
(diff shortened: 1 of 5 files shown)


DanielGibson:
those functions don't work correctly with UTF-8 anyway (for that they'd have to be able to handle multiple chars at once) - but I think they're only used with ASCII anyway

niacat:
The behaviour for codepoints outside of ASCII range (i.e. utf-8 values packed into unsigned char) is to ignore them. However, the behaviour for codepoints below 0 (i.e. utf-8 values packed into signed char) is undefined nasal demons.

PR #631, opened 2020-12-08
Fix P_ProjectSource forward declaration

Spirrwell:
The definition for P_ProjectSource looks like it changed here: https://github.com/yquake2/yquake2/commit/85fb607010ee151b2d6bc217647b41fa687aa0dc#diff-d8777409dd928ef41e564600cc9bdd7ccf260032f18b570a45f750e208936c03R44

This means that the forward declaration is wrong and this pull request fixes that.

I'm not sure if this bad forward declaration caused any issues, but this breaks compilation on my C++ fork of this project and this seemed better to fix here.

### DIFF
--- a/src/game/savegame/tables/gamefunc_decs.h
+++ b/src/game/savegame/tables/gamefunc_decs.h
@@ -133,7 +133,7 @@ extern void NoAmmoWeaponChange ( edict_t * ent ) ;
 extern void ChangeWeapon ( edict_t * ent ) ;
 extern qboolean Pickup_Weapon ( edict_t * ent , edict_t * other ) ;
 extern void PlayerNoise ( edict_t * who , vec3_t where , int type ) ;
-extern void P_ProjectSource ( gclient_t * client , vec3_t point , vec3_t distance , vec3_t forward , vec3_t right , vec3_t result ) ;
+extern void P_ProjectSource ( edict_t * ent , vec3_t distance , vec3_t forward , vec3_t right , vec3_t result ) ;
 extern void ClientEndServerFrame ( edict_t * ent ) ;
 extern void G_SetClientFrame ( edict_t * ent ) ;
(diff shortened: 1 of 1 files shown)


mjr4077au:
I'm very sorry that I missed this prototype during that commit. I guess I had a working build and didn't think much of it.

Thanks for fixing it up :)

mjr4077au:
As a side note, nice to see support for a modern toolchain for yquake2. Definitely prefer MSVC on Windows over mingw.

Spirrwell:
No worries! For some reason C doesn't seem to care about it. It seems like there's a giant list of function pointers which includes this function in gamefunc_list.h. Not sure if it's important or not lol.

But yeah, the original reason I did a C++ port was to use MSVC, because it was lacking proper C support. MSVC has C support now, but it doesn't seem perfect.

mjr4077au:
The prototype can probably just go since `P_ProjectSource()` is only used statically in weapon.c. This fix might also be needed in the other games hosted as well like rogue and xatrix.

I'll check your port out on the weekend. Your understanding about MSVC and C support is my understanding as well. Depending on how much refactoring you do, there could be a genuine reason to maintain it.

Spirrwell:
If people would rather remove the prototype instead, I can do that. So if somebody else wants to confirm which they'd prefer, I can leave it as is or do that.

----
Regarding my C++ port, most of it was just adding casts. I believe I did very little refactoring so I can update from upstream. Though it might benefit from more work to fix warnings and such.

Yamagi:
The function pointers are used by the savegame system. When saving the game, all function pointers in all structs holding the games state are placed with strings. When loading the game, the strings are turned into functions pointers again. This way the game can be altered (recompiled with another compiler version, bug fixes and even new features integrated,  etc.) without breaking savegame compatibility. While `P_ProjectSource()` isn't referenced through function pointers and therefor not used by the savegame system, correcting the prototype is correct way to go.

Issue #633, opened 2020-12-10
joyaltselector stays down if game is quickloaded with alt joystick button


lazd:
## Steps to reproduce

1. Set up a button on the joystick for `enable alt joy keys` (in my case, the Xbox button)
```
bind JOY17 "+joyaltselector"
```
2. Bind some alt buttons for quicksave/quickload (in my case, the L1 and R1 bumpers)
```
bind JOY7_ALT "echo Quick Loading...; wait; load quick"
bind JOY8_ALT "echo Quick Saving...; wait; save quick"
```
3. Quickload by pressing `JOY17` + `JOY7` on your controller
4. Press `JOY7` without pressing `JOY17`
5. If nothing happens, go back to step 3 and try pressing and releasing `JOY17` or `JOY7` at different times, it doesn't happen every time

## Actual

Sometimes, quickload is triggered without pressing `JOY17` -- you can keep pressing `JOY7` over and over and it will keep quickloading your game without ever having to press `JOY17`

## Expected

Quickload is not triggered without pressing `JOY17`

## Notes

I could not find the codepath to call `IN_JoyAltSelectorUp` after loading a game, but that would be the solution.

Yamagi:
Hi,
that's tricky. The problem here is, that all the savegame stuff and some other things like map changes or server initiated reconnects are in the server and game. The input stuff is in the client. The server can't (or shouldn't) call into the client, because it would break network transparency or the fix would only work when playing on a local server.

What about something like this, untested:
* Add `IN_JoyAltSelectorUp()` to `Key_MarkAllUp()`. That will make sure that not only keyboard keys are marked as up at several places throughout the client, but also the joystick selector.
* Call `In_FlushQueue()` from `CL_ParseServerData()` to flush the SDL event queues (so pending key events don't get processed) and mark all keys as up.

PR #634, opened 2020-12-11
Flush keys when server data is parsed

lazd:
This makes sure key states are reset when the game is reloaded, fixing #633 

Additionally, a debug statement in `sdl.c` has been changed to only output in debug mode as it was making testing hard (couldn't see the messages because it kept overwriting output from quicksave) :)

### DIFF
--- a/src/client/cl_parse.c
+++ b/src/client/cl_parse.c
@@ -829,6 +829,9 @@ CL_ParseServerData(void)
 	char *str;
 	int i;
 
+	/* Clear all key states */
+	In_FlushQueue();
+
 	Com_DPrintf("Serverdata packet received.\n");
 
 	/* wipe the client_state_t struct */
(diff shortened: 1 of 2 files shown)


lazd:
@Yamagi thanks for the insight! Adding `IN_JoyAltSelectorUp()` to `In_FlushQueue()` and calling `In_FlushQueue()` from `CL_ParseServerData()` and did the trick.

I did not opt to call `IN_JoyAltSelectorUpI()` inside of `Key_MarkAllUp()` as `cl_keyboard.c` is not currently including `input.h`, and it seemed out of place to deal with joystick matters in `cl_keyboard.c`. Let me know if you feel otherwise and I will make the change.

Yamagi:
> I did not opt to call IN_JoyAltSelectorUpI() inside of Key_MarkAllUp() as cl_keyboard.c is not currently including input.h, and it seemed out of place to deal with joystick matters in cl_keyboard.c. Let me know if you feel otherwise and I will make the change.

Nope, I'm okay with this.

Issue #588, opened 2020-08-18
Feature Request:  Resolution scaling.


ConHuevosGuey:
Hi. Would it be possible to implement a resolution scaling function similar to Quakespasm's OpenGL implementation for the original Quake?  It a nice touch of nostalgia to try and capture that original look of playing it on an old PC at a low resolution.

A snippet of the feature is available here from Spike's fork of Quakespasm:

https://github.com/Shpoike/Quakespasm/blob/master/quakespasm/Quake/gl_rmain.c#L1029

Attached is a screenshot of Quake with and without the feature being used to display what it does.  Thank you.

![normal](https://user-images.githubusercontent.com/445334/90545580-78497480-e14e-11ea-808e-e32299f70287.png)
![scaled](https://user-images.githubusercontent.com/445334/90545590-7aabce80-e14e-11ea-99b5-e72dcac725ac.png)

Yamagi:
Hi,
that's a nice idea. At a very first thought resolution scaling needs the same mechanism as the still missing underwater shaders (see #514). For both effect the the scene needs to be rendered into a buffer or something like that, the effects are applied to the buffer. We'll look into it, but it may take some time.

ConHuevosGuey:
Okay.  Thank you for the response.  I would really love to see this added.  

  As a side note, is there an IDE you could recommend to modify the Yamagi source that's easy to use with a CMake project such as this?  Sorry to ask in this issue but I couldn't find any other contact methods.  I'm no professional programmer, just a hobbyist and familiar with Visual Studio.  I find it challenging to poke around with the source using only a text editor, so I'm curious if there's a more amateur friendly way.

One last question.  I downloaded and compiled the latest source from August and noticed a massive performance hit.  I cannot reach 120 FPS in the OpenGL 1.4 renderer.  Is there some debug flag set that isn't present in the release candidate?  

Thank you once again.

Yamagi:
> As a side note, is there an IDE you could recommend to modify the Yamagi source that's easy to use with a CMake project such as this? Sorry to ask in this issue but I couldn't find any other contact methods. I'm no professional programmer, just a hobbyist and familiar with Visual Studio. I find it challenging to poke around with the source using only a text editor, so I'm curious if there's a more amateur friendly way.

[...]

PR #636, opened 2020-12-21
Add resolution scaling to Vulkan renderer

rg3:
This commit adds a new cvar called vk_pixel_size that represents how big
pixels should look in the rendered world in order to simulate lower
screen resolutions. With its default value of 1 everything looks normal,
but with bigger sizes (e.g. 4) the rendered world starts to look
"pixelated" due to pixels appearing bigger.

To implement the effect, the viewport and scissor are modified when
drawing the world so the rendering results cover a smaller area in the
top-left corner of the image.

The post-processing fragment shader is used to scale the image back to
the swapchain size before drawing UI elements on top of it.

The UI is not affected by this change, so the existing UI scaling
options continue to work as before with no changes, adding some
flexibility to the mix.

Related to feature request #588.


AnotherLife:
Interesting!! So what do values like 2 or 3 do?

rg3:
> Interesting!! So what do values like 2 or 3 do?

It's similar, just less noticeable depending on the resolution, screen size, etc. vk_pixel_size determines the length of the square side in pixels. Value 1 means pixels are 1x1 squares (as usual), value 2 means they are 2x2, 3 means 3x3, 4 means 4x4, etc.

The following screenshots were taken with sizes 1, 2 and 4.

q2_0000
q2_0001
q2_0002

DanielGibson:
> The world is still rendered in the chosen screen resolution and the game performs accordingly.

Wouldn't it make more sense to render at the appropriate lower resolution (e.g. 0.5*w, 0.5*h for `vk_pixel_size 2`) and scale that up to the real resolution in the postprocess step?
Should look the same, but have the advantage of additionally making rendering cheaper (maybe especially interesting once Vulkan support on Raspberry Pi is available in Raspbian)

(Might be more work though)

rg3:
Indeed, it probably made more sense but I thought it would be much more complicated. However, I gave it a shot today and it turns out we can simply scale back the viewport and scissor when rendering the world and it works. I've updated the branch and the top comment.

0lvin:
Could you please add the new flag to a [documentation](https://github.com/yquake2/yquake2/blob/master/doc/040_cvarlist.md#graphics-vulkan-only)?

rg3:
Yes, added and updated branch.

Yamagi:
@DanielGibson, are you okay with merging this?

DanielGibson:
Oh sorry, totally forgot about this!
Yes, I think this looks good! :)

Yamagi:
In that case: Merged! Thank you, @rg3!

PR #638, opened 2020-12-23
Proposing to make vulkan support eventually optional.

devnexen:
Still some alternative oses does not have vulkan support or not full (e.g. Raspberry).


0lvin:
Could you please add a new flag to a [documentation](https://github.com/yquake2/yquake2/blob/master/doc/040_cvarlist.md#graphics-vulkan-only)?

Yamagi:
And merged. Thank you! :)

PR #642, opened 2021-01-09
Makefile: Fix architecture detection on Windows

DanielGibson:
$PROCESSOR_ARCHITECTURE seems to contain the architecture of the host,
but we need the architecture the current MinGW shell is targeting.
$MINGW_CHOST seems to be just that, and on my system it's either
i686-w64-mingw32 (mingw32.exe) or x86_64-w64-mingw32 (mingw64.exe)
(No idea what it looks like for Windows on ARM...)

Now this is correctly set and also SSE is used for floating point math which fixes weird lightmap bugs we've been seeing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,6 +84,13 @@ endif
 
 # Detect the architecture
 ifeq ($(YQ2_OSTYPE), Windows)
+ifdef MINGW_CHOST
+ifeq ($(MINGW_CHOST), x86_64-w64-mingw32)
+YQ2_ARCH ?= x86_64
+else # i686-w64-mingw32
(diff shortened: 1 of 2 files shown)


DanielGibson:
alternatively we could probably enforce SSE math on x86 (haven't tested that though)

DanielGibson:
I'll change this to use SSE, also, as Yamagi pointed out, I'll have to look into savegame compatibility, as `$YQ2_ARCH` is used there and is now different (correct).

DanielGibson:
done

Yamagi:
And merged. Thank you!

PR #643, opened 2021-01-10
Enable depthWriteEnable for particles

0lvin:


Yamagi:
Windows binaries are here: https://deponie.yamagi.org/quake2/windows/testbuilds/quake2-0lvin-20210110-3b9aa64.zip

Yamagi:
As far as I can tell this fixes the problem. So I'm merging it. Good work, keep it up!

PR #644, opened 2021-01-12
Skybox fixes

### LINKED ISSUE #639: Skybox corners visible with Vulkan and low screen resolutions
When screen resolutions are low (see the attached 320x240 screenshot), the skybox corners appear to be visible. I think the problem also happens with OpenGL renderers, but it's not as noticeable there.

q2_0003

rg3:
A couple of skybox-related fixes for Vulkan.

* Fix skybox edges being visible due to texture coordinates by using CLAMP_TO_EDGE addressing in samplers when drawing it
* Fix black pixels along skybox edges due to skyboxes not being watertight.

Fixes #639.

### DIFF
--- a/src/client/refresh/vk/header/qvk.h
+++ b/src/client/refresh/vk/header/qvk.h
@@ -86,6 +86,7 @@ typedef struct
 	VkFormat  format;
 	VkDescriptorSet descriptorSet;
 	uint32_t mipLevels;
+	qboolean clampToEdge;
 } qvktexture_t;
 
 #define QVVKTEXTURE_INIT     { \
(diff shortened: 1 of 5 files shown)


Yamagi:
Thank you for looking into this and coding the fix. I wonder why we never saw this in GL3. As far as I remeber we had a lot of problems with (unfixable) inconsistencies in the precalculated visibility sets of world geometry, but never with the skyboxes. Anyways: Merged!

rg3:
Glad I could help! I barely know OpenGL, but when I looked at how a frame is drawn in OpenGL from Renderdoc, it looked like it was using the same trick I pulled off in Vulkan: the contents of the previous frame are still in the color buffer while the new frame is being drawn, so it could mask the error.

PR #653, opened 2021-01-18
Rename `cl_anglekicks` to `cl_kickangles` for consistency with Q2PRO

Calinou:
Q2PRO already has a `cl_kickangles` cvar (which behaves the same as Yamagi Quake 2's and is also cheat-protected). I think we should rename Yamagi Quake 2's cvar so that Q2PRO users can feel more at home :slightly_smiling_face:

### cvar in Q2PRO

Q2PRO console


Yamagi:
Good Idea. I didn't know that q2pro has the same feature. Thank you!

PR #654, opened 2021-01-18
Sync variable names for msaa, retexturing and anisotropic filtering

0lvin:
Rename msaa, retexturing and anisotropic filtering variables to use `r_` prefix.


Yamagi:
Thank you for doing this. :)

PR #655, opened 2021-01-21
Support of wav stereo sound

0lvin:
* validate count of audio channels (only mono/stereo audio are supported)
* make internal functions static.

### DIFF
--- a/src/client/refresh/soft/sw_main.c
+++ b/src/client/refresh/soft/sw_main.c
@@ -1509,7 +1509,7 @@ RE_SetMode(void)
 			vid_fullscreen->modified = false;
 			R_Printf(PRINT_ALL, "%s() - fullscreen unavailable in this mode\n", __func__);
 
-			if ((err = SWimp_SetMode(&vid.width, &vid.height, r_mode->value, 0)) == rserr_ok)
+			if (SWimp_SetMode(&vid.width, &vid.height, r_mode->value, 0) == rserr_ok)
 			{
 				return true;
 			}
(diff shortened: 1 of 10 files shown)


Yamagi:
So the only real change is to allow the upload of samples with 2 channels into OpenAL and SDL. That apparently works for samples that are played on the players position, because they aren't spartialized. But what happens if stereo samples are spartialized? I think that nothing would happen with OpenAL, because the library is usually safe against such things. But what about SDL? Not that I'm against the change, I just try to understand what possible could go wrong.

0lvin:
> So the only real change is to allow the upload of samples with 2 channels into OpenAL and SDL. 

Yes, changes contain only remove mono/stereo check. All other changes are related only to the code style or remove unused code.

About SDL/OpenAl behaiviour:
* SDL audio works without any issues or changes from our side for mod. Maybe it has some issues that I have not caught.
* OpenAL has differences in sound behavior, when defender soldier took attention - sound is too loud and long if compare to original sound. 

I have resolved OpenAl by replace to stereo files based on origin:
```
ffmpeg -i SOLSGHT1.wav -af "pan=stereo|c0=c0|c1=c0" sound/soldier/solsght1.wav
ffmpeg -i SOLSGHT1.wav -af "pan=stereo|c0=c0|c1=c0" sound/soldier/solsrch1.wav
SOLSGHT1.wav: RIFF (little-endian) data, WAVE audio, Microsoft PCM, 16 bit, mono 22050 Hz
sound/soldier/solsght1.wav: RIFF (little-endian) data, WAVE audio, Microsoft PCM, 16 bit, stereo 22050 Hz
```
As i expect that it related to `sound volume <-> distance` function to source.

If we speak about regressions with original mono sound - code works without any issues as I see. And we can advertise this functionality as initial implementation that required changes in mods or at least checking ;-)

Yamagi:
Okay. I'm merging this now. Thank you for looking into it. I'll add some kind of known problems sections to the documentation soon and will mention the stereo samples in it. They're supported to some extend, but users or mod authors shouldn't complain it they run into problems.

Issue #646, opened 2021-01-14
Make the MSAA video menu option work with Vulkan


rg3:
Currently, the MSAA video menu option changes `gl_msaa_samples`. It would be nice if it would change `vk_msaa` when the Vulkan renderer is selected. I'll try to provide a pull request to fix this.

0lvin:
I think use `gl_msaa_samples` in vulkan render will be better instead create hacks for render type. 

Or we can rename `gl_msaa_samples` to something like `r_msaa_samples` with backward compatibility in [the replacement list](https://github.com/yquake2/yquake2/blob/QUAKE2_7_45/src/common/cvar.c#L39). Also `*_retexturing` and `*_nolerp` looks as good candidate for such replacing.

@Yamagi @DanielGibson What do you think?

rg3:
Reusing `gl_msaa_samples` for Vulkan sounds a bit weird to me.

I think there's some value in keeping two separate variables merely because the multisample modes available in Vulkan and OpenGL may differ (e.g. on NVIDIA cards you may be able to select 16x in OpenGL but only 8x in Vulkan), so having both is a bit flexible. The idea of having a unified variable sounds very sensible as well. If we go the unification route, I prefer to keep the "Vulkan way" of simply specifying the exponent as the variable value.

In any case, I can do whatever you think it's better, just let me know.

Yamagi:
If we want one cvar for all renderers it has to be `r_msaa_samples`. And of course all renderes need to implement it the same way, otherwise unexpected thing may happen if a player switches from one renderer to the other. Since 8.0 will be a major release we can introduce smaller config breakages like altering the scale of the MSAA Samples cvar in the GL renderers if it's unavoidable.

0lvin:
I have added logic in #654 for use same msaa values as in gl renders. Values are msaa samples count as it was with gl renders and old users should not see any visual settings differences.

rg3:
Thanks a lot for implementing this! I've just tested it and it seems to work perfectly. :ok_hand: 

I've detected the console output indicating the number of samples needs an adjustment. I'll send a small PR with the fix.

rg3:
I think this is already addressed, so closing...

PR #656, opened 2021-01-23
Fix text indicating MSAA samples for Vulkan

rg3:
Minor fix so the number of samples is printed correctly. Related to #646.


0lvin (approved):
Good catch :-) Thank you :-)

Yamagi:
And merged :)

PR #657, opened 2021-01-24
Fullscreen Vulkan fixes

### LINKED ISSUE #648: Vulkan renderer  scaling issues  when  vid_fullscreen = 1
vulkan renderer uses only smaller portion of available screen for game content when vid_fullscreen =1 and r_mode  is not desktop resulution
the console, rendered movies and the HUD is as expected, only the actual rendered image is way to small.

configure game resolution smaller than desktop resolution  (e.g.  game 1920x1080 /  desktop 2560x1440 )  and set vid_fullscreen = 1

1920x1080
q2_0002

1280x960
q2_0003

1024x768
q2_0004

rg3:
This branch fixes a number of problems present in Vulkan when in fullscreen mode, and other related problems.

* First off and most importantly, it fixes #648, making fullscreen images look correctly and have the right aspect ratio when vid_fullscreen is 1.
* Second, it restores textured borders around the world view when scaling the view down.
* Finally, it center the game "window" in the screen with vid_fullscreen 1, which should be more comfortable.

Edit: edited to reflect branch updates.


src/client/refresh/vk/vk_draw.c
```diff
@@ -169,6 +169,9 @@ refresh window.
 */
 void RE_Draw_TileClear (int x, int y, int w, int h, char *name)
 {
+	// This is handled in a special way in the Vulkan renderer. Clearing tiles
+	// are skipped and replaced by black bars during the UI render pass.
+#if 0
```
0lvin, line 174: @Yamagi Do we need clear function? Has this function used only for clean up borders?

rg3: While analyzing a frame with renderdoc, I noticed these borders were being drawn in the wrong place at the beginning of the frame and, in any case, black borders were being drawn around the main world view later as part of the UI, so I think it's just not useful and commented it out apparently without visual consequences. :shrug: 

Maybe in the future the black borders from the UI can be replaced by this kind of tile?

0lvin: Maybe it related to my hack with [R_CleanupBorders]( https://github.com/yquake2/yquake2/blob/46c24d4/src/client/refresh/vk/vk_rmain.c#L1176). 

Could you please try to comment out this function and rerun? Maybe RE_DrawTileClear should do exactly such but it has regressed in some moment of time.

0lvin: It looks exactly such by debug prints in RE_DrawTileClear. :-(

When I have removed R_CleanupBorders in master code - borders is painted by correct placeholder. 

Could you please remove R_CleanupBorders at all?

rg3:
I've reverted part of the changes with the latest commit. After the minor fix that precedes it, we no longer need to clean the frame when the refdef changes sizes.

src/client/refresh/vk/vk_draw.c
rg3: Sure, let me try to see what comes out and I'll report if there's any situation in which it's broken. We could handle those in a separate pull request.

rg3: So from what I've seen, the border is broken with vk_pixel_size and, at least in the state of this branch, it's also being drawn in the wrong place with vid_fullscreen 1. It will need some more work before it can be fully restored.

rg3:
I've also detected taking screenshots is broken in this branch. I'll try to fix those as well.

src/client/refresh/vk/vk_draw.c
rg3: Fixed.

rg3:
I'm mostly done. I've detected a missing black bar in the bottom of the screen when inside some option menus, so I'll try to pinpoint the cause and fix that too.

rg3:
Nevermind. It seems to happen also in OpenGL and Software, so I consider this ready for review again.

For the black bar problem, see the bottom of the screen in the two screnshots below:

q2_0012
q2_0015

0lvin (approved):

Yamagi:
> Nevermind. It seems to happen also in OpenGL and Software, so I consider this ready for review again.
The black bar is likely drawn somewhere(tm) in menu.c.  It might have something to do with the status bar used in some menu. But I don't know.

Nevertheless. This is very good work. Keep it up! :)

PR #658, opened 2021-01-31
More Vulkan fixes

rg3:
A few more Vulkan fixes for things I noticed this week.

* Textured borders for scaled down views were not working properly with MSAA enabled, due to the texture quad pipeline being prepared only for single-sample attachments (it's mainly used for drawing UI elements in the UI render pass, which is always single-sampled). Could this have been the original reason for the black borders hack?

* Framebuffer contents from the previous frame (used to mask geometry glitches) were not being correctly loaded with MSAA enabled. Geometry glitches, as seen in the skybox, are not as noticeable with MSAA enabled so I didn't notice this until I analyzed a frame with renderdoc. This is a mistake I introduced earlier and its solution also increases framerate with MSAA enabled. :shrug: In my GTX 1070 I jump from 375 fps to ~440 fps in 1600x1200 with MSAA 8x.

* Given the Vulkan backend uses 3 render passes with a *single* subpass each, specifying barriers using subpass dependencies felt weird to me, so I've replaced them with proper memory barriers between render passes. I believe they're easier to understand and reason about.


Yamagi:
> In my GTX 1070 I jump from 375 fps to ~440 fps in 1600x1200 with MSAA 8x.

And from 629 fps to 343 fps on my Radeon 5700XT with Mesa3D.

@0lvin, you're okay with this?

rg3:
> And from 629 fps to 343 fps on my Radeon 5700XT with Mesa3D.

Oh, you mean MSAA regresses in performance for you with these changes?

Yamagi:
Arg. That was a typo, I meant  small increase from 629 to 643.

turol:
I think Vulkan recommended behavior is to use subpass dependencies instead of explicit memory barriers, the driver might be able to schedule them better. Also using global memory barries like you're doing (instead of image or buffer barriers) is a rather heavy hammer and might cost some performance on some devices.

Yamagi:
How do we proceed here?

rg3:
I'll look into it, no problem. I think @turol has a point and I need to review the spec in detail. External subpass dependencies are used in Vulkan to put memory barriers controlling layout transitions when the initial image layout doesn't match the layout used in the subpass, and when the layout during the subpass does not match the final image layout. This does happen here, so we probably need them in some form. In addition, we may have to keep some of the extra barriers I'm adding in this PR.

I know I promised to take a look at some of this stuff last week, including crashes reported in another issues, but I've been a bit busy due to pandemic-related matters. I'll take a look ASAP.

rg3:
So I've changed the branch a bit restoring and tuning the previous external subpass dependencies and keeping the convenient Qvk_MemoryBarrier function. I've studied the spec in more detail and I could even write a blog post explaining my reasoning for the changes, because Vulkan synchronization is not easy to explain in a few words. The summary would be:

First, as I mentioned above, external subpass dependencies are used to control layout transitions from the initial layout of the attachment in the render pass to the layout in the first subpass that uses the attachment, and from the last layout to the final layout specified in the render pass. In our case, these layout transitions happen and must be controlled so images are not transitioned before all operations have completed on them, creating a race. Most of my initial changes have been reverted because the memory barriers I introduced, while correct from the point of view of stages and data dependencies, do not allow us to control these layout transitions.

Second, the first external subpass dependency has been modified. After the skybox fixes we now want to load the color buffer contents from the previous frame before drawing on it again, and this was not correctly taken into account. For the first subpass dependency (used from previous render passes to rendering the world), we need to transition the color attachment before color attachment read operations (i.e. the attachment load) are performed, so I've added that access to the dependency.

Third an finally, a remark on global memory barriers and the question of them being a heavy hammer: I've seen several people, most notably Axel Gneiting from id Software who worked on the renderers for the new Doom games, mention in the Khronos Slack that they use memory barriers for synchronization as the general mechanism, and only fall back to image memory barriers when they want to also perform an image layout transition. While a global memory barrier does look like a heavy hammer, they're much easier to use and apparently it's standard industry practice, so I wouldn't worry too much about it if you want to use them somewhere else.

Sorry for the wall of text. Hope this clears things up. :smile: 

PS: Please do not merge yet. I'll add more changes to fix #659.

rg3:
Pushed more changes to at least partially fix #659, please review.

Yamagi:
> I know I promised to take a look at some of this stuff last week, including crashes reported in another issues, but I've been a bit busy due to pandemic-related matters. I'll take a look ASAP.

I didn't want to chase you into action. It's not, that I'm having plenty of time to spend for Q2. It's a hobbyist project, when things take a little longer, it's totally okay.

> Sorry for the wall of text. Hope this clears things up. smile

I does. :) I don't have opinion regarding this. Or better said I don't have the Vulkan knowledge necessary to form an opinion. All I can say is, that these changes give a nice performance improvement on:

* Intel under Linux.
* Nvidia under Windows.
* Nvidia behind Optimus under Linux.
* Radeon with ACO under Linux.

@DanielGibson, can you gave it a try with your Radeon under Windows and maybe Intel on Windows? We want to know if there any performance regressions between this branch and master. I don't think so, but better be save than sorry.

> Pushed more changes to at least partially fix #659, please review.

I can confirm that the first part of #659 (missplaced window) is fixed by these changes. The second part isn't. I'll update the bug report in a few minutes.

rg3:
I've pushed one more commit to fix the validation errors you see in every frame and reported in #659. Those were not showing with the version of the layers present in my system (Fedora 33), so I hadn't caught them yet. I built the layers from source (version 1.2.170) and they started showing up. Thanks! I'll try to take a look at the crashes when Alt+Tabbing in the following days.

Yamagi:
An I confirm, that the error is gone. Good work. :) What about merging this pull request now and open a new one for further fixes?

rg3:
Sounds good to me. 👍

Yamagi:
A new Windows testbuild with these changes can be found here: https://deponie.yamagi.org/quake2/windows/testbuilds/quake2-20210226-d713944.zip

0lvin:
@rg3 After merge looks like we have new warning message with vk_validation=2, could you please check?

```
VK_WARNING: Validation Performance Warning: [ UNASSIGNED-BestPractices-vkCreateRenderPass-image-requires-memory ] Object 0: handle = 0x62800001c100, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x4003982 | Attachment 2 in the VkRenderPass is a multisampled image with 4 samples, but it uses loadOp/storeOp which requires accessing data from memory. Multisampled images should always be loadOp = CLEAR or DONT_CARE, storeOp = DONT_CARE. This allows the implementation to use lazily allocated memory effectively. (performance)
```

rg3:
> @rg3 After merge looks like we have new warning message with vk_validation=2, could you please check?
> 
> ```
> VK_WARNING: Validation Performance Warning: [ UNASSIGNED-BestPractices-vkCreateRenderPass-image-requires-memory ] Object 0: handle = 0x62800001c100, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x4003982 | Attachment 2 in the VkRenderPass is a multisampled image with 4 samples, but it uses loadOp/storeOp which requires accessing data from memory. Multisampled images should always be loadOp = CLEAR or DONT_CARE, storeOp = DONT_CARE. This allows the implementation to use lazily allocated memory effectively. (performance)
> ```

That's a consequence of the skybox fixes. To hide geometry errors in levels and the way the skybox coordinates are calculated, creating geometry that's not watertight, I made the game load the contents of the previous frame before drawing the next one on it, visually masking many errors in the same way the OpenGL renderer does. These changes mean the color attachment (single sample or multisample) has to be loaded when the render pass starts and always stored after drawing, hence the warning. Nothing to worry about, IMHO.

Issue #661, opened 2021-01-31
software renderer stretches image ignoring aspect ratio  in vid_fullscreen 1


j4reporting:
with vid_fullscreen 1  the sw renderer stretches the selected game resolution to full desktop resolution. This is great, but it is ignoring the aspect ratio.

e.g.  desktop  is 2560x1440 
game resolution is 1600x1200

Yamagi:
I can reproduce that. @0lvin, can you have a look?

0lvin:
By first look, if selected resolution has different aspect ratio, final image is stretched to full screen. That looks correct because `1` is render without change resolution.

I think we can ignore selected resolution and scale resolution to nearest resolution with same aspect ratio as the real screen resolution.

Yamagi:
I've merged the pull request. I think this is fixed. If not feel free to reopen the issue.

PR #663, opened 2021-02-06
Soft render updates

0lvin:
#661 - save buffer size and scale selected resolution to desktop aspect ratio.


Yamagi:
Looks good! :) Thank you for fixing this.

PR #665, opened 2021-02-17
Fix segmentation fault on FreeBSD.

shamazmazum:
Specify correct protection flags when calling mmap() on FreeBSD.
prot | PROT_MAX (prot) are the correct flags because just PROT_MAX
(prot) leaves current protection flags set to PROT_NONE which causes
segmentation fault when reading or writing to the mapped region.


Yamagi:
I wonder why I've never stumbled into that crash. Thank you!

Issue #659, opened 2021-01-31
Problems with vid_fullscreen 1 and Vulkan


Yamagi:
Hi,
while investigating how we could make the `vid_fullscreen` cvar a little bit more intuitive I've stumbled over some problems with `vid_fullscreen 2` (switch resolution) in the vulkan renderer.

1) When switching into a lower resolution then the game was started with the image is missplaced in the lower right corner. It doesn't happen when switching to a higher resolution. This is easy to reproduce: Start the game with `vid_fullscreen 2` and 2560x1440 -> switch to 1920x1080 -> image gets misplaced. As an educated guess this might have something to do with the viewport or the scissor not getting updated. Screenshots are looking correctly.

2) The game crashed several times when changing the resolution or reinitalizing the renderer. So far I haven't found a reliable way to reproduce this. And the crash is always in  /usr/lib/libvulkan_radeon.so, called from different places throughout the renderer. So this could be a driver problem. Do this to trigger it sooner or later: Open some window, a terminal or browser for example -> start the game with `vid_fullscreen 2` and a resolution lower than the desktop -> use alt-tab to switch between Q2 and the other windows -> When entering or leaving Q2 the resolution is changed. Sometimes the game crashes when changing the resolution.

One example backtrace:

```
Thread 1 "quake2" received signal SIGSEGV, Segmentation fault.
0x00007fffa82c57ad in ?? () from /usr/lib/libvulkan_radeon.so
(gdb) bt
#0  0x00007fffa82c57ad in ?? () from /usr/lib/libvulkan_radeon.so
#1  0x00007fffa82ca426 in ?? () from /usr/lib/libvulkan_radeon.so
#2  0x00007fffa82ca59e in ?? () from /usr/lib/libvulkan_radeon.so
#3  0x00007fffd00e5240 in QVk_DrawColorRect (ubo=0x7fffffffe400, uboSize=32, rpType=RP_UI)
    at src/client/refresh/vk/vk_common.c:2462
#4  0x00007fffd00e7378 in RE_Draw_Fill (x=0, y=0, w=1680, h=1050, c=0)
    at src/client/refresh/vk/vk_draw.c:229
#5  0x00005555555b7a3d in Draw_Fill (x=0, y=0, w=1680, h=1050, c=0) at src/client/vid/vid.c:655
#6  0x00005555555956ae in M_Draw () at src/client/menu/menu.c:4958
[...]

PR #676, opened 2021-03-07
Avoid restarting the whole video on VK_ERROR_OUT_OF_DATE_KHR

rg3:
While attempting to solve #659 I started the game in fullscreen mode under Gnome and immediately discovered the game entered an infinite renderer restart loop. A few seconds after the renderer had been initialized and was already drawing frames, it received VK_ERROR_OUT_OF_DATE_KHR, which triggered a restart of the whole video system. The game flickered while restarting it (including creating a new game window, I guess) and the story repeated itself a few seconds later.

It appears this is a known issue and also affects or affected [vkQuake](https://github.com/Novum/vkQuake/issues/211) and [vkQuake2](https://github.com/kondrak/vkQuake2/issues/95) and @kondrak solved some months ago by restarting Vulkan without touching the game Window. This pull request implements a similar solution, simplified (so probably incomplete :stuck_out_tongue:) and adapted to yquake2.

This PR can be considered work-in-progress and testing is welcome. At least it solves the infinite reload problem for me. Hopefully, but unlikely, it also solves #659. While doing some back-and-forth switching from "keep resolution" to "switch resolution" and vice versa, I also made the running demo "crash" to the game console with:

```
********************
ERROR: Mod_PointInLeaf: bad model
********************
```

And in those situations I also get the following validation error:

```
==== ShutdownGame ====
VUID-VkPresentInfoKHR-pImageIndices-01296(ERROR / SPEC): msgNum: -945112042 - Validation Error: [ VUID-VkPresentInfoKHR-pImageIndices-01296 ] Object 0: handle = 0x3b83d28, type = VK_OBJECT_TYPE_QUEUE; | MessageID = 0xc7aabc16 | vkQueuePresentKHR(): Images passed to present must be in layout VK_IMAGE_LAYOUT_PRESENT_SRC_KHR or VK_IMAGE_LAYOUT_SHARED_PRESENT_KHR but is in VK_IMAGE_LAYOUT_UNDEFINED. The Vulkan spec states: Each element of pImageIndices must be the index of a presentable image acquired from the swapchain specified by the corresponding element of the pSwapchains array, and the presented image subresource must be in the VK_IMAGE_LAYOUT_PRESENT_SRC_KHR layout at the time the operation is executed on a VkDevice (https://github.com/KhronosGroup/Vulkan-Docs/search?q=)VUID-VkPresentInfoKHR-pImageIndices-01296)
    Objects: 1
        [0] 0x3b83d28, type: 4, name: NULL
```

So my guess is there's still work to do.

### DIFF
--- a/src/client/refresh/vk/header/qvk.h
+++ b/src/client/refresh/vk/header/qvk.h
@@ -260,6 +260,8 @@ extern qvktexture_t vk_colorbuffer;
 extern qvktexture_t vk_colorbufferWarp;
 // indicator if the frame is currently being rendered
 extern qboolean vk_frameStarted;
+// Indicates if the renderer needs to be restarted.
+extern qboolean vk_restartNeeded;
 
 // function pointers
 extern PFN_vkCreateDebugUtilsMessengerEXT qvkCreateDebugUtilsMessengerEXT;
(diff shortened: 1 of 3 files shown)


Yamagi:
> Hopefully, but unlikely, it also solves #659.

At least I was unable to crash the game with switching the resolution for about 25 times. That's much better than without this fix.

I'll give it some mote testing tomorrow.

Yamagi:
I gave it some more testing. I'm pretty sure that the crash is fixed. So I'm merging this.

> I'll probably try to find the cause for the Mod_PointInLeaf error in the near future.

Now that I know that the problem exist and how I can trigger it, it's easy to reproduce. I've never seen that error before, though.

> I also want to play the full game with the Vulkan renderer to see if I detect some other issues.

I recently did a full run of baseq2, xatrix and rogue on commit 6f045f71afc9df5a705e70a5893c82b203b9483a. I didn't observe any render glitches, at least nothing obvious.

PR #678, opened 2021-03-10
Fix fullscreen value in hacks (f1b6b827)

0lvin:


0lvin:
@Yamagi Could you look? Should we remove this hacks, what do you think? Look as it is not required after set `r_mode` to -2
for fullscreen.

Yamagi:
I think we should keep them. While the menu forces `r_mode -2` when selecting 'fullscreen window' the user can still set `vid_fullscreen 2` with any `r_mode` through the console. And there's the corner case that the game might be unable to determine the desktop resolution.

Anyways, thank you for fixing my oversights.

Issue #647, opened 2021-01-16
Windows: vulkan renderer,  fullscreen changed to 0  when switching to desktop via CONTROL+ ESC


j4reporting:
System is windows 10 using latest  pre 8.0 build from Jan 13. 
 Intel iGPU Iris 540 (Skylake)  

`Vulkan API: 1.2
Header version: 148
Devices found:
   #0: Intel(R) Iris(R) Graphics 540
Using device #0:
   deviceName: Intel(R) Iris(R) Graphics 540
   resolution: 2560x1440
   driverVersion: 1647081 (0x1921E9)
   apiVersion: 1.2.148
   deviceID: 6438
   vendorID: 0x8086 (Intel)
   deviceType: VK_PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU
   gfx/present/transfer: 0/0/0
Present mode: VK_PRESENT_MODE_FIFO_KHR
Swapchain image format: 44
Supported present modes: VK_PRESENT_MODE_IMMEDIATE_KHR VK_PRESENT_MODE_FIFO_KHR 
Enabled extensions: VK_KHR_surface VK_KHR_win32_surface 
Enabled layers: 
Reallocate space for 512 models.
Successfully initialized Vulkan!
Successfully loaded ref_vk.dll as rendering backend.
`

by trying to switch back to Desktop ( Control + ESC ) and leaving yquake2 running off-screen, the vid_fullscreen variable changes to 0 as well.
Using the GL1 or GL3 renderer this works as expected.

EDIT:  s/fullscreen/vid_fullscreen/

Yamagi:
I cannot test it right now, but this may be fixed in the latest build: https://deponie.yamagi.org/quake2/windows/testbuilds/quake2-20210310-2a18f02.zip

j4reporting:
I was about to give an update on this. unfortunately it's got worse.  With SDL2.12  vulkan renderer now crashes when switching to desktop by either  ALT+TAB, left window key  or CTRL + ESC.

Faulting application name: yquake2.exe, version: 0.0.0.0, time stamp: 0x00000000
Faulting module name: ntdll.dll, version: 10.0.18362.1411, time stamp: 0x9ace5a5f
Exception code: 0xc0000005
Fault offset: 0x0004dbcd
Faulting process id: 0x269c
Faulting application start time: 0x01d715bdcb7b1d81
Faulting application path: E:\GOG Games\yquake2_pre\yquake2.exe
[...]

PR #680, opened 2021-03-21
Vulkan restart tweaks

rg3:
This pull request essentially fixes two issues:

* ERROR: Mod_PointInLeaf: bad model when restarting the Vulkan renderer. This is solved by making the restart happen at the end of the frame instead of at the beginning of it, and signaling the server the client needs to be re-registered again to rebuild the list of models.

* Memory leak when restarting the Vulkan renderer, due to "models_known" not being freed before allocating it again.

For the first point I used the same approach taken in vkQuake2 of using a global cvar_t called vid_refresh. It's my understanding, from reading other parts of the code, that such an approach is a bit hackish and there should be a message sent from the client  to the server to properly communicate this, but my knowledge of the code base is lacking here, so I preferred to get the fix out and let others with more experience tune it as needed.

### DIFF
--- a/src/client/vid/vid.c
+++ b/src/client/vid/vid.c
@@ -286,6 +286,7 @@ VID_GetModeInfo(int *width, int *height, int mode)
 // Global console variables.
 cvar_t *vid_gamma;
 cvar_t *vid_fullscreen;
+cvar_t *vid_refresh;
 cvar_t *vid_renderer;
 
 // Global video state, used throughout the client.
@@ -489,6 +490,12 @@ VID_CheckChanges(void)
 		// Unblock the client.
 		cls.disable_screen = false;
 	}
+
+	if (vid_refresh->modified)
+	{
+		vid_refresh->modified = false;
+		cl.refresh_prepped = false;
+	}
 }
 
 /*
@@ -500,6 +507,7 @@ VID_Init(void)
 	// Console variables
 	vid_gamma = Cvar_Get("vid_gamma", "1.0", CVAR_ARCHIVE);
 	vid_fullscreen = Cvar_Get("vid_fullscreen", "0", CVAR_ARCHIVE);
+	vid_refresh = Cvar_Get("vid_refresh", "0", CVAR_NOSET);
 	vid_renderer = Cvar_Get("vid_renderer", "gl1", CVAR_ARCHIVE);
 
 	// Commands
--- a/src/client/vid/vid.c
+++ b/src/client/vid/vid.c
@@ -286,7 +286,6 @@ VID_GetModeInfo(int *width, int *height, int mode)
 // Global console variables.
 cvar_t *vid_gamma;
 cvar_t *vid_fullscreen;
-cvar_t *vid_refresh;
 cvar_t *vid_renderer;
 
 // Global video state, used throughout the client.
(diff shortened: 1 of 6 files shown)


rg3:
Note: I can still reproduce a crash when Alt+Tabbing out of fullscreen to the desktop and back under Gnome. This is probably related to #647 but does not attempt to solve it. My main goal was to fix the "ERROR: Mod_PointInLeaf: bad model" problem. I'll try to take a look at the crash in the following days.

src/client/vid/vid.c
```diff
@@ -489,6 +490,12 @@ VID_CheckChanges(void)
 		// Unblock the client.
 		cls.disable_screen = false;
 	}
+
+	if (vid_refresh->modified)
```
0lvin, line 494: Could we use some boolean flag instead cvar `vid_refresh` ? Look like exact value of variable is never used and only checked `modified` status.

rg3: Can you clarify what you mean with a boolean flag? If you mean something like a global variable, I don't think that would be any more practical. Since the renderer is dynamically loaded with dlopen(), the main program would have to get the address of such a global variable from the renderer lib. The cvar approach, copied from vkQuake2, is simple in the sense that it allows the variable to be shared between the renderer and the server with ease using the existing mechanisms without modifying the interface.

0lvin: I see, thank you.

0lvin (approved):

src/client/vid/vid.c
Yamagi: Can't we just use `vid_renderer` for this? We renamed `vid_refresh` to `vid_renderer` when we changed the renderer API to make sure that the renderer selection is empty / at default when starting with an old config.

Yamagi:
> For the first point I used the same approach taken in vkQuake2 of using a global cvar_t called vid_refresh. It's my understanding, from reading other parts of the code, that such an approach is a bit hackish and there should be a message sent from the client to the server to properly communicate this, but my knowledge of the code base is lacking here, so I preferred to get the fix out and let others with more experience tune it as needed.

Abusing a scvar is a little bit hacky, but the vanilla client does it the same way. The alternative would be to fiddle another shared function pointer into the renderer API. That would break external renderers.

src/client/vid/vid.c
rg3: I have tested changing the code to use vid_renderer instead of introducing a new variable and it mostly works. The only unpleasant side-effect of this change is that the client is re-registered just after changing vid_renderer from the console, before they have the chance to use vid_restart. This does not happen when using the video menu, since changing the renderer and hitting "apply" triggers a renderer restart immediately. I'll push the commit so you can check. If you prefer the status quo, we can just drop the new commit.

Yamagi: I think we take this way, because it's the least invasive option. Reintroducing `vid_refresh` would confuse users and adding a new function might break the renderer API. We can evaluate the new function at a later time, let's get this into master first. I'll give a last round of tests later today and will merge it after that if nothing comes up.

Yamagi:
And finally merged, thank you!

src/client/vid/vid.c
Rone-s9: Hey. F.c.ing with ME isn’t AT ALL a smart thing to do;). Not anymore…enough is enough. (Boolean-is he using steroids….). I’m soon 42 and then 31y. with PC. DO NOT UNDERESTIMATE ME…aaaaaaanymore. My time has ARRIVED. Just need my soon new MacBook Pro. Can’t do anything because hacked since 1994….AND NOW: ME TIME…MELTING TIME…FINALLY. Madness NB: The tag should say the most lol.

Issue #662, opened 2021-01-31
r_anisotropic:  lowest possible setting differ in GL and Vulkan


j4reporting:
In vulkan renderer  r_anisotropic  0  is accepted, and it's needed with Intel Windows drivers, because any setting > 0 enables texture filtering, ignoring VK_TEXTUREMODE. 

the GL renderer resets r_anisotropic 0  to 1.

rg3:
This one should be fixed now after #681.

PR #681, opened 2021-03-21
Vulkan anisotropic filter tweaks

rg3:
My original goal with these changes was just fixing #662, but I've fixed a couple of additional issues.

First, I've tweaked the code so changing the anisotropic filter value from the video menu does not apply the change immediately. Instead, now you must hit "Apply" and let the renderer restart. The problem this tries to solve is that, in Vulkan, the anisotropic and texture filter modes are part of the texture sampler object. If you want to support any combination of these parameters independently, you only have two options:

* You create a huge number of texture samplers with every combination of texture and anisotropic filter values, and select the one to apply on the fly based on user settings. This is what's used now, but to cut down on the number of combinations the anisotropic filter value in Vulkan is now an on/off switch, and the maximum anisotropy is applied if not off. I wanted to change that.

* The second approach is to carefully rebuild texture samplers on the fly when the user requests a change in texture sampling parameters, but this has to be done very carefully. In theory, you must not delete a texture sampler that's already recorded in a command buffer before the command buffer finishes executing and is destroyed. This is not too hard, but the current state of the code doesn't make this easy.

Because video options are changed seldomly, I don't think it's a problem to restart the renderer when the anisotropic filter value is changed. When the renderer starts, texture samplers are created with the appropriate anisotropy value chosen by the user, so it's not longer an on/off switch and we also cut the number of texture samplers almost in half.


0lvin (approved):

Yamagi:
Good work! Thank you!

PR #688, opened 2021-03-31
Add a cvar `g_commanderbody_nogod` to make the com. body destructable.

carlos-cpp:
The commanders body entity is special, because it's spawned in god mode.
That's no problem in the baseq2 and addons campaigns. But it may break
some custom maps and prevents some hacks, one example is putting the
entity inside an killbox.

Submitted by Тихомиров Eвгений.

Yamagi, please add 
self->die = commander_body_die;
on line 1674  src/game/g_misc.c
cause if not it will fault the game

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -144,6 +144,10 @@ it's `+set busywait 0` (setting the `busywait` cvar) and `-portable`
   set to `0` footstep sounds are never generated. Cheat protected to
   `1`.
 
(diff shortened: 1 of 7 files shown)


carlos-cpp:
Ok, I have to agree.
I will mention to set c_var in my code changes.
I have tested new feature in all cases. We shouldn't fail the game structure.
So, you can commit to master branch.
Thanks, for your comments.

carlos-cpp:
reopen pull request to commit

Yamagi:
I've merged the change into master.

PR #690, opened 2021-04-01
game data packing representation of the headers

devnexen:


DanielGibson:
I'd suggest using a name like `savegameHeader_t` or similar instead of `savegame_t`, to make its purpose/meaning more clear.

devnexen:
right that s just the headers after all :)

Yamagi:
Can you do the same for xatrix and rogue? We try to keep baseq2 and bot official addons ins sync to make merging changes between them easier.

DanielGibson:
ideally also zaero (the last time I changed something in the savegames I also did it there so it should be in about the same state as the official addons)

devnexen:
ok done ... did not notice these repos until today :-) I guess that s too much to refactor in one repo like most of doom projects do.

Yamagi:
> I guess that s too much to refactor in one repo like most of doom projects do.

And the licenses are incompatible. Quake II itself is under GPL2, the addons and most 3rd party mods are under the crappy SDK license. Some projects have merged baseq2, ctf, rogue and xarix into one. We won't do it, because it violates the SDL license and id Software, ZeniMax, Microsoft, whoever may not like that.

PR #691, opened 2021-04-01
unix network little change proposal. socket port fits within unsigned…

devnexen:
… 16 bits type.


Yamagi:
Every byte is precious, don't waste them. ;) Thank you!

PR #693, opened 2021-04-03
WIP: enable neon vector support for the gl3 renderer.

devnexen:
using neon aligned data to support older compilers too.
most of the intrinsics would work on 32 bits too.

Sort of "call for testers" rather than to be merged necessarily
 unless it brings benefits.

### DIFF
--- a/src/client/refresh/gl3/header/HandmadeMath.h
+++ b/src/client/refresh/gl3/header/HandmadeMath.h
@@ -20,15 +20,15 @@
   
   =============================================================================
   
-  To disable SSE intrinsics, you MUST
+  To disable SIMD intrinsics, you MUST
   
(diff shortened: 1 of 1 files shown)


0lvin:
Look like code should be updated [in first place](https://github.com/HandmadeMath/Handmade-Math/blob/master/HandmadeMath.h) and reimported.

devnexen:
Ah that was thirdparty dependency good to know :-) but looking further I see there is already ongoing work https://github.com/HandmadeMath/Handmade-Math/issues/79 :-)

Yamagi:
Yeah, this should be done upstream and imported in YQ2 by upgrading the file to the new upstream version. Changing the code on our side means, that we must merge our changes with the upstream changes on every upgrade.

Issue #673, opened 2021-03-03
Windows, Vulkan: map Main Gate  water texture drawn incorrectly when player rotates/moves


j4reporting:
Map Main Gate,  enter first door  and look down at the water below. Then start to rotate 
Using latest windows test build 

![q2_0007](https://user-images.githubusercontent.com/45177907/109738131-8da45e80-7bc7-11eb-81cd-ba342805be79.jpg)

Edit:  this is caused by a texture from https://deponie.yamagi.org/quake2/texturepack/q2_textures.zip   
   r_retexturing 0   is fine

Yamagi:
There's also #670. Looks like vk still has some problem with custom game data.

Yamagi:
This should be fixed in the latest test build: https://deponie.yamagi.org/quake2/windows/testbuilds/quake2-20210407-26bda18.zip

j4reporting:
I can confirm that this is indeed fixed. Kudos to all involved.

Issue #670, opened 2021-02-27
Render `chicken` model failed in vulkan render.


0lvin:
Buffer overflow is raised when render `chicken` model from [custom player models](http://tastyspleen.net/quake/downloads/all-players.zip)

```
==27383==ERROR: AddressSanitizer: global-buffer-overflow on address 0x7f2557d0a35c at pc 0x7f2557b9f506 bp 0x7ffff5102ec0 sp 0x7ffff5102eb8
WRITE of size 4 at 0x7f2557d0a35c thread T0
    #0 0x7f2557b9f505 in Vk_DrawAliasFrameLerp src/client/refresh/vk/vk_mesh.c:237
    #1 0x7f2557ba5e2b in R_DrawAliasModel src/client/refresh/vk/vk_mesh.c:764
    #2 0x7f2557bb6fe2 in R_DrawEntitiesOnList src/client/refresh/vk/vk_rmain.c:349
    #3 0x7f2557bc184b in RE_RenderView src/client/refresh/vk/vk_rmain.c:977
    #4 0x7f2557bc286e in RE_RenderFrame src/client/refresh/vk/vk_rmain.c:1130
    #5 0x55b994ecf9c6 in R_RenderFrame src/client/vid/vid.c:589
    #6 0x55b994e72cdf in PlayerConfig_MenuDraw src/client/menu/menu.c:4803
    #7 0x55b994e73b5d in M_Draw src/client/menu/menu.c:4966
    #8 0x55b994e44cf6 in SCR_UpdateScreen src/client/cl_screen.c:1565
    #9 0x55b994e2faf8 in CL_Frame src/client/cl_main.c:838
    #10 0x55b994ef6f87 in Qcommon_Frame src/common/frame.c:642
    #11 0x55b994ef5a77 in Qcommon_Mainloop src/common/frame.c:175
    #12 0x55b994ef6476 in Qcommon_Init src/common/frame.c:393
    #13 0x55b994f927a6 in main src/backends/unix/main.c:145
    #14 0x7f257d486d09 in __libc_start_main ../csu/libc-start.c:308
    #15 0x55b994df4f19 in _start (/home/denis/Desktop/id-test/q2/quake2-build/quake2+0x58f19)

0x7f2557d0a35c is located 4 bytes to the left of global variable 'shadowverts' defined in 'src/client/refresh/vk/vk_mesh.c:337:16' (0x7f2557d0a360) of size 24576
[...]

PR #694, opened 2021-04-04
Vulkan code improvements and cleanup

0lvin:
* Disable blend for water(transparent) surfaces in world rendering #673
* Reallocate vert buffers before use #670
* Fix warning on attachment creation
* simplify mesh rendering in model code
* full size retextured images (it_pic) in ref_soft


Yamagi:
Look good :) Thank you!

PR #695, opened 2021-04-04
Little optimisation attempt targeted only for ARM 32 bits,

devnexen:
enabling "RunFast" mode ; benefits might depend on the board
 at hand.


Yamagi:
Thank you!

PR #696, opened 2021-04-07
Fix segfault when turning the console before the game is launched

devnexen:
(thus audio backend not yet initialised).


Yamagi:
Thank you! :)

Issue #689, opened 2021-03-31
Juggernaut, area inescapable or inaccessible, and one enemy missing


Froyd1:
I want to give a heads up in this level of Juggernaut, that this area in The Heart Of Darkness level, is inescapable or inaccessible, without cheats. Also a missing enemy which is this one:

`monster_soldier_ss: -720.000000 -504.000000 -319.000000`

In this video I show you where the problem is, and it shows the enemy is stuck in the wall underwater somewhere, shooting at you as your health declines. I'm not sure this major bug can be fixed, as it seems this expansion pack was rushed to begin with.

https://www.youtube.com/watch?v=eXevvAWs5og

Yamagi:
And the source code was never released. It might be possible to fix the maps, but everything that requires source code changes is impossible. :/

Froyd1:
Oh I see, that's a shame about the source code not being released for this expansion pack. Do your best to fix the maps and bugs for Juggernaut. Thanks for your hard work. :)

DanielGibson:
What Yamagi meant is: We only fix such bugs in addons/mods that we have the source code of (like the official addons or maybe Zaero), we don't work on other addons and generally avoid modifying any game data.

BjossiAlfreds:
As far as I know Juggernaut didn't ship with a gamex86.dll, it is purely a map pack with some reskins of Quake 2 assets. At least that is the case for the CD copy I own that I bought back in 1999. So we could actually just fork or use the baseq2 game.dll for it.

Though to be honest here, is this mission pack even worth salvaging? :-)

DanielGibson:
Hmm I thought it had its own gamex86.dll, but I'm not sure at all and I don't think I've ever played that addon (I heard it's not very good).
Maybe the gamex86.dll was available as/in a patch?
I found this patch: http://www.markshan.com/thesinraven/juggernaut_patch_06-Jul-99.htm but I wasn't able to extract any useful data on Linux and didn't try to actually install it on Windows.

Also, according to https://ogdb.eu/index.php?section=article&articleid=470 there were apparently two versions of Juggernaut: A "Deluxe Edition" that was a scam (only contained 11 levels instead of the full 21), installed in the `JUG/` directory which was 46.9MB, and the normal/proper/full version, which contained 21 levels, installed into `JUGFULL/` and was about 133MB.
[...]

Issue #687, opened 2021-03-29
Missing enemy triggers on "Research Lab" [skill: easy]


moobaa:
After collecting blue key & pressing the keyboard to activate the Maintenance Bridge, the Medic that is below the floor in the room with the Bridge activation switch does not trigger.

Also, the two Parasites in the roof that should be dropped when rounding the nearby corner (with the three Brains) don't get dropped.

All three enemies remain in the map, in their little hidey-holes ;)

Have only noticed this on Easy; will update if it also happens on higher difficulty levels.

BjossiAlfreds:
There is a trigger_once with target t308 that won't spawn on Easy, it is supposed to get the medic into the level for you to fight. It was probably an oversight of the level creator to not set the spawnflags of those monsters correctly.

BjossiAlfreds:
Here is a fix for the lab level, extract to baseq2/maps. See if the monster kill counts match on easy and medium.
[lab.zip](https://github.com/yquake2/yquake2/files/6225365/lab.zip)

moobaa:
That worked (on Easy; I haven't tried Medium) - thanks :)

I get your point about map errors being outside the scope of the project; I just couldn't recall any such discrepancies back-in-the-day.

Should I close this item now? Or leave it (and re-frame it) as a catch-all for any other map issues?

Yamagi:
We could ship the entity files somewhere under stuff/. A pull request would be appreciated. And - I know, I sound demanding - it would be nice if someone could have a look at the monster bug in baseq2: https://github.com/yquake2/yquake2/blob/master/src/game/g_spawn.c#L660

moobaa:
@Yamagi  - I don't know whether your "pull request" comment was directed to @BjossiAlfreds or myself :} I'm happy to keep my eyes out for other errors & can endeavour to provide fixes, though my knowledge of Q2 scripting is non-existent. The code (& entity script) was pretty straightforward, though, and I suspect that a Q2 map editor will reveal entities and triggers :)

I have just encountered another issue: Inner Hanger, there's a Berzerker in a room at the level start that never gets triggered on Easy. Here's the "listentities monsters" that I remembered to grab this time. Let me know if these map errors should be broken into their own Issues, or just kept under one issue.

Thanks for the hard work!

![image](https://user-images.githubusercontent.com/81551227/112939925-0ae7d280-9178-11eb-88ee-ab3a385b57d3.png)

BjossiAlfreds:
[...]

PR #698, opened 2021-04-08
Map fixes for old and newly discovered bugs + removed code hacks

BjossiAlfreds:
This PR fixes the following bugs: https://github.com/yquake2/yquake2/issues/689 https://github.com/yquake2/yquake2/issues/687

In addition I implemented the code hacks as map fixes (city2, city3, cool1, lab), as well as fixing some additional bugs reported within the bug reports (city1). There are other more minor fixes, like fixing "duration is not a field" console warning in biggun and using an orphaned target_explosion entity in base2.

I removed the code hacks for the following reasons:
1. They make correct configuration more complicated and easy to mess up.
2. They are incomplete. A lot of the bugs found were not addressed by them and the hack for map "lab" was not even fully working.
3. Only the absolutely most die-hard Q2 fans will notice any difference, and even then I doubt they will mind having to kill an extra enemy or two in a couple of levels.

I have a question for @Yamagi, how would you like me to document the map fixes in the YQ2 docs?

### DIFF
--- a/doc/020_installation.md
+++ b/doc/020_installation.md
@@ -134,6 +134,45 @@ some global settings to saner defaults. To use it copy *stuff/yq2.cfg*
 into the *baseq2/* directory.
 
 
+### Fixed Map Data Files
+
+As an optional feature, Yamagi Quake II provides fixed map data files
+(.ent files). These files provide a replacement entity list in order
+to fix some map bugs that have been discovered by players over the
+years.
+
+These fixes include fixes for bad  monster counts in some
+maps/difficulty settings, fixing broken spawn chains that made it
+impossible to spawn some monsters, removing DM-only items that spawn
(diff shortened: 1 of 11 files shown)


Yamagi:
> I have a question for @Yamagi, how would you like me to document the map fixes in the YQ2 docs?

I would mention the entity files in the installation guide under https://github.com/yquake2/yquake2/blob/master/doc/020_installation.md if you want to document the map fixes itself (what was changed and for what reason) we'll need a new section in the docs. I don't think that such a detailed documentation is necessary, the entity files itself and the comments you added to them should be enough.

BjossiAlfreds:
I've written some docs now. Let me know if I made any errors or you don't like its location within the file.

Yamagi:
Nope, it's perfect. I'm merging this now. Thank you for looking into it.

PR #699, opened 2021-04-09
network: port being unsigned being w/o named service when above the p…

devnexen:
…ort limit


DanielGibson:
I strongly vote for instead reverting your other commit and making `port` an `int` again.
1. Who knows what else is subtly broken by that change
2. Consistency with Windows backend

Yamagi:
Yeah, I shouldn't have merged that. Let's revert it.

devnexen:
ah right the windows backend

PR #705, opened 2021-04-18
Bind keys using only a controller

madmodder123:
Let's you remap keys without a keyboard, useful for a Raspberry Pi setup.
Also let's you clean bindings.

Tested on:
    Device: Raspberry Pi 4 - 4GB
    OS: Raspbian 10 (buster) on kernel 5.10.17-v7l+
    RetroPie Version: 4.7.11

### DIFF
--- a/src/client/menu/menu.c
+++ b/src/client/menu/menu.c
@@ -999,11 +999,15 @@ Keys_MenuKey(int key)
     {
     case K_KP_ENTER:
     case K_ENTER:
+    case K_JOY1:
+    case K_JOY3:
         KeyBindingFunc(item);
         return menu_in_sound;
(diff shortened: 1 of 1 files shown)


Yamagi:
I don't think that the buttons should be hardcoded. The obvious problem is, that you can never be sure were these buttons are located on the connected controller or if they're available at all. Our controller support is rather rudimentary, mostly due to Quake II not really suited for controller input. The minimum requirement would be something like the 'backbutton hack': https://github.com/yquake2/yquake2/blob/master/src/client/input/sdl.c#L1438 Of course it would be even better to switch to bindings through `SDL_GameControllerGetBindForButton()` everywhere...

Issue #704, opened 2021-04-18
[Original game bug] Dead sharks have messed up bbox and fall out of world


BjossiAlfreds:
When the shark finishes its death animation, its bbox gets messed up (almost completely outside its visible body) and it falls out of the world if too close to a ground surface.

I'll provide PRs that fix this and explain in more detail there what is going on.

Yamagi:
I've merged all the pull requests.

PR #706, opened 2021-04-18
Shark bbox fix and added check for inuse after entity thinking

BjossiAlfreds:
This PR fixes bug https://github.com/yquake2/yquake2/issues/704

flipper_dead() sets the dead bbox top and bottom planes incorrectly. This is likely a copy/paste error by id. Unlike most monsters, the flipper's point of origin is at the bottom of its model, so mins[2] is 0, however the death code sets mins to -24 and maxs to -8, not only moving the bbox but actually leaving the model completely.

This makes it hard for the player to know where to actually shoot to gib the dead body, and because the bbox moves, it means it might clip into the world and this causes the entity to freefall forever.

I fixed this by keeping maxs[2] unchanged and doing a trace downwards by 8 units, so mins[2] will be anywhere from 0 to -8 based on if/when the trace hits an obstacle.

I also snuck in a small inuse check after SV_RunThink() in SV_Physics_Toss(). Entities are very often freed while thinking so this both prevents wasteful physics calculations and reduces the risk of writing data to a freed entity.

### DIFF
--- a/src/game/g_phys.c
+++ b/src/game/g_phys.c
@@ -924,6 +924,12 @@ SV_Physics_Toss(edict_t *ent)
 	/* regular thinking */
 	SV_RunThink(ent);
 
+	/* entities are very often freed during thinking */
+	if (!ent->inuse)
+	{
+		return;
+	}
+
 	/* if not a team captain, so movement
 	   will be handled elsewhere */
 	if (ent->flags & FL_TEAMSLAVE)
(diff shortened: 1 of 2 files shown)


maraakate:
Nice work, the ent->inuse thing is a good change too.  Easily missed by many things in the code.

Yamagi:
Thank you!

PR #707, opened 2021-04-30
Fix for some player sound bugs

BjossiAlfreds:
This PR fixes some bugs with player sounds.

* Player does pain sound in lava with godmode on

A small oversight by id. They check for the invul powerup but not the godmode flag. I just added the godmode flag check.

* Player gib sound is rarely replaced by a footstep

This is a side-effect of change to sound priority ordering. Footstep events schedule the sound after the server frame has ends so it would override the gib sound. This was fixed by adding a check for player's health, and also by changing the channel the gib sound is played on (see later).

* Player gib sound is played on the wrong channel

This is a bug in the original game and causes player death scream and gib sound to play over each other if the player is killed by a powerful shotgun blast or is gibbed very quickly after dying. It was playing on the CHAN_BODY channel but should have been CHAN_VOICE. This change is also consistent with monsters, which play their gibs on CHAN_VOICE.

* Incorrect death sound plays some times

Another side-effect of the change to sound priority is that pain sound could override the death sound. This would happen if for example a shotgun blast first hurts the player and then kills him, all in a single server frame. Was easily fixed by adding a check for player health for the pain sound code.

I also prevented the death scream from overriding the gib sound which could happen after  I made both play on the same channel. Fixed by using the *sounds* variable to aggregate the sound effect, and then play it at the end of the server frame in P_DamageFeedback.

### DIFF
--- a/src/game/header/local.h
+++ b/src/game/header/local.h
@@ -1048,7 +1048,7 @@ struct edict_s
 	int dmg;
 	int radius_dmg;
 	float dmg_radius;
-	int sounds; /* make this a spawntemp var? */
+	int sounds; /* now also used for player death sound aggregation */
 	int count;
 
 	edict_t *chain;
--- a/src/game/player/client.c
+++ b/src/game/player/client.c
@@ -908,9 +908,8 @@ player_die(edict_t *self, edict_t *inflictor, edict_t *attacker,
 
 	if (self->health < -40)
 	{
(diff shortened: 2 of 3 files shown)


Yamagi:
And again: Thank you for fixing all these corner cases! I've merged this one and the addon version. Ill publish a new Windows test build with all the changes from the past weeks later today.

PR #709, opened 2021-05-01
cold attribute introduction proposal, applying for Sys_Error at first.

devnexen:


DanielGibson:
Does this have any measurable performance impact?

devnexen:
that is difficult to benchmark depends on the compiler and version, prioritising hot code paths and performance outcomes might differ case to case.

Yamagi:
I don't think these changes are worth the (potential) hassle. Quake II runs more than fast enough on all systems build in the last 15 years, the only exception might be cheap some ARM boards. And those are more GPU than CPU limited. Introducing compiler (family) specific attributes to the code might give some small speedups that no one really needs, at the cost of much worse portability to other compiler families like MSVC. And it diverges us even more from other source ports.

devnexen:
fair enough you re right

PR #717, opened 2021-06-01
Add ogg/vorbis sound for effects

0lvin:


0lvin:
Code is checked with convert all mp3 files to ogg/vorbis in Hardly Workin' from Machinima.

[yquake2: Hardly Workin' - Machinima](http://www.youtube.com/watch?v=mMf1ESxiROM)

src/client/sound/ogg.c
```diff
@@ -709,3 +710,62 @@ OGG_Shutdown(void)
...
+		/* fill in wav structure */
+		info->rate = ogg_file->sample_rate;
+		info->width = 2;
+		info->channels = ogg_file->channels;
+		info->loopstart = -1;
+		info->samples = stb_vorbis_stream_length_in_samples(ogg_file);
+		info->dataofs = 0;
+
+		/* alloc memory for uncompressed wav */
+		final_buffer = Z_Malloc(info->samples * sizeof(short));
```
DanielGibson, line 744: shouldn't this also take the number of channels into account?

src/client/sound/sound.c
```diff
@@ -205,17 +205,63 @@ S_IsSilencedMuzzleFlash(const wavinfo_t* info, const void* raw_data, const char*
...
+	/* Remove the extension */
+	memset(namewe, 0, 256);
+	memcpy(namewe, path, len - (strlen(ext) + 1));
+
+	if (len < 5)
+	{
+		return;
+	}
+
+	if (strcmp(ext, "wav"))
```
DanielGibson, line 239: what's happening here?

src/client/sound/sound.c
```diff
@@ -205,17 +205,63 @@ S_IsSilencedMuzzleFlash(const wavinfo_t* info, const void* raw_data, const char*
...
+		/* Non wav? */
+		return;
+	}
+
+	/* Combine with ogg */
+	Q_strlcpy(filename, namewe, sizeof(filename));
+
+	/* Add the extension */
+	Q_strlcat(filename, ".", sizeof(filename));
+	Q_strlcat(filename, "ogg", sizeof(filename));
```
DanielGibson, line 250: it doesn't really matter, but why not just do it in one `Q_strlcat()` call with ".ogg"?

src/client/sound/ogg.c
```diff
@@ -709,3 +710,62 @@ OGG_Shutdown(void)
 
 	ogg_started = false;
 }
+
+void
+OGG_LoadAsWav(char *filename, wavinfo_t *info, void **buffer)
```
DanielGibson, line 715: I know lots of code in q2 does this, but maybe we should finally start using `const char*` for strings, at least in new code

src/client/sound/ogg.c
0lvin: I have changed it in 3cc39bb48e3b42847e6e5f6024e7c460f83ce114, looks like stb_vorbis_stream_length_in_samples has returned samples multiplied by channels.

src/client/sound/sound.c
0lvin: It tried to check that originally file was wav, i have removed such check in 3cc39bb48e3b42847e6e5f6024e7c460f83ce114 as redundant.

src/client/sound/sound.c
0lvin: fixed

src/client/sound/ogg.c
0lvin: Filename has used in `FS_LoadFile` and `FS_LoadFile` has used `char*`, i have updated chars to const where it not required to add `const char*` -> `char*` before use.

Yamagi:
@0lvin, @DanielGibson: Are we happy with this?

0lvin:
@Yamagi Nothing to add from me, I have added little more code for flexible haptic feedback in #722 . I will be appreciate if someone will leave feedback how it "sounds", Code in #722 has exactly same duration and volume of haptic feedback as audio files and tried to detect attack and fade duration.

Yamagi:
@DanielGibson  Are you okay with this? I want to merge this PR before continuing with #722.

DanielGibson:
looks good!

Yamagi:
Thank you!

PR #718, opened 2021-06-02
Fix some leaks

Spirrwell:
Most of these are probably "harmless", but I could definitely see the missing/improper SDL_Quit/SDL_QuitSubsystem causing headaches later.

EDIT: I thought this was pretty small, but I could break down each leak fix into individual commits if that's preferable and redo the pull request.

EDIT 2: I also do hope this doesn't fall into the category of "unnecessary cleanups" per contributing guidelines. This isn't really a refactor which is what I think that's referring to. But if that's not the case, I'm sorry I could just fix the ones introduced specifically by yquake2.

### DIFF
--- a/src/client/cl_keyboard.c
+++ b/src/client/cl_keyboard.c
@@ -1129,6 +1129,20 @@ Key_Init(void)
 	Cmd_AddCommand("bindlist", Key_Bindlist_f);
 }
 
+void
+Key_Shutdown(void)
+{
+	int i;
+	for (i = 0; i < K_LAST; ++i)
+	{
+		if (keybindings[i])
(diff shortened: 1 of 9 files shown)


maraakate:
I like it.  Any of these that get knocked out, no matter how benign, reduces noise from leak detector tools.

Yamagi:
Yeah these are looking good. I'll merge it after 8.00 is done. While I don't see any problems, there's always a small possibility of some corner cases getting broken. We'll do a release candidate next week and if nothing comes up the actual release a week later.

Yamagi:
As promises: Merged. Thank you!

PR #721, opened 2021-06-13
solairs/illumos build fix proposal.

devnexen:


devnexen:
dunno if tool late for this release :)

DanielGibson:
looks harmless enough IMO

Yamagi:
In that case: This is the last addition to 8.0.

PR #727, opened 2021-07-03
build fix proposal on mac where archs are well defined.

devnexen:
the arch normalization leads to arm32 bit build under mac m1.


Yamagi:
And merged. Thank you :)

PR #734, opened 2021-08-09
Preserve sounds/maps and images on map change

0lvin:
* use SmoothColorImage as sw_anisotropic filtering replacement
* port preserve image/model code from vulkan render.


Yamagi:
Thank you and keep up the good work!

PR #742, opened 2021-09-20
Update for benchmark mode

0lvin:
Code mostly based on ideas from: [phoronix-test-suite test profile](https://github.com/phoronix-test-suite/test-profiles/pull/202).
* gl1, gl3: add scale 8bit textures 
* gl1, gl3: try to set lower msaa level on error(test profile can set unsupported level)
* gl1,gl3: preserve images and models on level change if list of image has space.
* gl1,gl3: use local currententity and currentmodel. I hope that it will help with debug. I still see sometime error that model has incorrect type.
* gl1,gl3,soft: reuse common code for get model radius and add to PVS code const modifier.


Yamagi:
I'll have a look at the coming weekend.

Yamagi:
So, finally: From a source code point of view I'm okay with this. But what about renaming `r_retexturing 2` to something like `r_scale8bittextures` to make clear that it's not a texture replacement and just scales the original textures.

If it's for benchmarking, wouldn't it be better to remove the 1000 FPS lock for `timedemo 1`? I'm not 100% sure if its possible, though. As far as I remember the timing code timedemos don't have timings and shouldn't suffer from the '1 millisecond is the lowest time unit' problem.

0lvin:
@Yamagi  Could you look? 

I have added additional `r_scale8bittextures` flag for scale. 

As addition timedemo is limited to 5 x vid_maxfps. I have checked timedemo with my weak gpu in 240x320 and received 2320 fps, and I expect that top level gpu will provide 5k fps.

Yamagi:
And merged! Thank you!

0lvin:
thank you, I ported 1e0d75857ba64b60d6a68d6e78fd542a6f429cd1 change to ref_vk https://github.com/yquake2/ref_vk/commit/049f26a212a889867c66a52af0cf7e45f7cded42

DanielGibson:
while at it, you could also port https://github.com/yquake2/yquake2/commit/3873c76e125b3b4344c158e7397fe27bff77efbe ;)

0lvin:
@DanielGibson Done :-) https://github.com/yquake2/ref_vk/commit/0ee36fc8016a2496614519e0b70cc4844047024b

DanielGibson:
Sorry, I only noticed this now: "limit timedemo to vid_maxfps * 5" - why?
Isn't the whole point of timedemo to ignore framerate limitations to see how fast the game can render?

0lvin:
No real logic hides under it, just limit fps to some reasonable value and 5000 fps should be enough for everyone and save little bit power :-) I suppose that initial limit of frametime is just for save power. 

As i see nobody came near to it by https://openbenchmarking.org/test/pts/yquake2, when someone will be near to it - i would prefer to send new patch with new limit.

DanielGibson:
AFAIK `timedemo 1` is only used for benchmarking, not for normal playing, so power consumption doesn't matter - or am I missing another usecase?

And you're only able to get 5000fps if you set vid_maxfps to 1000 - the default is 300, and it's totally reasonable to set it to an even lower value (for example if vsync causes trouble but you want to limit frames to save power and keep your system more quiet).

0lvin:
I have tailored it for openbenchmarking and have used `+exec pts.cfg +set vid_fullscreen 0 +set r_retexturing 2 +set r_scale8bittextures 1 +set vid_maxfps 999` as common part of command line. ( https://openbenchmarking.org/innhold/b247efe97b6bf5340359e4aff1f34cf565aa1493 )

I was afraid that changes of limits of max_fps and minimal frame time would create some regressions. 

Maybe best solution will be remove multiply to 5 and just make maxfps as 9999.

DanielGibson:
IIRC allowing maxfps > 1000 can fuck up things because then one frame can take < 1ms. But maybe that's only relevant for `cl_async 0`? Not sure.

However, frames <1ms only cause problems when actually playing, not when running demos.
So I'd argue that maxfps should remain as it is and there should be no restrictions at all with `timedemo 1`.

0lvin:
@DanielGibson I have created fast fix in #823

DanielGibson:
thank you! :)

PR #744, opened 2021-09-21
Consistent naming

apartfromtime:
Keep names consistent with OGG options.


Yamagi:
That's a good suggestion. Thank you!

Issue #745, opened 2021-09-25
vSync limits the fps to 80fps on macOS


warrex1979:
I was just playing the 8.00 release on macOS Big Sur 11.6 (Mac Mini 2020 with M1 SoC connceted to a Sony KD-48A9 OLED TV) and noticed that even when vid_maxfps and cl_maxfps are both set to 120 vSync limits the fps to 80 (OpenGL 1.4 and 3.2). Turning it off results in an absolutely stable 120fps@1080p. I did not encounter this behaviour with the OpenGL backend of GZDoom.

Btw.: Thank you for compiling builds for macOS! It would be nice if you could cross compile for Apple Silicon though.

Yamagi:
The MacOS is maintained by the community. Neither @DanielGibson not me have a mac, so we're unable to help here. If someone is able to figure out whats wrong and sends a pull request with a fix we're happy to merge it.

devnexen:
might depend on the gzdoom settings tough, I have same fps limitations on my mac on both gzdoom and yquake2 on my mac.

For the cross compiling part, theoretically a separate build for arm64 then for x86_64 (rosetta) on a mac m1, then "gluing" them together with lipo to make an universal binary would be ideal.

warrex1979:
> might depend on the gzdoom settings tough, I have same fps limitations on my mac on both gzdoom and yquake2.

At least in GZDoom 4.7.0 I cannot reproduce this. I tested OpenGL, OpenGL ES and Vulkan with VSync on. All of them are locked to 120fps if I set the target fps respectively in the menu.

Thank you for considering cross-compiling (https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary).

Please make sure to sign and notarize the release then (the current release of QuakeSpasm crashes on M1 Macs because the team did not follow Apple's guides):

"New in macOS 11 on Macs with Apple silicon, and starting in macOS Big Sur 11 beta 6, the operating system enforces that any executable must be signed before it’s allowed to run."
https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11_0_1-universal-apps-release-notes

Code Signing hub page on Apple.com:
https://developer.apple.com/support/code-signing/

Notarizing macOS Software Before Distribution: https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution

devnexen:
[...]

PR #749, opened 2021-10-10
VSync on macOs build fix proposal.

devnexen:
using native (but deprecated) OpenGL api instead of SDL on this
platform.


devnexen:
did not refactor purposely gl3/gl1 there might be a reason I ignore.

DanielGibson:
Are we sure that this fixes #745 ?

devnexen:
I tried on an external TV but not a 120Hz, was hoping one of the folks testing it to confirm.

devnexen:
seems not working out regarding the SDL2 topic.

Issue #757, opened 2021-10-27
[Bug] Maps with special chars in cmd line


NightFright2k19:
Yamagi 8.0 apparently cannot properly load maps with special chars in it during startup via command line (not ingame).

Example: Scorched Earth
Map name (e.g.): SGC9-1.bsp

When trying to load the mod with 
**yquake2.exe +set game scorched +set skill 0 +map sgc9-1**
you get an error ("Can't find maps/sgc9.bsp"). This means the "-1" part of the map name is not parsed. Same with "-2", "-3" etc. It works if you rename the file, e.g. to sgc9_1.bsp.

Maps load without issue when launched via ingame console (map sgc9-1).

At first I thought this was an issue with my launcher (Simple Quake Launcher 2.5), but [the SQL author checked it](https://github.com/m-x-d/Simple-Quake-Launcher-2/issues/9) and says it must be port-related.

maraakate:
If I remember correctly, Cbuf_AddLateCommands is for some bologna with setting CVARs at the right time during Qcommon_Init() which may not be initialized until after the CVAR_Init(), Server_Init(), etc. friends are done.  So it basically adds it twice.  However, it's been a long time since I've goofed with it and may be wrong.

DanielGibson:
I just pushed a minimal fix for this

NightFright2k19:
I can confirm this works now.

NightFright2k19:
It seems Yamagi still cannot load some maps with special chars, at least not when they are at the end of the map name. There is a DM map named _wtf!.bsp_ and only loads if you put something after the exclamation mark, e.g. _wtf!**1**.bsp_.

NightFright2k19:
Probably a mistake from my side. I had these maps in the mapcycle list of the Eraser bot (maps.txt) and when they wouldn't load, I assumed it was port-related. Anyway, Eraser game code is from 1998 and I guess it has something to do with that.

PR #758, opened 2021-10-29
Fixed incorrect handling of - in late args and simplified code

BjossiAlfreds:
This PR addresses issue https://github.com/yquake2/yquake2/issues/757. It also greatly optimizes and shortens the code by skipping all the unnecessary string copying.

### DIFF
--- a/src/common/cmdparser.c
+++ b/src/common/cmdparser.c
@@ -298,49 +298,63 @@ Cbuf_AddEarlyCommands(qboolean clear)
 qboolean
 Cbuf_AddLateCommands(void)
 {
+	int i, j;
+	int s;
+	char *text, c;
+	int argc;
 	qboolean has_args = false;
-	char *arg;
-	int i, argc = COM_Argc();
+
+	/* build the combined string to parse from */
+	s = 0;
+	argc = COM_Argc();
 
 	for (i = 1; i < argc; i++)
 	{
-		arg = COM_Argv(i);
+		s += strlen(COM_Argv(i)) + 1;
+	}
 
-		if (*arg != '+')
+	if (!s)
+	{
+		return false;
+	}
+
+	text = Z_Malloc(s + 1);
+	text[0] = 0;
+
+	for (i = 1; i < argc; i++)
+	{
+		strcat(text, COM_Argv(i));
+
+		if (i != argc - 1)
 		{
-			continue;
-		}
-
-		has_args = true;
-		i++;
-
-		Cbuf_AddText(arg + 1);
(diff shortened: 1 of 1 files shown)


DanielGibson:
Yamagi and me discussed this a bit, and there's still this issue I mentioned in #757:
> Are there legit usecases for - or at least + within an argv string? Like yquake2.exe +set game rogue+map whatever (without a space before +map that would turn rogue and +map into separate argvs)?

And while "legit" of course is always debatable, we fear that there's indeed .bat scripts from old mods and things like that which have a command without a space before the `+`.
While my example (`+set game rogue+map whatever`) would never have worked, because `Cbuf_AddEarlyCommands()` clears all `argv`s of `+set` and its arguments, `+map whatever+set r_fullscreen 0` would have worked before but not with this change.
So while the way you're doing it now is better code (and definitely is what id should have done in the first place - after all they basically do the same in `Cbuf_AddEarlyCommands()`), probably it would be better to keep the code like it was before (i.e. do that concatenation of the `argv`s) and only add special handling to the "skip `-`" case that makes sure to ignore `-` if the previous character is not a space

DanielGibson:
I just pushed a minimal invasive fix - if you want you can try to rework your (cleaner) code to also support this case, but maybe the result would be even uglier than the current state

BjossiAlfreds:
A lot of really weird stuff is possible with this game's strings if you try hard enough. The question is what can be reasonably expected to happen in practice, and I'm not sure if I share your concerns. If you can find 2-3 concrete examples of this then I'll stand corrected, but from where I stand I see no incentive for people to think that not separating arguments would even work, let alone do it for any other reason than typo.

But it's up to you guys in the end. I do think id's code can be simplified still, doing just one string build instead of two. I'd be happy to take on the challenge and update this PR.

BjossiAlfreds:
My algorithm is pretty much completely unreworkable around having to build a string, so I just used your commit as a new base. I got rid of the build string and just called Cbuf_AddText directly for each scanned command. Not a big change or anything but then at least we get cleaner code, backwards compatibility and the bugfix all in one. Give this code a review and see if you can make it break (I couldn't).

And just to add to what I said above, you can't even chain early args together or it will break, so why someone would add a space between +set's but then decide to chain the late args (and that is without even knowing about early vs late args)... I don't even want to try to get into that frame of mind haha.

Yamagi:
So, do we want this or not? :) To me the code looks good and so far I wasn't able to break it.

Yamagi:
Mkay, we want it. Thank you! :)

Issue #759, opened 2021-10-29
[Bug] Showstopper crash in Oblivion


NightFright2k19:
When playing Oblivion, you will eventually experience a crash to console in "The Harvester" map, making it impossible to continue the game.

Crash messages (one or the other):
**ERROR: Game error: ED_Alloc: no free edits
ERROR: *Index overflow**

Port: Yamagi v8.0
Mod: [Oblivion (with 2012 patch) from ModdB](https://www.moddb.com/mods/turoktirol/downloads/oblivion1)
Map: "The Harvester" (harvester1.bsp)

Reproduction:
- The "index overflow" error occurred while playing through the campaign and making progress on that map. Eventually I got kicked to console, either when trying to shoot a mob or by just running away. Curiously enough, I managed to get past the spot which triggered the second error described below.
- The "no free edicts" error occurred when trying to start the map via launcher (SQL 2.5) and blowing up the barrels on the upper deck (after climbing the ladder near the entrance to the room with imprisoned marines behind a forcefield).

NightFright2k19:
Outstanding, my friend. Much appreciated!

BjossiAlfreds:
If Oblivion has source code available, you could replicate my changes there. Otherwise your only option is another port I'm afraid.

NightFright2k19:
I need to check if there is source available for Oblivion. Mods released back in the days sadly didn't have that so often. 

But just updating the executable wouldn't fix the crashes? Oblivion must have had increased limits in its custom exe back then, otherwise it would have crashed already back then.

BjossiAlfreds:
Entity allocation is done in the game library, not the main client. I guess Oblivion made no changes to the allocation algorithm, just upped the entity limit itself, which would require quite a bit of changes in both the game library and the client.

There is always making your own ent files for these maps though, with non-essential entities removed, like armor shards and stimpacks. Though you'd have to play again from the hub or the beginning for the ent files to take effect. But I'd suggest just using another port and avoid all that extra work.

NightFright2k19:
Well, already tried the unofficial Q2 v3.24 patch and KMQuake2, both crash as well. However, since the dll remains the same, I wonder if it works anywhere at all.

maraakate:
[...]

PR #760, opened 2021-10-29
Made the game able to handle entity overload better

BjossiAlfreds:
This PR amortizes the issues reported in https://github.com/yquake2/yquake2/issues/759 by easing entity allocation policies.

These are the main points:
* If no free entity is found according to the default policy, a second search is done, ignoring any freetime restrictions
* New function G_SpawnOptional, instead of calling gi.error() it will return NULL, allowing the caller to react more gracefully
* Made ThrowDebris, ThrowGib and AI tempgoals use G_SpawnOptional

### DIFF
--- a/src/game/g_ai.c
+++ b/src/game/g_ai.c
@@ -1196,8 +1196,15 @@ ai_run(edict_t *self, float dist)
 		return;
 	}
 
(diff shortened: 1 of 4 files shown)


NightFright2k19:
Thanks for doing this, guys!

Issue #763, opened 2021-11-05
[Original bug] player_noise entity leak


BjossiAlfreds:
I discovered this one pretty recently. Any time you go through a level transition, players lose their link to their player_noise entities, so PlayerNoise() will spawn new ones.

Steps to reproduce:
1. Open rogue (we need their "entcount" command)
2. Open ware2 and execute entcount
3. Fire the blaster
4. Execute entcount again and notice 2 entities were spawned
5. Go back to ware1 through the bulkhead
6. Return to ware2
7. Fire the blaster again
8. Execute entcount
9. Notice that 2 more entities have been spawned
10. You can keep repeating this process all the way until the game crashes to console

This is not a big issue in the moderately sized levels/units of the base campaign but this could be a showstopper in very entity heavy maps with lots of backtracking potential.

I have fixed this on my end by verifying mynoise and mynoise2 links when PlayerNoise() is called. If one or both of them are bad it will search the level entity list for matching player_noise and link them back to the player, or spawn new ones if one or both of them are missing. If you want I can replicate that code for YQ2 and do a PR.

Also, would you like player_noise to be optional entities? It would merely cause monsters to not hear player sounds any more, if the entity fails to spawn.

Yamagi:
> I discovered this one pretty recently. Any time you go through a level transition, players lose their link to their player_noise entities, so PlayerNoise() will spawn new ones.
> [..]
> This is not a big issue in the moderately sized levels/units of the base campaign but this could be a showstopper in very entity heavy maps with lots of backtracking potential.

I'm always impressed what interesting problems can still be found after so many years. That's very impressive, myself and most other users would never have noticed. A PR would be appreciated. And making player_noise entities optional so that they can be omitted in case of an entity overflow sound like a good idea too.

BjossiAlfreds:
A random thought led me to discovering this. I just randomly asked myself if those mynoise links are always maintained, and I found it likely level transitions would break them since player edicts are reset in those situations. Then I tested and confirmed that this is a leaky situation.

Yamagi:
I've merged the pull requests.

PR #764, opened 2021-11-05
Fixed player_noise entity leak and improved stability surrounding player noises

BjossiAlfreds:
This PR addresses issue https://github.com/yquake2/yquake2/issues/763 and also improves code stability.

* Added a safety check to make sure the entity making the noise is actually a player
* Made spawning of player_noise entities optional. If they fail to spawn, PlayerNoise() will not cause any sound event
* Each time PlayerNoise() is called it will verify the mynoise variables of the player, and make an effort to repair them if needed
* mynoise and mynoise2 can now be differentiated by spawnflags. self/weapon noises have none, while impact noises have spawnflag 1 set. This will help verification to re-link the noises correctly to the player in cases where impact noise appears first in the entity list. This is not crucial however, if the linking is incorrect it will correct itself quickly during gameplay and players are unlikely to notice any AI derps because of it.

### DIFF
--- a/src/game/player/weapon.c
+++ b/src/game/player/weapon.c
@@ -27,6 +27,9 @@
 #include "../header/local.h"
 #include "../monster/misc/player.h"
 
+#define PLAYER_NOISE_SELF 0
+#define PLAYER_NOISE_IMPACT 1
+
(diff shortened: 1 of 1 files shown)


Yamagi:
As always: Thank you and keep up the good work!

Issue #762, opened 2021-11-05
stb_image.h and stb_image_resize.h updates break renderer libs on windows


j4reporting:
It seems that commit 67f22cce17c059dd76deb6e62bfbc2a77085a3f1   and  5bae9145f42dc336b0334be5f9cd001b9f32cc1e  ( ref_vk)   break all renderer libs  on Windows. 
System runs WIn10 21H1   with Intel GPU and driver version 27.20.100.9466 installed.

```
stderr:    Error: No usable renderer found!

stdout:
----- refresher initialization -----
Loading library: ref_vk.dll
Sys_LoadLibrary failed: LoadLibrary returned 126 on e:\games\yquake2_pre_own_msys_builds\ref_vk.dll
Loading ref_vk.dll as renderer lib failed!
Retrying with gl3...
----- refresher initialization -----
Loading library: ref_gl3.dll
Sys_LoadLibrary failed: LoadLibrary returned 126 on e:\games\yquake2_pre_own_msys_builds\ref_gl3.dll
Loading ref_gl3.dll as renderer lib failed!
Retrying with gl1...
----- refresher initialization -----
Loading library: ref_gl1.dll
Sys_LoadLibrary failed: LoadLibrary returned 126 on e:\games\yquake2_pre_own_msys_builds\ref_gl1.dll
Loading ref_gl1.dll as renderer lib failed!
Retrying with soft...
----- refresher initialization -----
Loading library: ref_soft.dll
Sys_LoadLibrary failed: LoadLibrary returned 126 on e:\games\yquake2_pre_own_msys_builds\ref_soft.dll
Loading ref_soft.dll as renderer lib failed!
Shutting down OpenAL.
```

buildenv is an up to date  MSYS2 setup  with  gcc.exe (Rev1, Built by MSYS2 project) 11.2.0

Yamagi:
Ha, nice. I wanted to make another Windows testbuild tonight with this weeks fixes... Looks like this was broken by the thread local stuff added to stb_image.h about two years ago: https://github.com/nothings/stb/commit/eb48fbdcede98f93dab3a5e42ce5b0072b739cf1 Maybe the mingw stuff doesn't implement enough threading for that. Or something like that. I'll need to dig deeper to figure out what exactly is wrong. I'll commit a band aid fix shortly.

Yamagi:
[...]

PR #765, opened 2021-11-05
Re-enable thread-local stuff in stb_image, link winpthread statically

DanielGibson:
The real problem wasn't that thread-local variables are unsupported by
MinGW-w64, but that using them makes it link against libwinpthread-1.dll

Alternatively that lib can be linked statically, like we do now, at least
for the rendering backends that use stb_image.h/stb.c

For some reason it didn't work to add
`-Wl,-Bstatic -lpthread -Wl,-Bdynamic` in LDLIBS - it only worked without
the `-Wl,-Bdynamic` part, so it must be last on the linking commandline
(otherwise it'll try to link SDL2 statically as well).

refs #762

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1077,7 +1077,7 @@ endif
 ifeq ($(YQ2_OSTYPE), Windows)
 release/ref_gl1.dll : $(REFGL1_OBJS)
 	@echo "===> LD $@"
-	${Q}$(CC) $(LDFLAGS) $(REFGL1_OBJS) $(LDLIBS) $(DLL_SDLLDFLAGS) -o $@
+	${Q}$(CC) $(LDFLAGS) $(REFGL1_OBJS) $(LDLIBS) $(DLL_SDLLDFLAGS) -Wl,-Bstatic -lpthread -o $@
 	$(Q)strip $@
 else ifeq ($(YQ2_OSTYPE), Darwin)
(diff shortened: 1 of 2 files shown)


DanielGibson:
TBH no idea how much sense this makes, it's not like we're using threads, so stb_image doesn't really need thread-local variables anyway..
I mostly did this because I couldn't believe that mingw-w64 doesn't support thread-local variables - and as it turned out, it does, you only need libwinpthread for it.

Yamagi:
Under Linux and likely all other platforms besides Windows this problem doesn't show up, because `sdl2-config --libs` adds `lpthread` for the client and all renderer libs. Linking winpthreads under Windows would bring into onto par to the other platforms.  The question is, if linking winpthreads and spawning threads inside the renderer DLL can trigger some unwanted side effects. I don't think so, because Windows has this strange process model with very limited stacks anyways and the scary `longjmp()` is guaranteed to jump inside the main thread.

j4reporting:
> For some reason it didn't work to add `-Wl,-Bstatic -lpthread -Wl,-Bdynamic` in LDLIBS

I know this is old, but in case someone needs this.  MINGW needs  that enclosed with  `--whole-archive`  .....  `--no-whole-archive` 

e.g.   ` -Wl,-Bstatic,--whole-archive -lwinpthread -Wl,-Bdynamic,--no-whole-archive`

PR #770, opened 2021-11-18
Multiplayer\Coop menu key binds

apartfromtime:


apartfromtime:
I'm not sure why git merged the player_noise entity leak commit along with my changes.

Yamagi:
I've merged 0f80b29 by hand to avoid the garbage commits. Thank you for the contribution! :)

PR #771, opened 2021-11-28
Change the amount of listed renderers to 3 in documentation

VespuCore:
This is a tiny change.
Since **Vulkan** was removed from the list as a renderer,  there are only 3 renderers listed.


Yamagi:
Thank you!

Issue #773, opened 2021-12-08
[Bug] Map "Pumping Station 2" (waste3) "target_secret" missing (fix)


KPThug:
Solution:
In g_spawn.c -> "SpawnEntities" function

After:
if (!Q_stricmp(level.mapname, "command") && !Q_stricmp(ent->classname, "trigger_once") && !Q_stricmp(ent->model, "*27"))
   ent->spawnflags &= ~SPAWNFLAG_NOT_HARD;

Add:
if (!Q_stricmp(level.mapname, "waste3") && !Q_stricmp(ent->classname, "func_door") && !Q_stricmp(ent->model, "*12"))
   ent->target = "t117";

PD: No idea how to add it to the .ent files.

Regards, Gustavo.

BjossiAlfreds:
This was already fixed in the baseq2 dll, in g_func.c. Though of course an ent file is always better than a code hack, as it would work independent of the game dll being used.

BjossiAlfreds:
@Yamagi If you want I can commit a waste3.ent that addresses this and bad player start from console, and remove the code hacks.

I also spotted 2 gunners that have the spawnflags 1024 (not hard/nightmare). I feel like this is a bug and was intended to be 768 (not easy + not medium). I haven't change this but thought I'd mention it. Thoughts?

Yamagi:
> If you want I can commit a waste3.ent that addresses this and bad player start from console, and remove the code hacks.

That would be nice.

> I also spotted 2 gunners that have the spawnflags 1024 (not hard/nightmare). I feel like this is a bug and was intended to be 768 (not easy + not medium). I haven't change this but thought I'd mention it. Thoughts?

You could fix this in the entity files. Player who use the files will get the gunners on easy and medium. Purists playing without the entity files will keep the Vanilla behavior.

Yamagi:
I've merged the pull request.

PR #774, opened 2021-12-11
Moved code hacks for waste3 to waste3.ent

BjossiAlfreds:
This PR addresses https://github.com/yquake2/yquake2/issues/773 and removes the 2 waste3-related map hacks from the code as they are made redundant by the fixed ent file.

In addition to addressing those issues, the enr file also:
1. Fixed wrongly placed message for secret, the message is supposed to be in the target_secret entity but was in the trigger, causing a subtle sound glitch.
2. Fixed 2x monster_gunner with spawnflag 1024. The level designer most likely thought for a moment spawnflags were inclusive rather than exclusive, so viewed 1024 as hard-only rather than 768.
3. Added difficulty spawnflags to entities specific to monsters such as point_combat
4. Added not-deathmatch spawnflag (2048) to entities that are never used/accessible in DM
5. Start missing pumping sound effect for big pump mechanism in DM.

All changes are documented and trace-able by viewing the ent file with a text editor.

### DIFF
--- a/src/game/g_func.c
+++ b/src/game/g_func.c
@@ -1841,13 +1841,6 @@ SP_func_door(edict_t *ent)
 	{
 		ent->think = Think_SpawnDoorTrigger;
 	}
-
-	/* Map quirk for waste3 (to make that secret armor behind
-	 * the secret wall - this func_door - count, #182) */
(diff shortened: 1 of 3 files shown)


Yamagi:
Thank you!

PR #775, opened 2021-12-13
client curl update proposal.

devnexen:
displaying download progress based on a new cvar, 0 by default.


Yamagi:
Can you rename the commit so it doesn't have the "proposal" in it? That always looks somewhat ugly in the commit log. After that I'm gonna merge it.

devnexen:
sorry was trying to assess first if the idea itself would be welcomed :-) ok no pb.

Yamagi:
Thank you!

PR #778, opened 2022-01-12
cppcheck fixes

0lvin:
* small fix for shadowing ogg_file in function,
* add additional information about fan bug in mines,
* add const to function parameters.

### DIFF
--- a/src/client/input/sdl.c
+++ b/src/client/input/sdl.c
@@ -1167,10 +1167,12 @@ Haptic_Feedback(char *name, int effect_volume, int effect_duration,
 				return;
 			}
 
+			/*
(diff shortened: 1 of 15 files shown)


Yamagi:
Hi, I'll have a look at the weekend. :)

0lvin:
Thank you,

Last two commits is not related to fixes, its just improvements. If it is fine, i will port it also to vulkan render. I have tried port to soft render, but it looks like soft render scale draw has some issues related to 3/2 scale in some cases in menu. So i postponed port scale 3x images in soft render.

I have tried to look fan issues in mine map, reason of issues is not clear for now. so i just add additional note for debug.

Yamagi:
Aaand merged. Keep up the good work!

PR #789, opened 2022-02-12
Confirm delete save game through status bar update

apartfromtime:


Yamagi:
Aaand merged. Thank you! :)

PR #791, opened 2022-02-17
Introduce better cmake

bartekordek:
Hi!
I have added:
* Sdl 2 repo, so the build does not need it from somewhere else.
* Added cmake (tested under windows and linux).
* Minor fixes, so the project will compile with msvc.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "deps/curl"]
+	path = deps/curl
+	url = git@github.com:curl/curl.git
```
devnexen, line 3: encounters issue with url form. besides I m unsure these deps handling is necessarily that appealing on unixes. why not using vcpkg for example for windows ?

DanielGibson: I agree, we don't want (nontrivial) dependencies in the repo, and we don't want to build them as part of the yq2 build.

src/client/refresh/gl1/gl1_surf.c
```diff
@@ -210,7 +212,7 @@ R_DrawGLPolyChain(glpoly_t *p, float soffset, float toffset)
 
 			v = p->verts[0];
 
-            GLfloat tex[2*p->numverts];
+			GLfloat* tex = (GLfloat*)calloc(2 * p->numverts, sizeof(GLfloat));
```
DanielGibson, line 215: never call calloc() in a loop

Yamagi:
Just the high level stuff, I haven't looked into the code yet:

* Do not add dependencies into the repository. Neither by importing the code, nor by submodules.
* The build system must not download anything during build.
* We don't want any editor or IDE specific stuff in the repo. A lesson we've learned a long time ago.
* Contributed code (here miniz) must not be altered.

One big question remains: Why are all of these rather intruding changes in fragile and bug prone subsystems like the virtual filesystem are necessary? I understand that some users want better cmake support. But the ship for MSVC support has sailed years ago. We pondered a minimal invasive port for better integration into the MSVC debugger or WinDbg. Since then clang has added support for CodeView debug symbols and has become a first class citizen under Windows. When combined with the mingw libc it should be able compile the code with next to no changes.

DanielGibson:
I used to have a branch for MSVC support: https://github.com/DanielGibson/yquake2/tree/msvc

I updated it to current git master and fixed several issues (incl. alloca in loops): https://github.com/DanielGibson/yquake2/tree/MSVC-upd

Also fixed some general CMake issues there.
For external dependencies I checked out https://github.com/dhewm/dhewm3-libs/ and pointed CMake's YQUAKE2LIBS to the right directory, like `-DYQUAKE2LIBS=c:/path/to/dhewm3-libs/i686-w64-mingw32`

.gitmodules
bartekordek: So, do you want users to specifically build and install curl? Isn't it less demanding to have dependency (as already is, because is being used in code) automatically build and compile instead of the opposite? How about debugging if, for example, there is an error that leads to curl?

bartekordek:
> I used to have a branch for MSVC support: https://github.com/DanielGibson/yquake2/tree/msvc
> 
> I updated it to current git master and fixed several issues (incl. alloca in loops): https://github.com/DanielGibson/yquake2/tree/MSVC-upd
> 
> Also fixed some general CMake issues there. For external dependencies I checked out https://github.com/dhewm/dhewm3-libs/ and pointed CMake's YQUAKE2LIBS to the right directory, like `-DYQUAKE2LIBS=c:/path/to/dhewm3-libs/i686-w64-mingw32`

I have not seen this, thanks, I will look. After, I'll answer.

.gitmodules
devnexen: again package manager like conan/vcpkg are existing to make these sort of situations easier, especially if you plan to build several, for example, video games when you often see the same set of libraries ... sdl2/openal/opus etc ... you build/install once then that's it.

bartekordek:
Different assumptions about workflow and programmer experience.

PR #792, opened 2022-02-21
CL_ParseEntityBits overflow warning fix

devnexen:


Yamagi:
Thank you!

src/client/cl_parse.c
```diff
@@ -113,7 +113,7 @@ CL_ParseEntityBits(unsigned *bits)
 	/* count the bits for net profiling */
 	for (i = 0; i < 32; i++)
 	{
-		if (total & (1 << i))
+		if (total & (1u << i))
```
DanielGibson, line 116: I'm curious about this change: What kind of warning did you get for this code (from which compiler) that was fixed with using `1u`?
I assumed that some compilers might warn because `(int)1 << 31` returns a negative number, but I couldn't get any compiler on gcc.godbolt.org to warn about it (with `-Wall`), and I couldn't find anything about it being undefined or implementation defined behavior either (but didn't investigate that for too long).

devnexen: I got it with clang (`-Wshift-sign-overflow`), it seems "nitpicking" here for sure.

turol: Shifting a signed value so that a 1-bit ends up in sign bit is UB.

PR #795, opened 2022-02-27
Support building with Visual Studio (via CMake)

DanielGibson:
No idea if we want this, and I haven't tested it *that* much (but a bit at least), but it should generally be in an ok-ish shape now (no more `alloca()` in loops and such).
Also fixed some general issues in the CMakeLists.txt, like properly detecting the target CPU

~~Unfortunately, it currently requires VS2019 16.8 or newer; maybe this could be reduced to VS2015, if we can find a working alternative for `_Alignas(TYPE)` that doesn't require C11..~~
(`#define YQ2_ALIGNAS_TYPE( TYPE )  __declspec(align(__alignof(TYPE)))` looks like it should work, but doesn't, some compiler error about __builtin_alignof or sth like that)
**UPDATE:** Pushed a hack that should make it work with VS2015 and newer

Thanks to @bartekordek in #791 for inspiring me to dust up my [old hacky msvc branch](https://github.com/DanielGibson/yquake2/tree/msvc) which this is based on :)

### DIFF
--- a/src/common/filesystem.c
+++ b/src/common/filesystem.c
@@ -25,7 +25,9 @@
  * =======================================================================
  */
 
+#ifndef _MSC_VER
 #include <libgen.h>
+#endif
 
 #include "header/common.h"
 #include "header/glob.h"
@@ -380,7 +382,7 @@ FS_FOpenFile(const char *rawname, fileHandle_t *f, qboolean gamedir_only)
 	// Remove self references and empty dirs from the requested path.
 	// ZIPs and PAKs don't support them, but they may be hardcoded in
 	// some custom maps or models.
-	char name[MAX_QPATH] = {};
+	char name[MAX_QPATH] = {0};
 	size_t namelen = strlen(rawname);
 	for (int input = 0, output = 0; input < namelen; input++)
 	{
@@ -1614,6 +1616,27 @@ FS_GetNextRawPath(const char* lastRawPath)
 	return NULL;
 }
 
+#ifdef _MSC_VER // looks like MSVC/the Windows CRT doesn't have basename()
+// returns the last part of the given pathname, after last (back)slash
+// NOTE: this is not a fully compliant basename() implementation, as it doesn't
+//       handle trailing (back)slashes at all - because we don't need that.
+static char* basename( char* n )
+{
+	char* r1 = strrchr( n, '\\' );
+	char* r2 = strrchr( n, '/' );
+	if (r1 != NULL)
+	{
+		if (r2 != NULL)
+		{
+			return (r2 > r1) ? (r2 + 1) : (r1 + 1);
+		}
+		return r1 + 1;
+	}
+
+	return (r2 != NULL) ? (r2 + 1) : n;
+}
+#endif // _MSC_VER
+
 void
 FS_AddDirToSearchPath(char *dir, qboolean create) {
 	char *file;
@@ -1971,7 +1994,7 @@ static void FS_AddDirToRawPath (const char *rawdir, qboolean create, qboolean re
 	}
 
 	// Make sure that the dir doesn't end with a slash.
-	for (size_t s = strlen(dir) - 1; s >= 0; s--)
+	for (size_t s = strlen(dir) - 1; s > 0; s--)
 	{
 		if (dir[s] == '/')
 		{
--- a/src/common/filesystem.c
+++ b/src/common/filesystem.c
@@ -1618,21 +1618,19 @@ FS_GetNextRawPath(const char* lastRawPath)
 
 #ifdef _MSC_VER // looks like MSVC/the Windows CRT doesn't have basename()
 // returns the last part of the given pathname, after last (back)slash
-// NOTE: this is not a fully compliant basename() implementation, as it doesn't
-//       handle trailing (back)slashes at all - because we don't need that.
+// if the last character is a (back)slash, it's removed (set to '\0')
 static char* basename( char* n )
 {
-	char* r1 = strrchr( n, '\\' );
-	char* r2 = strrchr( n, '/' );
-	if (r1 != NULL)
+	size_t l = strlen(n);
+	while (n[l - 1] == '\\' || n[l - 1] == '/') // cut off trailing (back)slashes, if any
 	{
-		if (r2 != NULL)
-		{
-			return (r2 > r1) ? (r2 + 1) : (r1 + 1);
-		}
-		return r1 + 1;
+		--l;
+		n[l] = '\0';
 	}
-
+	char* r1 = strrchr(n, '\\');
+	char* r2 = strrchr(n, '/');
+	if (r1 != NULL)
+		return (r2 == NULL || r1 > r2) ? (r1 + 1) : (r2 + 1);
 	return (r2 != NULL) ? (r2 + 1) : n;
 }
 #endif // _MSC_VER
@@ -1652,11 +1650,8 @@ FS_AddDirToSearchPath(char *dir, qboolean create) {
 
 	// The directory must not end with an /. It would
 	// f*ck up the logic in other parts of the game...
-	if (dir[len - 1] == '/')
+	if (dir[len - 1] == '/' || dir[len - 1] == '\\')
 	{
-		dir[len - 1] = '\0';
-	}
-	else if (dir[len - 1] == '\\') {
 		dir[len - 1] = '\0';
 	}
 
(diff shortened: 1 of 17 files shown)


mjr4077au:
Between this or the other commit, I'd certainly use it. While I can build in mingw, MSVC's debugger on Windows is second to none and I can't stand debugging in GDB.

Yamagi:
I'm not convinced if we / I really want this. These days Yamagi Quake II is (more or less) in maintenance mode. Nearly all features, which could by added, while staying compatible with Vanilla Quake II and not changing the gameplay, have been added. The feature requests, which are still open, are either some corner cases or rather trivial to implement. I've learned a lot from Yamagi Quake II, the project depicts my personal development from a university freshman to a IT professional. But life has moved on.  While we / I will continue to fix bugs, give user support and maybe implement some smaller new features, it's unlikely that Yamagi Quake II will ever see fast passed development or ground breaking new features again.

MSVC support by itself isn't a problem if done right. But, and that's a big but, it is another platform I / we need to support. It require test builds after code changes, there might be bug reports, etc. Exactly the points which prevented me from adding MSVC support even when I was spending whole weeksends hacking Yamagi Quake II. If it's just about debugger support, it might be better to look into refactoring the Makefile from hardcoded assumptions around mingw to something more generic which can use clang. clang can generate MSVC compatible debug symbols, I've used that for several other projects. When combined with the mingw libc it's just another gcc compatible platform without any maintenance overhead.

DanielGibson:
> MSVC support by itself isn't a problem if done right. But, and that's a big but, it is another platform I / we need to support. It require test builds after code changes

We already support kinda obscure platforms that we don't test ourselves, like Haiku.

As we're not planning to do major changes anymore, the risk of breaking a platform is relatively low, and if we do, the fix will (most probably) be small and easy as well.
So I think it's ok to not test all platforms regularly ourselves, and just fix things if/when someone reports it broken.

I'd also argue that most of the changes here are pretty harmless, the biggest/most invasive ones are:
1. All the changes to CMakeLists.txt - much of this isn't even MSVC-specific but works around a severe CMake bug (that they're considering expected behavior now)
2. The VLA stuff - yes, this is ugly, but most of it is in (parts of) GL1 code that we're unlikely to ever touch again

src/common/filesystem.c
```diff
@@ -1614,6 +1616,27 @@ FS_GetNextRawPath(const char* lastRawPath)
 	return NULL;
 }
 
+#ifdef _MSC_VER // looks like MSVC/the Windows CRT doesn't have basename()
+// returns the last part of the given pathname, after last (back)slash
+// NOTE: this is not a fully compliant basename() implementation, as it doesn't
+//       handle trailing (back)slashes at all - because we don't need that.
+static char* basename( char* n )
```
devnexen, line 1623: would `_splitpath` do the job ?

DanielGibson: nope, it cuts off the file extension

src/common/header/shared.h
```diff
@@ -65,12 +65,24 @@ typedef unsigned char byte;
 	// must be used as prefix (YQ2_ATTR_NORETURN void bla();)!
 	#define YQ2_ATTR_NORETURN       __attribute__ ((noreturn))
 #elif defined(_MSC_VER)
-	#define YQ2_ALIGNAS_SIZE(SIZE)  __declspec( align(SIZE) )
-	#define YQ2_ALIGNAS_TYPE(TYPE)  __declspec( align( __alignof(TYPE) ) )
+	// Note: We prefer VS2019 16.8 or newer in C11 mode (/std:c11), 
+	//       then the __STDC_VERSION__ >= 201112L case above is used
+
+	#define YQ2_ALIGNAS_SIZE(SIZE)  __declspec(align(SIZE))
+	// FIXME: for some reason, the following line doesn't work, which is why we require C11 support for MSVC
```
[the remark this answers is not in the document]
DanielGibson: I don't mind these handful of lines for a workaround if it means I don't have to explain the difference between VS2019 and VS2019 16.8+ to people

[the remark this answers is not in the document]
devnexen: alrighty..

devnexen:
> > MSVC support by itself isn't a problem if done right. But, and that's a big but, it is another platform I / we need to support. It require test builds after code changes
> 
> We already support kinda obscure platforms that we don't test ourselves, like Haiku or macOS.

Good things with those, their specific code paths are minimal and well concealed, even a bizarre system (not saying it in condescending way but rather affectionate) like haiku which become more and more posix compliant as the releases occur. MSVC code path is a lot more invading ...

> 
> I'd also argue that most of the changes here are pretty harmless, the biggest/most invasive ones are:
> 
> 1. All the changes to CMakeLists.txt - much of this isn't even MSVC-specific but works around a severe CMake bug (that they're considering expected behavior now)
> 2. The VLA stuff - yes, this is ugly, but most of it is in (parts of) GL1 code that we're unlikely to ever touch again

... and more bug prone. However, saying this I have nothing against MSVC support if everyone wants to go to that direction.

src/common/header/shared.h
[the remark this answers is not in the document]
DanielGibson: yeah, it's surprising that MS decided to bring in such a big feature (C11+C17 support) in a minor subrelease of a VS version, I wish they already had this in the initial version of VS2019

(and I wish they had VLA support, then this PR would be a no-brainer..)

src/common/filesystem.c
DanielGibson: yeah, using _splitpath (or more likely _splitpat_s) and then concatenating things again etc isn't really easier or less error-prone than this

src/common/header/shared.h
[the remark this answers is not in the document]
devnexen: nonetheless if we go towards MSVC support, as said above I prefer this PR over the other.

DanielGibson:
> Good things with those, their specific code paths are minimal and well concealed, even a bizarre system (not saying it in condescending way but rather affectionate) like haiku which become more and more posix compliant as the releases occur.

Note I'm not dunking on Haiku here, I think it's great to support non-mainstream systems :-)
Just pointing out that we already have code that neither Yamagi nor me test (or even just build) ourselves.

> ... and more bug prone.

Sure, but again, it's not *that* much code using VLAs so it's manageable.

OTOH, supporting another compiler (that doesn't try to emulate GCC) makes the code a bit more portable overall, and MSVC's diagnostics actually help *finding* bugs, see those WCHAR changes.
And it opens the door to use debugging or profiling tools that are VS-centric when it makes sense, which can be a nice addition to the stuff we get from the gcc/clang/Linux world.

And of course supporting VS makes it more likely to get contributions from developers using Windows who are used to it.

Side-rant: *I have no idea how people are productive with VS, I find using it for development (and not just compiling and a bit of debugging) painful because everything that has to do with code navigation like "find references" or especially "show call hierarchy" is ridiculously slow, but most developers who primarily use Windows, like most game developers, use it.. maybe it gets more usable with commercial plugins like resharper or visual assist or sth, I don't know. It is what it is :man_shrugging:*

maraakate:
You know I've been using VS since the days of VC6 and the newer VS has come a long way.  It still has it's nitpicks.  The intellisense/find references/etc. would be nice if it could scan the project properly and the find references is kind of poor if it's a heavy C++ project with templates and overloads (lord have mercy if you're using it in a large .NET project).  However, I do think it's debugger is top notch.  GDB is usable, but makes things a lot harder than it needs to be.  There's front ends, but I've never been impressed with any of them.  It's trivial to readapt the try/catch exception handler code from R1Q2 to auto generate a dump that can be used to analyze crashed with VS.  I think that alone is a huge bonus.

Since your project is eligible for a free PVS Studio license you can integrate this easily in VS and then scan the project and you can simply double click go right to the offending line and fix it.

And yes, like you said, it may attract more Windows developers contributing.

DanielGibson:
BTW: At least for Win32 (x86; didn't test x64) savegames created with YQ2 built with MSVC vs MinGW seem to be compatible

src/common/filesystem.c
```diff
@@ -1614,6 +1616,23 @@ FS_GetNextRawPath(const char* lastRawPath)
...
```
Yamagi, line 1623: What about turning this function into a global Q_basename() function?

DanielGibson: Also a possibility - but I'm not sure how much sense it makes as long as basename() is only used in this one place and a custom implementation is only needed for one platform/compiler

DanielGibson:
rebased it onto current master, changed basename() implementation a bit

DanielGibson:
Building with VS is now also documented: https://github.com/yquake2/yquake2/blob/master/doc/020_installation.md#compiling-on-windows-with-visual-studio-2015-and-newer

PR #797, opened 2022-02-28
cmake limiting the scope of concern to C only and detecting

devnexen:
the compiler from to one particular frontend since it can
happen to have differents for C and C++.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@ if(NOT CMAKE_BUILD_TYPE)
 endif()
 
 # CMake project configuration.
-project(yquake2)
+project(yquake2 C)
 
 # Cmake module search path.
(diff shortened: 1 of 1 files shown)


DanielGibson:
makes sense, I assume the usage of CMAKE_CXX_COMPILER_ID was a copy&paste fail.

one note though, my CMake complained about the quotes around `"${CMAKE_CXX_COMPILER_ID}"` being deprecated behavior or sth (which is why I removed them in my MSVC branch)

devnexen:
you re surely right but that might depends on cmake version tough since we support from 3.0 which is already not brand new anymore.
the only case we would need to care about C++ if you needed to use win32 C++ api wrapped as C symbols perhaps but you do not seem to need such.

DanielGibson:
We don't use C++, the Windows APIs we use are plain C

PR #803, opened 2022-03-20
Add r_lerp_list and r_videos_unfiltered CVars, fix #800

### LINKED ISSUE #800: Disable 2D Texture Filtering (HUD, Menu GFX, etc.)
Could we get an option to turn off the texture filtering that's specifically applied to HUD and menu elements? I think these were designed without the filtering in mind, since in the vanilla exe, they aren't filtered at all. The only part that _is_ filtered is the console background it seems.

I know that there's a cvar to disable certain graphics from using texture filtering, but that requires you to disable them all in a list, typing them out individually.

DanielGibson:
r_lerp_list is to allow exceptions to r_2D_unfiltered (like for having
pixely UI in general, but filtered console background).

r_videos_unfiltered controls whether videos should be filtered or not

I also made r_nolerp_list CVAR_ARCHIVE, like users probably expect it.

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -289,10 +289,19 @@ it's `+set busywait 0` (setting the `busywait` cvar) and `-portable`
   anti aliasing is expensive and can lead to a huge performance hit, so
   try setting it to a lower value if the framerate is too low.
 
-* **r_nolerp_list**: list separate by spaces of textures omitted from
-  bilinear filtering. Used by default to exclude the console and HUD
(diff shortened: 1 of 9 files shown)


DanielGibson:
Heads-up to @0lvin: You may wanna add these features to the Vulkan renderer :)

0lvin:
thank you! i will look.

0lvin:
Ported: https://github.com/yquake2/ref_vk/commit/5a133b7304654a72f6ce911aa1acf3c04950a7f8

PR #804, opened 2022-03-20
WIP: Add color light to soft render

0lvin:


0lvin:
Current results:
* 95fps without changes
* 75fps with color light (~22% regression)

[Color light in softrender](http://www.youtube.com/watch?v=6EsbpkWADNM).

Yamagi:
Wow, it must have been a lot of work to merge the colored lightning stuff back into our good. Or did you reimplement it? Nevertheless, that's very good work :) On more my Ryzen I come to nearly the same values. ~96 fps without the patch, ~77 fps with the patch. Am I seeing it right that the overhead is always there, regardless if `r_colorlight` is set to `0` or `1`?

0lvin:
I have reimplemented code, difference between r_colorlight is only in usage middle color in places where color get from light source or from lightmap in model. So code always use 3 component colors - and just have simulated canonical lighting.

Yamagi:
I did some more performance testing on my old Thinkpad T430s with an Intel Core i5-3320M. That's the slowest system I have. Everything at 1600x900:

* GL3: 131.8 FPS
* GL1: 109.7 FPS
* Soft, unpatched: 57.7 FPS
* Soft, patched, sw_colorlight 0: 54.7 FPS
* Soft, patched, sw_colorlight 0: 53.6 FPS

The regression is ony

DanielGibson:
Sorry, but this looks *very* different than before, much darker:

before:
image
now:
image

I think it looked better before, somehow it's pretty grainy now

DanielGibson:
I think it looks better now, but still not quite the same:
soft old:
q2_0141
soft new, sw_colorlight 0:
q2_0142
soft new, sw_colorlight 1:
q2_0143
gl3 `gl_texturemode GL_NEAREST`:
q2_0146
gl3 regular:
q2_0144
gl1 (too dark because gamma correction is done in hardware and thus doesn't show on screenshot):
q2_0145

0lvin:
6917b53 - should help with colors (as used old logic for color + light logic) but in same time it makes code slower in case color light.

DanielGibson:
BTW, I think the reason the original softrenderer was so bright (compared to the original GL renderer) is that apparently they used `lightstyle_t::white` for the light values, which is calculated as `r+g+b` - instead of `(r+g+b)/3.0` as one might expect, which would return (roughly) the same brightness as the colored light.

(No idea if this is news for you, but I thought I'd mention it)

0lvin:
thank you. I have missed such. I will look

Yamagi:
Thank you for all the hard work. I'm happy with this, I'm merging this now. Additional changes can be done in a separate PR; of course. :)

PR #811, opened 2022-04-07
Added cvar to define SDL gamepad's 'Back' button

protocultor:
This allows to redefine the button used as the "**Esc key**" in the SDL controller, by default the _Back/Select/Minus_ (center left) button. Now it can be changed to the _Start/Menu/Plus_ (center right) button, or even to the _Guide/Home/PS_ button (middle of the gamepad, normally big). Requires a game restart to take effect.
Useful for people not happy with the current location of the button.


Yamagi:
And merged. Thank you :)

PR #812, opened 2022-04-12
Fixes for soft render with disabled color lighting

0lvin:


0lvin:
Changes contains only changes for non color light. 

Color lighting can be wrong as for me and requires additional changes, look to 
```
	if ((light_masked[0] | light_masked[1] | light_masked[2]) <= vid_lightthreshold)
	{
		// FIXME: color conversion should be applied and based on vid_colormap
		return pix;
	}
```

0lvin:
I have reverted code to use colormap directly. it should be best solution for now.

Yamagi:
Looks good.Thank you :)

PR #814, opened 2022-04-16
Load gamecontrollerdb.txt for SDL2 joy mappings

protocultor:
This allows for the loading of _gamecontrollerdb.txt_, database of game controller mappings to be used with SDL2 controllers.
This is _optional_ and its an alternative to the use of setting the SDL_GAMECONTROLLERCONFIG variable to use a gamepad.
It's practical for gamepads that change their GUID when used wired or bluetooth, e.g. Nintendo Switch Pro Controller on Linux.
The text file is usually obtained from https://github.com/gabomdq/SDL_GameControllerDB, and can be put in both the game or "config" directories.


Yamagi:
Cool, thank you for this!

PR #816, opened 2022-04-24
frame cpu pause refactoring.

devnexen:
expanding to powerpc archs also for visual studio/MSVC while
keeping clang support for windows.

### DIFF
--- a/src/common/frame.c
+++ b/src/common/frame.c
@@ -81,6 +81,25 @@ char userGivenGame[MAX_QPATH];
 // Hack for the signal handlers.
 qboolean quitnextframe;
 
+static inline void Sys_CpuPause(void)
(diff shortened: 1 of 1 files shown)


turol:
This should be an inline function in the header. Adding function call overhead for a single instruction is kind of silly.

devnexen:
You re right it seems it was not really the trend of the overall code doing so so I did not dare but definitively makes more sense.

DanielGibson:
As it's only used in one place, it doesn't even have to be a function at all but could remain inline.

I'd even argue that it *should* remain inline, because with the change we'd have duplicated code in the two versions of system.c

Yamagi:
And merged, thank you :)

PR #820, opened 2022-04-25
Update retexturing command in configuration document.

ConHuevosGuey:
Update to the configuration document.  Removed the deprecated command 'gl_retexturing' and replaced it with the current command 'r_retexturing'


Yamagi:
Ah, good find. :) Merged!

PR #822, opened 2022-04-26
OpenGL ES 3.0 support

DanielGibson:
I modified the gl3 renderer to additionally support gles3.
Now both ref_gl3.dll/so/dylib and ref_gles3.dll/so/dylib are always built.

Unlike the gl3 renderer, this also works on the RPi4 (though the GL1 renderer is still much faster).


devnexen:
I gave a shot at your branch. While it works well on Linux, note there is no opengl es for macos but on ios.

DanielGibson:
Good to know, thanks for testing! :)

I wonder if we should disable building ref_gles3.dylib, or if we should keep it around in case someone wants to try it with ANGLE or MoltenGL ever supports GLES3 or whatever other wild idea people might get.

devnexen:
might be wiser to disable it for now.

DanielGibson:
Or we leave it as it is and ask the MacSourcePorts guy to not package it

devnexen:
The OpenGL situation on apple devices will never evolve, it is deprecated :-) nowadays it s Metal with its cuda-like specific language to play with the GPU ;-) but that's fair if you prefer changes in the packages side.

DanielGibson:
I know, and ANGLE translating OpenGL ES to Metal could help when Apple's OpenGL support breaks further (or is completely removed)

Yamagi:
I'm merging this now. If we want or must disable building the OpenGL ES renderer lib under Apple, we can do it at a later time.

Yamagi:
A new Windows testbuild with the OpenGL ES 3.0 renderer lib included can be found here: https://deponie.yamagi.org/quake2/windows/testbuilds/quake2-20220507-e294fc9.zip

PR #825, opened 2022-05-07
curl client show progress only when there actually some

devnexen:


Yamagi:
Good idea. Thank you!

Issue #802, opened 2022-03-19
[Feature] Separate Menu for Controller Binds


OpenRift412:
It can get a bit clausterphobic in the keybinds menu trying to configure controller binds without overriding both of your keybinds. I think to make it a bit more intuitive and less cluttered, controllers should have their own binds menu.

protocultor:
I think it's time to close this ;)

protocultor:
Since https://github.com/yquake2/yquake2/pull/827 fixes this and has been merged a while ago, and YQ2 8.10 includes it, shouldn't this be closed by now?

protocultor:
> It can get a bit clausterphobic in the keybinds menu trying to configure controller binds without overriding both of your keybinds. I think to make it a bit more intuitive and less cluttered, controllers should have their own binds menu.

This was solved by https://github.com/yquake2/yquake2/pull/827, and is included in YQ2 8.10.
I believe that, unless there's some feature missing, this is worth closing.

PR #827, opened 2022-05-08
Full transition to SDL_GameController for gamepad usage

protocultor:
Currently, _client/input/sdl.c_ handles input from buttons of gamepads/joysticks through SDL_Joystick, which depending on the OS and device may not be the best option. I.e. on MacOS High Sierra, DPad on the Nintendo Switch Pro Controller is not recognized by SDL2 as a hat but as buttons, making it unsuitable to handle game menus (weird, because Cocoa detects it as a hat).
Using SDL_GameController increases compatibility of devices that can be used with YQ2, across any OS supported by SDL2.
"Unrecognized buttons" is no longer an issue since 2.0.14, which recognizes Xbox Elite paddles and other buttons on newer gamepads.

New features:
- Named buttons, consistent across multiple gamepads.
- Improved menu handling on controller: "A" to select, "B" to get back, "Y" to delete.
- Separate menus to customize controller buttons, separated from the classic menu option (implements https://github.com/yquake2/yquake2/issues/802).
- Hot swapping: disconnect/connect a controller with the game running with no other problem than a small delay.

### DIFF
--- a/src/client/cl_keyboard.c
+++ b/src/client/cl_keyboard.c
@@ -135,89 +135,59 @@ keyname_t keynames[] = {
 	{"MWHEELUP", K_MWHEELUP},
 	{"MWHEELDOWN", K_MWHEELDOWN},
 
-	{"JOY1", K_JOY1},
-	{"JOY2", K_JOY2},
-	{"JOY3", K_JOY3},
-	{"JOY4", K_JOY4},
-	{"JOY5", K_JOY5},
-	{"JOY6", K_JOY6},
-	{"JOY7", K_JOY7},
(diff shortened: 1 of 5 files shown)


Yamagi:
I'll need a few days to review this.

Dremor8484:
> I'll need a few days to review this.

i wonder if you will find the time to also review my pull request, if i manage to make one :P (i am still coding)

Yamagi:
This is without any doubt very good work. Thank you for that! And therefore: Merged. :) Of course I would have looked at it if had been from @Dremor8484 ;)

Issue #819, opened 2022-04-25
Elevators and doors stutter


dezo2:
When using 144Hz/FPS with vsync on and default settings (cl_async 1), there are some nasty stutters on elevators, opening doors etc. It is somewhat smoother with cl_async 0, but then the physics can be broken according to yq2 docs. Is there some other cvar I can try to fix this? Tried the same map sections in q2pro and there was no stutter at all. Here is a save before such elevator in Ammo Depot map: [save4.zip](https://github.com/yquake2/yquake2/files/8555681/save4.zip)

dezo2:
Nevermind, it works smooth enough so far with cl_async 0.

DanielGibson:
this should also (and especially) work with cl_async 1

dezo2:
That would be great.
The game movement is OK overall, only the elevators / doors and similar linear animations exhibit these skips, Framerate is constant 144FPS all the time with flat frametime chart, GPU usage around 5%. Fiddling with cvars cl_maxfps, vid_maxfps, r_vsync didn't fix it until cl_async is set to 0 or cl_maxfps is set to 144 from default 60 to match display refresh. Which I guess is the same as set cl_async to 0.

Yamagi:
This could be caused by the special case for stairs and evelators: https://github.com/yquake2/yquake2/blob/master/src/client/cl_prediction.c#L298=

Does setting `cl_predict 0` help? That isn't a fix, just to figure out were to look.

Yamagi:
Position in the attached savegame is bunk1.bsp / -678 -102 46.

dezo2:
cl_predict 0 makes it even more visible and also adds skips when strafing.
So far the best settings for async for me is cl_maxfps 144 and leave the rest at default.

DanielGibson:
maybe cl_maxfps 60 (or maybe 63 or 64) and vid_maxfps 120 could work well?

Yamagi:
First: q2pro is a very good Quake II engine. From a source code standpoint likely the best. Since they effectively dropped singleplayer support and concentrate on multiplayer, they can do a lot of architectural changes and bugfixes which aren't possible for YQ2. We focus on singleplayer and mod compatibility. We try to stay mostly bug-compatible with Vanilla Quake II, even if it means that we cannot fix some problems.

[...]

PR #830, opened 2022-05-09
Improve timings in Qcommon_Frame()

DanielGibson:
* packet framerate doesn't have to be lower than render framerate (unless render framerate is too high) => that 0.95 factor didn't really help
* Improve the max framerate-handling with vsync, around vid_maxfps and the display refreshrate, especially considering the reported display refresh might be a bit too low
* The packet framerate should be a fraction of the render framerate; and it should be between 45 and 90 to prevent movement glitches and such. Setting `cl_maxfps` to `-1` now automatically chooses an appropriate packet framerate that meets those criteria.
    - If you want it extra smooth, make sure your render framerate is a multiple of 60Hz (by setting vid_maxfps accordingly, or use vsync with a 60Hz display)

Might fix (or at least improve) #819

### DIFF
--- a/src/client/header/client.h
+++ b/src/client/header/client.h
@@ -211,7 +211,7 @@ typedef struct
 	keydest_t	key_dest;
 
 	int			framecount;
-	int			realtime; /* always increasing, no clamping, etc */
+	int			realtime; /* always increasing, no clamping, etc, in MS */
 	float		rframetime; /* seconds since last render frame */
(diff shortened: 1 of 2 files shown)


Yamagi:
A new Windows testbuild with this timing improvements included can be found here: https://deponie.yamagi.org/quake2/windows/testbuilds/quake2-20220510-1833f16.zip

PR #838, opened 2022-05-18
adding aimfix to double barrel shotgun

Dremor8484:
adding aimfix to double barrel shotgun


maraakate:
Why?

Dremor8484:
> Why?

to center the bullet spread

maraakate:
I can't speak for Yamagi, but changes like these are likely out of the scope of YQ2 and really should be a separate mod.  But, maybe it will get merged.

Dremor8484:
some extra changes i made after calling the pull request on double barrel shotgun aimfix got added here

the extra changes are about adding a new server variable sv_custom_settings, while this variable is set to 1 the values of player health, max health, max ammo (and in the future weapon, items and monsters will be able to be customized if needed)

a simple autoexec.cfg file can decide the new values if present, otherwise the default values are loaded

apartfromtime:
The custom stuff would make a good mod. If not in the scope yq2. Not sure you should be using the "autoexec.cfg" for changes, is it even used for anything anymore? If you do make a mod from your changes, a "default.cfg" with the custom cvars in the mod directory has the same effect.

mjr4077au:
I'm also the person who ported over `aimfix` from Berserker. I don't have access to a build system to check the application of `aimfix` to the shotgun, but a quick play does show the pellet spread is not centered so this could be welcome. Please note though that the fix isn't purely cosmetic, it actually properly centers the trajectory of projectiles as well.

Also as a passing comment, this is a pretty messy PR... Never do pull requests from your master branch, do a branch per change so the PR stays relevant. This PR can't be merged as is because all of your tinkering in master is now part of this PR.

Yamagi:
If the `aimfix` for the shotgun is correct - I haven't checked - and calculates correct trajectories, we can merge it. That's what `aimfix` is for, fixing the slightly broken trajectory calculation for users who them correct.

The customization stuff isn't really in scope for YQ2. While we have added features in the past, we stick to feature that enhance the 'out of box experience'. Widescreen support, HTTP download, the `teleport` command for debugging, fixed bullet trajectories... We avoided "unnecessary" features and, at least when possible, everything which changes the game <-> server <-> client interactions. So this would be better done in a mod.

Dremor8484:
> I'm also the person who ported over `aimfix` from Berserker. I don't have access to a build system to check the application of `aimfix` to the shotgun, but a quick play does show the pellet spread is not centered so this could be welcome. Please note though that the fix isn't purely cosmetic, it actually properly centers the trajectory of projectiles as well.
> 
> Also as a passing comment, this is a pretty messy PR... Never do pull requests from your master branch, do a branch per change so the PR stays relevant. This PR can't be merged as is because all of your tinkering in master is now part of this PR.

thanks for the tip i never used github before... will try to do as you suggested

Dremor8484:
> If the `aimfix` for the shotgun is correct - I haven't checked - and calculates correct trajectories, we can merge it. That's what `aimfix` is for, fixing the slightly broken trajectory calculation for users who them correct.
> 
> The customization stuff isn't really in scope for YQ2. While we have added features in the past, we stick to feature that enhance the 'out of box experience'. Widescreen support, HTTP download, the `teleport` command for debugging, fixed bullet trajectories... We avoided "unnecessary" features and, at least when possible, everything which changes the game <-> server <-> client interactions. So this would be better done in a mod.

ok i will create a pull request of only the aimfix

PR #840, opened 2022-05-20
Update weapon.c

Dremor8484:
aimfix cvar now applies to the doublebarrel shotgun too

### DIFF
--- a/src/game/player/weapon.c
+++ b/src/game/player/weapon.c
@@ -1689,6 +1689,18 @@ weapon_supershotgun_fire(edict_t *ent)
 			DEFAULT_SSHOTGUN_COUNT / 2, MOD_SSHOTGUN);
 	v[YAW] = ent->client->v_angle[YAW] + 5;
 	AngleVectors(v, forward, NULL, NULL);
+	
+	if (aimfix->value)
+	{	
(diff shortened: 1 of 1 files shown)


Dremor8484:
i prepared the pull request from a branch of my fork, only the code used for aimfix double barrel shotgun is used.

i noticed that the double barrel shotgun did not have aim fix, when i changed its variables to reduce horizontal spread...
suddenly it was much more noticeable that all bullets hit on the bottom right side of the aiming cross.

i picked the same code used to fix the aim of the shotgun, and i copied in the necessary spot to achieve the correction effect. and it worked.

with the default spread of the double barrel shotgun the not fixed bullets appear to hit 95% below the cross and 25% on its left side while 75% on the right side.

Yamagi:
And finally merged. Thank you. Can you do the same for rogue and xatrix?

Dremor8484:
> And finally merged. Thank you. Can you do the same for rogue and xatrix?

sure, just gimme a few days cause hell is happening at home :P

i will use the procedure someone told me here, about
 1) fork,
 2) create a fork of my fork with just the same fix used here
 3) initiate pull request of the fork of my fork in your yamagi-master
 4) wait XD

PR #841, opened 2022-05-21
Add menu bitmap support

apartfromtime:
Quality of life improvement, this makes the code easier to read, and see how the retained mode menu functions.

### DIFF
--- a/src/client/menu/header/qmenu.h
+++ b/src/client/menu/header/qmenu.h
@@ -86,11 +86,11 @@ typedef struct
 
 typedef struct
 {
-	menucommon_s	generic;
-	char *			focuspic;	
-	char *			errorpic;
-	int				width;
-	int				height;
+	menucommon_s    generic;
+	char *          focuspic;	
+	char *          errorpic;
+	int             width;
+	int             height;
 } menubitmap_s;
 
 typedef struct
--- a/src/client/menu/header/qmenu.h
+++ b/src/client/menu/header/qmenu.h
@@ -32,16 +32,18 @@
 
 #define MAXMENUITEMS 64
 
-#define MTYPE_SLIDER 0
-#define MTYPE_LIST 1
-#define MTYPE_ACTION 2
-#define MTYPE_SPINCONTROL 3
-#define MTYPE_SEPARATOR 4
-#define MTYPE_FIELD 5
+#define MTYPE_SLIDER        0
+#define MTYPE_LIST          1
+#define MTYPE_ACTION        2
+#define MTYPE_SPINCONTROL   3
+#define MTYPE_SEPARATOR     4
+#define MTYPE_FIELD         5
+#define MTYPE_BITMAP        6
 
-#define QMF_LEFT_JUSTIFY 0x00000001
-#define QMF_GRAYED 0x00000002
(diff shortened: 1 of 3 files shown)


src/client/menu/header/qmenu.h
```diff
@@ -82,6 +84,15 @@ typedef struct
 	void (*cursordraw)(void *self);
 } menucommon_s;
 
+typedef struct
+{
+	menucommon_s	generic;
+	char *			focuspic;	
+	char *			errorpic;
+	int				width;
+	int				height;
```
0lvin, line 93: Looks as it has wrong indentation.

src/client/menu/menu.c
```diff
@@ -569,95 +607,144 @@ M_Main_Draw(void)
     {
         Draw_GetPicSize(&w, &h, names[i]);
 
-        if (w > widest)
-        {
+        if (w > widest) {
             widest = w;
```
0lvin, line 611: could we use origin code style?

Dremor8484: > looks as only code style change, could code have original code style?

*coug* *cough* i was wondering if you could have some spare time to review my pull request XD

apartfromtime: I may have retyped that. Yes code could have original code style. Does Yamagi have a code style convention? Are submissions required to follow the original id software code style convention?

src/client/menu/header/qmenu.h
apartfromtime: I'll fix.

src/client/menu/menu.c
apartfromtime: @Dremor8484 Your pull request won't go unnoticed, could you keep any comments related to this commits conversation. :)

Yamagi:
And finally merged. Thank you for your patience.

protocultor:
Sorry for ruining the fun, but this PR broke the binding menus; try to enter in "customize controls" in the options menu with the current "master" branch, the game will hang.
The problem is caused by using **QMF_GRAYED** for menu elements that must be skipped... but every element in the bind menus use it.

In the file src/client/menu/qmenu.c, if you search for every instance of:
```
if (citem->type != MTYPE_SEPARATOR &&
	(citem->flags & QMF_GRAYED) != QMF_GRAYED)
```
... and you restore it to the previous:
`
if (citem->type != MTYPE_SEPARATOR)
`
... the game will work again, but the main menu breaks; the left side pictures become selectable. As the original author has a better idea on how his idea works, I think @apartfromtime could fix this.

Yamagi:
I merged this pull request, because a better video menu with more generic options and a per-renderer submenus is an often requested features. This changes make it easier to fiddle dynamic submenus into the code...

I've committed a band-aid fix working around the infinite loop. I may be better to change the loops in `Menu_AdjustCursor()` to one iteration only. With the band-aid fix, there's still something wrong in all key binding / action menus. The left column is white, it should be green. @apartfromtime can you have a look?

apartfromtime:
I pushed a fix. The QMF_GRAYED is an exclusive menu flag that affects presentation, I made a choice on using it.

> It may be better to change the loops in `Menu_AdjustCursor()` to one iteration only.
Agreed.

PR #843, opened 2022-05-21
freebsd/clang and server only warnings build fix

devnexen:


src/client/refresh/soft/sw_main.c
```diff
@@ -1042,8 +1042,8 @@ RotatedBBox (const vec3_t mins, const vec3_t maxs, vec3_t angles, vec3_t tmins,
 
 	for (i=0 ; i<3 ; i++)
 	{
-		tmins[i] = INT_MAX; // Set maximum values for world range
-		tmaxs[i] = INT_MIN;  // Set minimal values for world range
+		tmins[i] = (vec_t)INT_MAX; // Set maximum values for world range
```
0lvin, line 1045: What is a warning shown by clang?  

You can use 99999.0 / -99999.0 (some really big/small value) as it was before 025764db79e805c035dd634a08e86289409fe850.

0lvin (approved):

src/client/refresh/soft/sw_main.c
devnexen: oh nothing dramatic really and the warnings show only on BSD complaining about assigning int to real.

Yamagi:
Merged. Thank you!

PR #844, opened 2022-05-22
Inventory screen keys/buttons display fix

protocultor:
When a key/button name is too long, and its bound to an inventory element, its display in the inventory screen misaligns the item it is bound to, even pushing it out of the inventory frame.

old

This single-line fix limits its display to the maximum of 6 characters the inventory screen allows.

new

### DIFF
--- a/src/client/cl_inventory.c
+++ b/src/client/cl_inventory.c
@@ -139,7 +139,7 @@ CL_DrawInventory(void)
 			}
 		}
 
-		Com_sprintf(string, sizeof(string), "%6s %3i %s", bind,
+		Com_sprintf(string, sizeof(string), "%6.6s %3i %s", bind,
 				cl.inventory[item], cl.configstrings[CS_ITEMS + item]);
 
 		if (item != selected)


apartfromtime:
Cannot see what controller stick the quad damage is bound to.  Maybe the key number to string bindings should have long and short abbreviated names? Or scrolling text?

There was similar issue recently involving the "help computer" and the "nightmare" difficulty setting being displayed through it.

protocultor:
> Cannot see what controller stick the quad damage is bound to. Maybe the key number to string bindings should have long and short abbreviated names? Or scrolling text?

True, but both of those solutions require much bigger, more intrusive work. What I'm proposing here is much simpler, clear to understand and guaranteed to not produce any further errors. This is, at worst, a "good enough" patch until either of what you mention gets implemented.

Also, for now, a truly interested player can check out the new gamepad buttons' menus to see what it is bound to quad damage.

> There was similar issue recently involving the "help computer" and the "nightmare" difficulty setting being displayed through it.

In that one, no solution left everyone happy either. The problems of targeting 320x240 resolution :)

Yamagi:
> The problems of targeting 320x240 resolution :)

And we want to keep 320x240, because there're some nerds using it. Mostly for sentimental reasons, I think.

Anyways, thank you for the PR. :)

protocultor:
> And we want to keep 320x240, because there're some nerds using it. Mostly for sentimental reasons, I think.

I'm one of those, I play yquake2 on a Raspberry Pi 3 B+ (among other platforms). Don't get rid of software rendering either, please :)

Thanks to you for merging this!

PR #845, opened 2022-05-27
Check >= maxsize in Info_SetValueForKey().  Found by ASAN.

maraakate:
This might be rare, but I found in Daikatana (which uses the same exact code here) that ASAN triggered a warning from the *s = 0; at the end of Info_SetValueForKey().  This can happen if you have a lot of SERVERINFO (especially custom ones like 'admin', 'email', and 'website') flagged CVARs and someone does a /serverinfo packet.  In Daikatana, it was more prevalent use of UB since the same command was sent every time a map transition happened on a public server to send the serverinfo stats to the gamespy master server.

This bug exists in R1Q2 and all varieties of Info_SetValueForKey() out there that I am aware of.  So the mods would need this fix as well.

### DIFF
--- a/src/common/shared/shared.c
+++ b/src/common/shared/shared.c
@@ -1398,7 +1398,7 @@ Info_SetValueForKey(char *s, char *key, char *value)
 
 	Com_sprintf(newi, sizeof(newi), "\\%s\\%s", key, value);
 
-	if (strlen(newi) + strlen(s) > maxsize)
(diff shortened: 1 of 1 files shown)


Yamagi:
At least rare enough that it was never reported until now. That's a good find. I've ported the change to xatrix and rogue. It will be included in the next release, maybe already tomorrow.

PR #849, opened 2022-06-07
Gyro aiming

protocultor:
Game controllers with a gyroscopic sensor has been a thing since years, and Steam and some input remappers provide support for mouse-controlled games to use it, since a gyro is in practice a mouse that is controlled through rotation instead of translation. With Nintendo, Sony, and now the Steam Deck supporting it, this can only get more popular, because gyro provides much better accuracy than the standard controller right stick, and removes the need for auto-aim in games.

This implementation requires SDL 2.0.16 or greater; of course, game will continue to work with lower versions. It has been successfully tested with a DualShock 4 and a Switch Pro Controller in the following operating systems:
- _Ubuntu 20.04_: both controllers work flawlessly, wired and wireless (bluetooth).
- _MacOS High Sierra_: Switch Pro Controller works wired and wireless. For the DualShock4, gyro was only recognized when wired, an OS problem I suppose.
- _Windows 7_: Same results as High Sierra; gyro doesn't work wireless on DS4.

Requires manual calibration of the gyro; a menu option for this is included. Other options added to menus are a button to disable gyro momentarily, select yaw or roll for in-game turning, and sensitivity.


src/client/input/sdl.c
```diff
@@ -130,9 +131,38 @@ static cvar_t *joy_axis_triggerright_threshold;
...
+static unsigned int num_samples;
+
+static cvar_t *gyro_calibration_x;
+static cvar_t *gyro_calibration_y;
+static cvar_t *gyro_calibration_z;
+
 // Support for hot plugging of game controller
 static qboolean first_init = true;
-static int init_delay = 30;
+static int in_delay = 30;
```
0lvin, line 159: Could we use old variable naming?

src/client/input/sdl.c
```diff
@@ -1373,6 +1508,33 @@ IN_Controller_Init(qboolean notify_user)
 				IN_Haptic_Effects_Info();
 			}
 
+#if SDL_VERSION_ATLEAST(2, 0, 16)	// support for controller sensors
+
+			if ( SDL_GameControllerHasSensor(controller, SDL_SENSOR_GYRO)
+				&& !SDL_GameControllerSetSensorEnabled(controller, SDL_SENSOR_GYRO, SDL_TRUE) )
+			{
+				float gyro_data_rate = SDL_GameControllerGetSensorDataRate(controller, SDL_SENSOR_GYRO);
+#ifndef _WIN32
```
0lvin, line 1517: Could you write some comment about differences between gyro in _WIN32 platform and others?

0lvin (approved):

src/client/input/sdl.c
protocultor: Yes, old name is no longer representative of what this counter is doing now. I invented that variable on an old PR precisely to add a "delay" to gamepad initialization on hotplug (since adding them inmediately after detected led to unwanted behaviour). On the new PR, as this variable existed, I didn't want to add another "counter" for the amount of inputs needed in the "gyro calibration" phase, so I reused it and wanted to reflect that in the name, with "in_" as "input", like in IN_Init(), IN_Update(), etc.

Still, I agree the name is ugly and doesn't clarify what it's doing now, but at least expresses it's more of a "general purpose" variable.

Of course, I'll change the name; I just don't think the old one explains both uses. Is "updates_counter" appropiate for the project? Any suggestions are welcome.

src/client/input/sdl.c
protocultor: I did in the commit, but you're right, it should be more visible, especially since it's a feature that should get homogeneous between platforms in the (near?) future of SDL2. I'll add them in the code.

Seizing the opportunity, I'll be a little bit more verbose here:
SDL2 provides different gyro readings in Windows, compared to Linux and Mac for the Switch Pro Controller. Meaning, the same movement (let's say a 90° turn left) provides bigger numbers in Linux and Mac than in Windows. I have no proof, but seeing that the DualShock 4 provides the same numbers across all plaforms, and they coincide with the Pro Controller on Windows, I'd say the numbers in Linux and Mac are wrong. I'll ask on the SDL forums once yquake2 master gets stable and this can be tested by other people.
As you can see in the code, if the gyro's data rate is 200 Hz or less in anything "not Windows", it's a Pro Controller, so the "normalizing factor" is smaller. This factor is something used to achieve a "natural scale" in the final output, meaning a 90° turn left on your controller means a 90° turn left in-game, with sensitivity 1.0; having 2.0 sensi means a 180°, 0.5 is 45°... you get the idea. Messing this "normalizing factor" will not break the game or its functionality but kills the usefulness of the sensitivity. That's all.
I'll try to explain this in a shorter manner in the code comments... if I can :)

Thank you for your time to review this.

protocultor:
@0lvin can you check out the new commit?

0lvin (approved):
@protocultor Looks good as me. Thank you!

Yamagi:
And merged. Keep up the good work!

Yamagi:
A new testbuild for Windows with this PR included is here: https://deponie.yamagi.org/quake2/windows/testbuilds/quake2-20220617-c6f6a17.zip

vamshi-cell (commented):
.

PR #850, opened 2022-06-10
Infinite loop in bind menu fix

apartfromtime:
Bug introduced in #841.
Added QMF_INACTIVE.
QMF_INACTIVE menu items are skipped.


Yamagi:
Thank you very much for the quick fix!

PR #851, opened 2022-06-11
client download, adding two new cvar to have the ability to set

devnexen:
a threshold to abort the download if a certain transfer rate is too low.
Disbabled by default.


0lvin:
small typo: Dis_b_abled by default.

0lvin (approved):

Yamagi:
Thank you! :)

PR #856, opened 2022-06-25
Iterate once through menu item list

apartfromtime:
Avoids infinite loop cases.


Yamagi:
Thank you very much for implementing this :)

PR #857, opened 2022-07-05
Gyro aiming through Bluetooth for PS4 & PS5 controllers

protocultor:
This is an extension of https://github.com/yquake2/yquake2/pull/849, where gyro aiming was available wireless on Nintendo Switch controllers, but not on PS4 ones (on Windows and Mac). This commit fixes this, even enabling it for the PS5 controller.
What this does is to enable "extended input reports", which handles sensors and rumble through bluetooth.
Wanted to add this on my next PR, but that will take more work and more time.

EDIT: Second commit is something that was requested on Discord. I don't think it hurts to make some people happy...

### DIFF
--- a/src/client/input/sdl.c
+++ b/src/client/input/sdl.c
@@ -1425,6 +1425,14 @@ IN_Controller_Init(qboolean notify_user)
 
 	if (!SDL_WasInit(SDL_INIT_GAMECONTROLLER | SDL_INIT_HAPTIC))
 	{
+
+#ifdef SDL_HINT_JOYSTICK_HIDAPI_PS4_RUMBLE	// extended input reports on PS controllers (enables gyro thru bluetooth)
+		SDL_SetHint( SDL_HINT_JOYSTICK_HIDAPI_PS4_RUMBLE, "1" );
+#endif
+#ifdef SDL_HINT_JOYSTICK_HIDAPI_PS5_RUMBLE
+		SDL_SetHint( SDL_HINT_JOYSTICK_HIDAPI_PS5_RUMBLE, "1" );
+#endif
+
 		if (SDL_Init(SDL_INIT_GAMECONTROLLER | SDL_INIT_HAPTIC) == -1)
 		{
 			Com_Printf ("Couldn't init SDL joystick: %s.\n", SDL_GetError ());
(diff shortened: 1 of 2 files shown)


src/client/input/sdl.c
```diff
@@ -1425,6 +1425,14 @@ IN_Controller_Init(qboolean notify_user)
 
 	if (!SDL_WasInit(SDL_INIT_GAMECONTROLLER | SDL_INIT_HAPTIC))
 	{
+
+#ifdef SDL_HINT_JOYSTICK_HIDAPI_PS4_RUMBLE	// extended input reports on PS controllers (enables gyro thru bluetooth)
```
Yamagi, line 1429: I wonder if it's worth some option the the Makefile for enabling these hints or if they're obscure enough that it is not necessary and we can expect the very few interested users to pass the option at the `make` command line or set the hint by specifying the corresponding environment variable when starting the game.

protocultor: Can I ask the reason?
If there's an "_#ifdef_" before them, it's because I wanted to retain compatibility with SDL 2.0.10, so YQ2 can continue to be compiled on a Raspberry Pi, for example.
It's an alternative to use _SDL_VERSION_ATLEAST()_.
It must be said that these hints, when supported and enabled, provide no conflict in the system and there's no practical advantage for having them disabled. If they are disabled by default, it's because they "break DirectInput handling for applications that don't use SDL", according to SDL documentation. But in practice, DirectInput is used only in very old games, and every person using a PS4/PS5 controller with their PC uses an input remapper to emulate XInput anyway (Steam input, DS4Windows, reWASD, etc), which probably uses the same HIDAPI, with the same side effect (solved by power cycling the controller).
Having "native input" provides YQ2 with an usability advantage, especially for gyro configuration, which is always a pain in input remappers. Again, if the user doesn't use any of this, having them disabled provides no advantage.

If you want, I can replace the "_ifdefs_" with "_SDL_VERSION_ATLEAST()_", for 2.0.22 in this case.

tekgeekster: > I wonder if it's worth some option the the Makefile for enabling these hints or if they're obscure enough that it is not necessary and we can expect the very few interested users to pass the option at the `make` command line or set the hint by specifying the corresponding environment variable when starting the game.

I'm personally interested in this as someone who not only enjoys quake, but also someone who plays it exclusively on controller. SDL2 is a great controller plugin, especially for offering gyro aiming which is feel benefits quake on controller exponentially as seen in the quake remaster on ps4/5 and switch.

Yamagi: I totally missunderstood this. I didn't realize that the HINT_* defines are defined by the SDL headers. :/

Yamagi:
Thank you!

PR #862, opened 2022-07-20
R_DrawParticles2 signature slight change fixing build warning.

devnexen:


DanielGibson:
what's the warning that this fixes?
does the table not really have 768 entries?

devnexen:
because we pass a argument with 256 entries.

Yamagi:
I think this warning came with one of the last gcc updates. I'm seeing it on my Arch Linux with gcc 12.1, but not on my FreeBSD with an gcc 11.3.0.

PR #863, opened 2022-07-23
Hunk_Begin/Alloc attr malloc addition.

devnexen:


Yamagi:
Cool idea. Unfortunately my Ryzen is way too fast to see if it has a measurable effect ;)

PR #864, opened 2022-07-30
Enhanced sliders

protocultor:
Every person that has implemented a slider in Q2 knows that "_menuslider_s_" has their own internal values, which have to be "translated" to a value accepted by a certain cvar by a callback function. This is unnecesarily annoying and bloated, so this PR address this by making _menuslider_s_ to directly associate to a cvar, so their currently selected value is get from this cvar. Also, their _minvalue_ and _maxvalue_ are expressed in terms of real values this cvar can get. By default, the cvar changes by steps of 0.1, but this can be overridden by setting _menuslider_s->slidestep_, as can be seen in "mouse speed" in "options", and the "field of view" slider in the video options.

The heart of the implementation can be seen in `src/client/menu/header/qmenu.h` and `src/client/menu/qmenu.c`.
All current sliders have been updated, since the "old way" doesn't work anymore. You still can associate a callback function, but I don't see the reason for it :) . Note the amount of callback functions deleted.

This PR also adds something important to the user: the current value of the option changed is visible, beside the slider. This makes them useful because the slider without values only gives a "reference of a change" according to the slider alone, which is a weak point of reference. This has been tested in multiple resolutions, and values don't escape the screen even in 320x240.
Format of the printed value can be set with _menuslider_s->printformat_. Again, "field of view" is a good example of usage.

q2_0000 q2_0001
q2_0002 q2_0003
q2_0004


Yamagi:
That's a very good idea! Both the direct association between a cvar, its value and the slider and printing the value next to it :+1: Merged!

PR #865, opened 2022-08-01
Faster weapon switching with 'cycleweap'

protocultor:
This PR makes _cycleweap_ "count the amount of key presses" so it can navigate through its defined list quicker and pull out the desired weapon faster, instead of waiting for the appearance and animation of every weapon on its list.

e.g. you have the following bind:
`bind e "cycleweap weapon_machinegun weapon_chaingun"`
..and you are using any other weapon. By double-tapping "e", you can immediately swap to the chaingun, instead of having to wait for the machinegun to come out first.
It's the equivalent of pressing "4" and quickly after, "5". The press of "5" overrides the previous "4", so the chaingun is the weapon selected at the end.


Yamagi:
Merged. Thank you! :)

PR #866, opened 2022-08-06
Make skin icon bitmap part of the player setup menu

apartfromtime:
Adding menu bitmaps makes this possible.
Bounds check on skin box current value.


Yamagi:
Good idea. But there's something wrong: When I'm choosing the skin left to the first one (index -1 or something like that) the whole menu turns black. By me "Claymore" is the first skin -> press "left" again -> menu disappears. The same seems to happen when I'm on the last model "male" and try to switch to the right again.

apartfromtime:
The bounds check I included is causing the issue. The player config is a gnarly piece of code, it could benefit being reworked to use indexing.

Yamagi:
And finally merged. Thank you for fixing the back screen issue and sorry for the long delay.

0lvin:
Do you have issue that user can press right one more time in skin selection and menu will disappear?

protocultor:
> Do you have issue that user can press right one more time after select last element in skin selection and menu will disappear? Could you please check?
> 
> (Model selection does not have such issue.)
Yes, this PR broke the master branch.
If you press "right" one more time after the last skin, the screen goes to black. If you press "right" a couple more times, the game crashes.
@apartfromtime, like last time, this is not a "corner case" and was detected by simply using the game.

apartfromtime:
> Do you have issue that user can press right one more time after select last element in skin selection and menu will disappear? Could you please check?
> 
> (Model selection does not have such issue.)

> > Do you have issue that user can press right one more time after select last element in skin selection and menu will disappear? Could you please check?
> > (Model selection does not have such issue.)
> 
> Yes, this PR broke the master branch. If you press "right" one more time after the last skin, the screen goes to black. If you press "right" a couple more times, the game crashes. @apartfromtime, like last time, this is not a "corner case" and was detected by simply using the game.

@0lvin & @protocultor, Could you confirm this #893 fixes the issue?

This could be because I removed the guard pointers, spincontrol checks for null itemnames, but standard string functions used with qsort don't handle null stings well.
Potentially any string handling could be done with a count and not a check for null.

protocultor:
> @0lvin & @protocultor, Could you confirm this #893 fixes the issue?
> 
> This could be because I removed the guard pointers, spincontrol checks for null itemnames, but standard string functions used with qsort don't handle null stings well. Potentially any string handling could be done with a count and not a check for null.

Yup, that fixes it, the game no longer crashes.

PR #868, opened 2022-08-10
Cmake build fix2

devnexen:


Makefile
```diff
@@ -168,7 +168,7 @@ endif
 #   and the game code requires it.
 #  -fvisibility=hidden to keep symbols hidden. This is
 #   mostly best practice and not really necessary.
-override CFLAGS += -std=gnu99 -fno-strict-aliasing -fwrapv -fvisibility=hidden
+override CFLAGS += -std=gnu11 -fno-strict-aliasing -fwrapv -fvisibility=hidden
```
Yamagi, line 171: Why the switch to C11? Okay, most up to date compilers have C11 support these days, but there are some users who still play the game on old hardware and maybe old Linux distros. I don't want to break them without good reason.

src/common/header/shared.h
```diff
@@ -82,12 +93,14 @@ typedef unsigned char byte;
 	// must be used as prefix (YQ2_ATTR_NORETURN void bla();)!
 	#define YQ2_ATTR_NORETURN       __declspec(noreturn)
    	#define YQ2_ATTR_MALLOC         __declspec(restrict)
+	#define YQ2_ATTR_INLINE         inline
```
Yamagi, line 96: Should be `__forceinline`

Makefile
devnexen: ok I m going to roll it back.

Yamagi:
@DanielGibson, you're okay with this?

src/common/header/shared.h
```diff
@@ -59,12 +59,23 @@ typedef unsigned char byte;
 	#define YQ2_ALIGNAS_TYPE(TYPE)  _Alignas(TYPE)
 	// must be used as prefix (YQ2_ATTR_NORETURN void bla();)!
 	#define YQ2_ATTR_NORETURN       _Noreturn
+#if defined(__GNUC__)
+	#define YQ2_ATTR_MALLOC         __attribute__ ((__malloc__))
+	#define YQ2_ATTR_INLINE         __attribute__((always_inline))
+#elif defined(_MSC_VER)
+   	#define YQ2_ATTR_MALLOC         __declspec(restrict)
```
DanielGibson, line 66: the __forceinline case is missing

DanielGibson:
Apart from the one missing `__forceinline` case I just commented:
I'm not sure about the C standard version..
On the one hand forcing C11 would suck for people on old systems, on the other hand using C11 support if available (for `_Alignas()` etc) would be nice - *especially* in CMake for MSVC (where we otherwise don't have proper support for `YQ2_ALIGNAS_TYPE()`)..

I think CMake allows suggesting a standard version without enforcing it?

DanielGibson:
By the way, in CMake you can set the language standard globally (instead of per project), by adding the following lines somewhere (probably ideally near the top?) in the CMakeLists.txt:

```
set(CMAKE_C_STANDARD 11)
set(CMAKE_C_STANDARD_REQUIRED OFF) # this makes it optional
```
(irrelevant for YQ2, but the same exists for C++ with `CMAKE_CXX_...`).
It requires CMake 3.1, so we'd have to modify the `cmake_minimum_required(VERSION 3.0 FATAL_ERROR)` accordingly (I think it's reasonable to require CMake 3.1, the current version is 3.24)

See also https://cmake.org/cmake/help/latest/manual/cmake-variables.7.html#id6 and https://cmake.org/cmake/help/latest/variable/CMAKE_C_STANDARD.html and https://cmake.org/cmake/help/latest/variable/CMAKE_C_STANDARD_REQUIRED.html

src/common/header/shared.h
```diff
@@ -59,12 +59,23 @@ typedef unsigned char byte;
 	#define YQ2_ALIGNAS_TYPE(TYPE)  _Alignas(TYPE)
 	// must be used as prefix (YQ2_ATTR_NORETURN void bla();)!
 	#define YQ2_ATTR_NORETURN       _Noreturn
+#if defined(__GNUC__)
```
DanielGibson, line 62: and while I'm nitpicking, I'd indent this `#if` (and the corresponding `#elif`, `#else` and `#endif`) a by two spaces, like I did below in the `_MSC_VER` -> `_WIN64` case, so it's easier to tell which `#else` (etc) belongs to which `#if`

Yamagi:
Thank you :)

Issue #782, opened 2022-01-18
Audio improvements


jan-kleks:
I have recently found this video on a channel which specializes in demonstrating 3D audio in games -- "Yamagi Quake II: Broken OpenAL Soft doppler effect": https://www.youtube.com/watch?v=59t38Ju0qYk

Just in case, is this something that you are aware of (and it has been completely fixed)?

Yamagi:
We've disabled the doppler effect some time ago. I can still be enabled with `s_doppler` or something like that. It's not worth fixing it, because it wasn't in Vanilla Quake II and the sound system is very limited anyways. Sound in Quake II is always two dimensional, the samples are 22 kHz, etc.

jan-kleks:
@Yamagi I see, but according to this video the effect seems to be working fine in Q2Pro, wouldn't its source code be helpful here?

maraakate:
Q2Pro is entirely different from regular Q2 engines, so the answer is "maybe".

Yamagi:
YQ2s OpenAL backend is based on q2pro, so that should be possible. But... I just had an albeit short look at q2pro and there's no mention of a doppler effect inside their *al.c* and I wasn't able to find something which looks like doppler effect calculations in the frontend. Are we sure that q2pro has a nativ doppler effect?

Enokilis:
Broken? I thought it was by design.

https://github.com/yquake2/yquake2/blob/8358e386fcdc76d70508239afc7b771795ee1017/src/client/sound/openal.c#L888

DanielGibson:
Maybe we could feed the value of `s->doppler` directly into `qalDopplerFactor()` to make it configurable (with special case `0.0` to disable it)?

jan-kleks:
> And yes, q2pro doesn't seem to have a doppler effect, it never calls (q)alDopplerFactor() or (q)alDopplerVelocity(), and it doesn't set listener's `AL_VELOCITY` either.

Hmm... I wonder what the creator of the video comparison I linked to in the first post had had in mind then when trying to compare the "doppler effect" in Yamagi vs Q2Pro.

Anyway, I have just found out there is a custom version of Q2RTX by SkacikPL with "Open AL effects extensions ported from Yamagi Quake 2" as well as **additional audio enhancements**:

- reverb
- occlusion effects
- footsteps sounds for all ground enemies (which were added to make an even better use of additional effects)
[...]

PR #869, opened 2022-08-12
Add step sound and reverberation

0lvin:
* Cherrypick of commits from https://github.com/SkacikPL/Q2RTX  #782
* Add monster spawn command
* Add monster step sound
* Add reverberation effect
* Sync function definitions in game.so


0lvin:
@Yamagi @DanielGibson Do we need to export all function from game.so or we can cleanup list little bit? And remove all not related to monster fields functions.

Yamagi:
> Do we need to export all function from game.so or we can cleanup list little bit? And remove all not related to monster fields functions.

It's been ten years or something since the last time I looked into the depth into the savegame system. As far as I remember it's enough to export all functions which are directly or indirectly referenced through function pointers by `edict_t` and `mmove_t` instances. The problem was to figure out what the function were. The tool I used to generate the tables (https://github.com/yquake2/extractfunctions) was taken from the RTCW SDK and took the easy way to include all functions.

On a very first look this change looks reasonable, better sound effects are an often requested feature. But the standard settings should preserve the original 'looks and feel', all added effects must be optional. What does this do to the server <-> game API and the network protocol? Does it hack the new features in without breaking things? Will the game.dll still work with original binaries from 1998 without spilling errors into the console and the other way round, will old game.dll still work with YQ2? Can YQ2 still connect to other servers? And is savegame backward compatibility preserved?

Reviewing this will take some time. I don't think that @DanielGibson and me can start before September. :/

0lvin:
> On a very first look this change looks reasonable, better sound effects are an often requested feature.

By default all effects are disabled. Q2RTX fork contains voice input with apply effect, and i am not sure that such echo makes sense. So i have skipped such logic with additional code for add 0.1 second silence for each track, its known issue for sound that some of tracks has loud begging. 

>  Reviewing this will take some time

Its fully fine, maybe i will found something additional.

>  The tool I used to generate the tables (https://github.com/yquake2/extractfunctions) was taken from the RTCW SDK and took the easy way to include all functions.

Thank you. I will look, may be can hide some of functions than are not related to models logic. (Like get address by name as example.)

0lvin:
@Yamagi @DanielGibson I have created separated PR for new commands that i have used for check steps #871

src/game/header/local.h
```diff
@@ -1095,4 +1115,14 @@ struct edict_s
 	monsterinfo_t monsterinfo;
 };
 
+/*
+ * Uncomment for check that exported functions declarations are same as in
+ * implementation. (-Wmissing-prototypes )
+ *
+ */
+#if 0
```
Yamagi, line 1123: This is a cool idea. I wonder I never had it ;)

Yamagi (commented):
I've added some comment. Overall this looks good and is way less invasive than I though. 

* The game API isn't changed and I'm about 90% sure that it doesn't break the network protocol. Adding new sound entities should be fine.
* I think we need to bump the savegame version. Otherwise old DLLs may end up with disconnected function pointers to the newly added stepsound functions or something like that.

src/game/header/game.h
```diff
@@ -34,7 +34,7 @@
  * !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
  */
 
-#define GAME_API_VERSION 3
+#define GAME_API_VERSION 4
```
Yamagi, line 37: Not the game api version, the savegame version somewhere in savegame.c. :) Or am I missunderstanding the change?

I thought: Bump the savegame version so that older client cannot load new savegames and add a simple fallback case so that newer clients can still load old savegames. There's already code for that, we've done it several times before.

doc/040_cvarlist.md
```diff
@@ -255,6 +266,15 @@ Set `0` by default.
 
 * **s_underwater**: Dampen sounds if submerged. Enabled by default.
 
+* **s_occlusion_strength**: If set bigger than `0` sound occusion effects
```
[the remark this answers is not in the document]
0lvin: Thank you, fixed.

src/client/sound/openal.c
```diff
@@ -238,6 +246,339 @@ AL_DeleteSfx(sfx_t *s)
...
+	"Smallwater Room"
+};
+
+/*
+ * Update reverb setting without apply
+ */
+static void
+AL_SetReverb(int reverb_effect)
+{
+	EFXEAXREVERBPROPERTIES reverb = ReverbPresets[reverb_effect];
```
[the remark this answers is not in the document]
0lvin: Thank you, fixed.

src/game/header/local.h
0lvin: Thank you!

src/client/menu/menu.c
```diff
@@ -5656,6 +5656,27 @@ PlayerConfig_MenuInit(void)
...
+
+    cl_start_frame = Cvar_Get("cl_model_preview_start", "-1", CVAR_ARCHIVE);
+    cl_end_frame = Cvar_Get("cl_model_preview_end", "-1", CVAR_ARCHIVE);
+    startFrame = cl_start_frame->value;
+    endFrame = cl_end_frame->value;
+
+    if (startFrame >= 0 && endFrame > startFrame)
+    {
+        /* salute male 84..94 frame */
+        entity->frame = (curTime / 100) % (endFrame - startFrame) + startFrame;
```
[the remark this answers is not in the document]
0lvin: All player models(`cyborg`, `female`, `male`) have same frame values for salute. I have made to show it by default. 

I had looked to show all animation available for the model, but model structure is unavailable for menus (opaque type) https://github.com/yquake2/yquake2/blob/ed918cf/src/client/vid/header/ref.h#L57. And hard coded  values are best solution for now :-(

src/game/header/game.h
0lvin: Sorry, looked to wrong place. Now should be better :-)

Yamagi:
Thank you for the updates. It looks good now. Very good work! :) I'll do a (partially) play-through with these changes. If nothing comes up I think we can merge it.

Yamagi:
I've played through the whole game and nothing was obviously broken. I'm merging this. Let's see if anything comes up.

It would be nice to get the step sounds into the xatrix and rogue addons. :)

PR #873, opened 2022-08-15
Remove broken pusher delta yaw manipulation

skullernet:
This didn't work correctly for multiple reasons:

1. `deltayaw` was wrongly initialized for the pusher itself, rather than for pushed client.
2. `delta_angles[YAW]` is a short, adding plain `amove[YAW]` to it is wrong.

To support yaw angle rotation properly, delta_angles must be interpolated on the client. But this is hardly practical as it would introduce other bugs. Thus, simply remove delta yaw manipulation code altogether.

Fixes infamous Q2 bug when player standing on a blocked lift gets turned into wrong direction.

I'll also send PRs for mission packs if this is accepted.

### DIFF
--- a/src/game/g_phys.c
+++ b/src/game/g_phys.c
@@ -579,7 +579,6 @@ typedef struct
 	edict_t *ent;
 	vec3_t origin;
 	vec3_t angles;
-	float deltayaw;
 } pushed_t;
 
 pushed_t pushed[MAX_EDICTS], *pushed_p;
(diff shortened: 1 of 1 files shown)


apartfromtime:
Nice find. I think i've seen this in a speed-run.

Yamagi:
Thank you. That's a very good find :) I'll merge the addon PRs as soons as you send them.

PR #882, opened 2022-08-24
Added brackets to mod folder names in mods menu

TedThompson:
Folder names are fixed length, truncated if too long, and wrapped in brackets to designate them as navigable using left/right arrows, as is done in the video options menu.

I did this more or less to see if I could, so its fine if you find it unacceptable.


TedThompson:
Sorry for the mess of commits, I'm still new to this "pullrequest" stuff.

apartfromtime:
These too should be detected on initialization, which currently are only detected on game start.

TedThompson:
> These too should be detected on initialization, which currently are only detected on game start.

I have no idea what you mean. 

I build an alternate array of display names the same time the modnames array is constructed.

The array of display names have brackets appended to them. 

Then I edited to menu routine to use my new array instead of the array containing folder names. 

If you are refering to when the function to collect game names is called, thats beyond the intended scope of this contribution.

apartfromtime:
I meant as a note in general to the source code maintainers, not your source changes in particular. You're right it is beyond the scope of this contribution.

.gitignore
```diff
@@ -3,3 +3,5 @@
 *.mk
 *.user
 *.d
+src/yquake2.pnproj
+src/yquake2.pnps
```
Yamagi, line 7: These don't belong in the upstream repo. Editor specific stuff should be in the local ~/.gitignore, otherwise we would need to add support for all editoris in the world...

Yamagi:
Can you squash the changes into one commit? When you force push you branch after that Github will pick up the changes and autoupdate the pull request

TedThompson (commented):
removed offending lines of .gitignore

src/client/menu/menu.c
```diff
@@ -2915,9 +2915,37 @@ Mods_MenuInit(void)
...
+	// create array of bracketed display names from folder names - TG626
+	char **displaynames=malloc(sizeof(*displaynames) * nummods);
+	
+	for (int i=0; i < nummods; i++)
+	{
+		strcpy(modname, "[");
+		if (strlen(modnames[i]) < 16)
+		{
+			strcat(modname, modnames[i]);
+			for (int j=0; j < 15 - strlen(modnames[i]); j++)
```
DanielGibson, line 2931: this loop could be a single call to memset() (with `' '` as char)

src/client/menu/menu.c
```diff
@@ -2915,9 +2915,37 @@ Mods_MenuInit(void)
...
+			{
+				strcat(modname, " ");
+			}
+		}
+		else
+		{
+			strncat(modname, modnames[i], 12);
```
DanielGibson, line 2938: strncat() is dangerous - it doesn't guarantee \0-termination!
So after the char from `modnames[i][11] it won't insert a \0 character, and the call to strcat() below might start there.

it's possible that all this would be easier with snprintf() with the correct arguments (you can specify how many chars of a '%s' argument should be copied).

src/client/menu/menu.c
TedThompson: I am not familiar with that - I also don't see the "harm" in how it's done.

TedThompson: Oh wait... it's the code BELOW this comment, not above...  Still finding my footing here.

Now I get what you mean (2933 `strcat(modname, " ");` ), tho I still see no harm in how it's donw.

src/client/menu/menu.c
```diff
@@ -2915,9 +2915,37 @@ Mods_MenuInit(void)
...
 
     Mods_NamesInit();
+	
+    // create array of bracketed display names from folder names - TG626
+    char **displaynames=malloc(sizeof(*displaynames) * nummods);
 
+    for (int i=0; i < nummods; i++)
+    {
+        strcpy(modname, "[");
+	if (strlen(modnames[i]) < 16)
```
DanielGibson, line 2928: try something like
```c
if(strlen(modnames[i]) < 16) {
    snprintf(modname, sizeof(modname), "[%-16s]", modnames[i]);
} else {
    snprintf(modname, sizeof(modname), "[%.12s...]", modnames[i]);
}
```
(I didn't really test this, maybe 16 and 12 aren't the right numbers for the right width, but something like this should work)

src/client/menu/menu.c
TedThompson: It doesn't cause any crashes - short names, 60 character long names were both tested, tho I see your point.

src/client/menu/menu.c
TedThompson: later, all use strcat, and therefore are null terminated.

DanielGibson: these two snprintf() calls (and one strlen()) are all you need, no strcat(), no strcpy()

TedThompson: What I have is a working system, I see no clear advantage to _changing_ it to your method, more elegant tho it may be. 
If and when this gets committed to the master branch, feel free to implement your version - I won't be offended.

DanielGibson: hmm ok, I misremembered what was wrong with strncat() (it *does* null-terminate, but use the number of chars in src as argument instead of the buffer size, which is error-prone but not a problem in this case. I confused ith with **strncpy()** which does *not* guarantee null-termination if src has more than n chars), so yes, it should work, though it's a lot of code for what could be a one-liner :)

TedThompson:
> 

Done (I think?)

Yamagi:
@DanielGibson, are you okay with this one?

DanielGibson:
> @DanielGibson, are you okay with this one?

Generally the code looks ok, but the indentation is wonky (mixing tabs and spaces), this is what it looks like here (with 4 spaces tabs):
image
The surrounding code seems to use only spaces for indentation, so that's what this change should do.

Additional small nitpick: `displaynames[i] = malloc(strlen(modname) + 1); strcpy(displaynames[i], modname);` could be replaced with `displaynames[i] = strdup(modname);`.

**Note for us:** When merging this we should remember to *squash it all into one commit*.

src/client/menu/menu.c
```diff
@@ -2437,572 +2437,572 @@ static const char *idcredits[] = {
...
-	    }
-        }
-        else
-        {
-            strncat(modname, modnames[i], 12);
-            strcat(modname, "...");
-        }	
-        strcat(modname, "]");
+	// create array of bracketed display names from folder names - TG626
+	char **displaynames=malloc(sizeof(*displaynames) * nummods);
```
DanielGibson, line 2923: this should've been (nummods+1), otherwise `displaynames[nummods] = 0;` below acccesses invalid memory
(just a heads-up, I already fixed this)

DanielGibson:
I manually squashed, fixed (the `(nummods+1)` thing) and merged this (without reformatting the whole file, because that makes `git annotate` aka `git blame` useless).
Thanks a lot for your contribution! :)

PR #883, opened 2022-08-25
Improved controller thumbsticks precision + Flick Stick

protocultor:
I know all of you are very busy right now, but doesn't matter how much time you take to review this, I wanted to do this before the 25th anniversary of Quake II :)

The present PR rewrites joystick reading and handling to introduce some new features that improve thumbsticks' precision and customization.

First, what's been removed:

1. Assigning a function to each stick axis individually.
2. Classic "threshold" deadzones by axis, so no more axial deadzone either.
3. Having "joystick_up" as input for IN_Move(), along with its sensitivity. No gamepad standard uses it, it would be useful only on a ladder or underwater, and only people who have looked at the source code would've known about it, since the name of the function was needed to be assigned to an axis (not present in documentation).

**Deleted cvars: 14.**

What's introduced:

1. A cvar to quickly assign stick layout from a list of classic predetermined ones (plus a new surprise); also available in menu.
2. Circular, inner deadzone, for each stick.
3. Sloped axial deadzone (also called "angular deadzone") replacing the previous axial one, configured separately from the inner deadzone. This maintains the "snap-to-axis" feeling of the previous one, the facility to mark perfectly horizontal or vertical movements, but without sacrificing space of diagonal movement.
Having separated deadzones allows for a more custom experience, since people that prefer a larger inner deadzone and get rid of the axial one can do that now, or on the contrary, eliminate the inner deadzone but require a bigger "snap-to-axis" feeling can do that too.
Deadzones

4. "Expo" is now different for each stick, and it's applied to the "magnitude" of the stick, meaning it's measured on the distance from the center of the stick. Currently, it's applied separately on each axis, which wouldn't make possible to reach the full curve on diagonals, since diagonals don't reach the whole range of either axis.
5. Among the stick layouts, a new one is introduced: [Flick Stick](https://en.wikipedia.org/wiki/Flick_Stick), a state-of-the-art control method designed by @JibbSmart, currently featured in games like _CS:GO_ and _Fortnite_. In this layout, the right stick functions as the camera of a racing game, capable of covering horizontally your surroundings in 360º. Another way to describe it is that makes the game operate as a 2D twin-stick shooter in 3D. Its creator explains it better: https://www.youtube.com/watch?v=C5L_Px3dFtE
To look up and down, gyro aiming is required, so at least to set it up with the menus, a gyro-capable controller is required.

**Added cvars: 9.**

Finally, I wanted to ask what is needed to be included in the Yamagi credits. I'll continue to contribute to the project, regardless.

#### Bibliography

- Introduction to deadzones and its handling
http://joshsutphin.com/2013/04/12/doing-thumbstick-dead-zones-right.html
- Advanced deadzone handling (from this, implemented what the author refers to as a "hybrid" solution)
https://github.com/Minimuino/thumbstick-deadzones
- Flick Stick and how to implement
http://gyrowiki.jibbsmart.com/blog:good-gyro-controls-part-2:the-flick-stick

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -476,6 +476,9 @@ Set `0` by default.
   - `2`: *Legacy*, left moves forward/backward and turns, right strafes
     and looks up/down
   - `3`: *Legacy Southpaw*, inverted sticks version of previous one
+  - `4`: *Flick Stick*, left stick moves, right checks your surroundings
+    in 360º, gyro required for looking up/down
+  - `5`: *Flick Stick Southpaw*, swapped sticks version of last one
 
 * **joy_left_deadzone** / **joy_right_deadzone**: Inner, circular
   deadzone for each stick, where inputs below this radius will be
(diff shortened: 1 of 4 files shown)


Yamagi:
Let me first say that you are doing a good job. All your pull requests so far have been very good. Basically, you implemented the joystick input more or less on your own. That's definitely enough to get you added to the credits. Just send a PR with the entry in the credits list (https://github.com/yquake2/yquake2/blob/master/src/client/menu/menu.c#L2418) and I'll merge it.

For joystick input, my personal problem is that I never really played Quake II with gamepads or worse joysticks. I can say that this PR feels good with a PS3 gamepad. But I don't know if it's an improvement for long-time gamepad players. I'll have to trust you (and any other comments) on that.

We'll look over the code tomorrow and if we don't find any obvious bugs, I'll merge it.

Yamagi:
Of course the code looks good :)

PR #893, opened 2022-09-09
Include guard pointers

apartfromtime:
Spincontrol checks for null itemnames, which could cause undefined behaviour (menu disappears or crashes) in some instances.

### DIFF
--- a/src/client/menu/menu.c
+++ b/src/client/menu/menu.c
@@ -5228,7 +5228,6 @@ PlayerDirectoryList(void)
     char* findname = "players/*";
     char** list = 0;
     int num = 0;
-    int n = 0;
 
     // get a list of "players" subdirectories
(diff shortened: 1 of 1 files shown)


Yamagi:
Thank you :)

protocultor:
> Thank you :)

I know it wasn't what you meant, but just in case, this needed to be merged, to solve the problem introduced in https://github.com/yquake2/yquake2/pull/866 when selecting player skin.

apartfromtime:
This doesn't appear to have been merged and as @protocultor mentioned a problem was introduced in #866.

Yamagi:
This was a manual merge and I didn't push it last night. Here it is: f63b5914f6b14e4e4f1f4bda648b7c53eb6fbbb4

PR #896, opened 2022-09-19
Options to invert gyro axes

### LINKED ISSUE #879: Can't disable joystick/gamepad in Windows?
I can't seem to disable the joystick in YQuake.

I've created a shortcut whose target is "[quake 2 directory]\yquake2.exe" +set in_initjoy 0", which Google tells me is the way to do it in vanilla Quake 2. But still, Quake recognizes joystick inputs - no good.

Would really rather not have to physically unplug my joystick each time I want to play Quake 2, so... is there any command to completely disable joystick/gamepad input?

protocultor:
This adds a couple of spin controls to invert yaw and pitch in the gyro options.
Now sliders are able to use the "absolute value" (ignore negative sign) of its associated cvar, if there's a need for it. This is because the "reversal" of the gyro axes is achieved by converting their sensitivities to a negative value; I believe it's better than adding a new cvar only for this.

Additionally, I resurrected the "in_initjoy" cvar from Quake II 3.21. Fixes #879 
Finally, I added myself to the credits. Thank you for this!

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -471,6 +471,10 @@ Set `0` by default.
 
 ## Game Controller
 
+* **in_initjoy**: Toggles initialization of game controller. Default is
+  `1`, which enables gamepad usage; `0` disables its detection at
+  startup. Can only be set from command line.
+
 * **in_sdlbackbutton**: Defines which button is used in the gamepad or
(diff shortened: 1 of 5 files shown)


Yamagi:
> Finally, I added myself to the credits. Thank you for this!
Thank you for your contributions!

Issue #894, opened 2022-09-17
Buffer overflow crash in the collission model


Yamagi:
While testing #883 I stumbled over this buffer overflow in the collision model:

```
==621882==ERROR: AddressSanitizer: global-buffer-overflow on address 0x55da42f364df at pc 0x55da42743acb bp 0x7ffd9d2bdc70 sp 0x7ffd9d2bdc60
READ of size 1 at 0x55da42f364df thread T0
    #0 0x55da42743aca in PF_inPHS src/server/sv_game.c:329
    #1 0x7f21edd74f8a in FindTarget src/game/g_ai.c:671
    #2 0x7f21edd6ec24 in ai_walk src/game/g_ai.c:193
    #3 0x7f21ede2f6a7 in M_MoveFrame src/game/g_monster.c:581
    #4 0x7f21ede2fa59 in monster_think src/game/g_monster.c:604
    #5 0x7f21ede37a25 in SV_RunThink src/game/g_phys.c:138
    #6 0x7f21ede4f3de in SV_Physics_Step src/game/g_phys.c:1268
    #7 0x7f21ede4f660 in G_RunEntity src/game/g_phys.c:1299
    #8 0x7f21eddf7118 in G_RunFrame src/game/g_main.c:479
    #9 0x55da4274b578 in SV_RunGameFrame src/server/sv_main.c:354
    #10 0x55da4274bed2 in SV_Frame src/server/sv_main.c:425
    #11 0x55da42653fad in Qcommon_Frame src/common/frame.c:672
    #12 0x55da42651f20 in Qcommon_Mainloop src/common/frame.c:202
    #13 0x55da42652cee in Qcommon_Init src/common/frame.c:417
    #14 0x55da427695cd in main src/backends/unix/main.c:145
    #15 0x7f2215a2928f  (/usr/lib/libc.so.6+0x2928f)
    #16 0x7f2215a29349 in __libc_start_main (/usr/lib/libc.so.6+0x29349)
    #17 0x55da42479064 in _start (/home/yamagi/temp/yquake2/release/quake2+0x3c0064)

0x55da42f364df is located 31 bytes to the right of global variable 'pvsrow' defined in 'src/common/collision.c:72:39' (0x55da42f344c0) of size 8192
0x55da42f364df is located 1 bytes to the left of global variable 'phsrow' defined in 'src/common/collision.c:73:6' (0x55da42f364e0) of size 8192
[...]

PR #899, opened 2022-09-21
fix attempt for #894

devnexen:


DanielGibson:
Could you explain how this change relates to the issue?
It's not really obvious from looking at the change ;)

devnexen:
Sure. Well I could not reproduce however those global variables are affected [here](https://github.com/yquake2/yquake2/blob/2fae1662c2edcb8e2e2cd4739be57339bedd68d0/src/common/collision.c#L1894) I artificially reproduced having too large number :)

DanielGibson:
But the code you changed is executed at **level load**, and Yamagis error was in **base1**, so we can rule out invalid map data (that would trigger your error), except maybe if he had data corruption either on his hard-disk or in RAM.

What I think really happened in #894:
It says `0x55da42f364df is located 1 bytes to the left of global variable 'phsrow'` - and `mask` is from `CM_ClusterPHS()` which always returns `phsrow`.
So the index used to access that array ways `-1` ("11 bytes **to the left**").
So `cluster = CM_LeafCluster(leafnum);` is -1. That can happen, it probably means that the cluster is invalid or non-existant or something like that.

Other code explicitly handles `CM_LeafCluster()` returning -1, for example the code you linked (`CM_ClusterPHS()` or `CM_ClusterPVS()` which then memsets the returned data to 0) or this code here: https://github.com/yquake2/yquake2/blob/2fae1662c2edcb8e2e2cd4739be57339bedd68d0/src/server/sv_world.c#L327-L330 (`clusters[i]` is set in line 290).

Anyway, the correct fix for the issue probably is to explicitly handle `cluster < 0` in `PF_inPHS()` and `PF_inPVS()`, to avoid a negative cluster-number from used as an index into the `mask` array.

DanielGibson:
I pushed a fix: https://github.com/yquake2/yquake2/commit/4d6bdcc8530b1da3221e51a631e3e428faa88978

devnexen:
ah right thx for the hint

PR #900, opened 2022-09-23
Quicksaves can be deleted from menu

### LINKED ISSUE #892: Quicksaves cannot be deleted from menu
Quicksaves cannot be deleted from the 'Load' and 'Save' menus using the delete key as autosaves and manual saves can. This is due to the pattern of numbered directory names (save1, save2, etc.) on which the delete function is otherwise predicated not being followed for quicksaves (quick).

I have resolved this myself by adding a check `if (strstr (item->name, "QUICKSAVE") != NULL)`, but there may be more elegant ways to go about it.

protocultor:
Closes #892 
Also, avoids asking for deletion in "\<empty\>" savegames.


Yamagi:
Thank you!

Issue #903, opened 2022-09-24
[Original bug] Items that spawn underwater at level start play splash sound


BjossiAlfreds:
Title says it all. Examples of this is Comm Center (base3) and Swamp (xswamp in Reckoning). In base3 the splash sound you hear is the grenades next to the dead body, and in xswamp it is the items in the water like the armor shards.

I will provide a PR that addresses this.

PR #904, opened 2022-09-24
Fix items already in water at level start playing splash sound

BjossiAlfreds:
This PR addresses issue https://github.com/yquake2/yquake2/issues/903

The check for level.framenum > 3 ensures items fully settle their physics state before these sounds start playing.


maraakate:
Nice one.

Yamagi:
Indeed

PR #908, opened 2022-09-25
Shared model load

0lvin:
Reuse models loads code


src/common/clientserver.c
```diff
@@ -105,6 +105,30 @@ Com_VPrintf(int print_level, const char *fmt, va_list argptr)
...
+					msgLen --;
+					memmove(msg, msg + 1, msgLen);
+					break;
+				// white
+				default:
+					Sys_ConsoleOutput("\033[40m\033[37;1m");
+					break;
+			}
+		}
+
```
Yamagi, line 131: Where are these colors used anyways? :) Nevertheless, I would just leave this alone. Terminal output is hard if one starts to use special features. This misses a few things. At least:

* Linux isn't the only system with colored terminals.
* One cannot assume that a terminal supports colors. Even if most modern terminals do, some do not. Terminal combined with screen readers are often monochrome. Braille lines are terminals with next to no advanced features. And so on. This must be combined with termcap / terminfo checks to make sure, that the terminal supports colored output.
* The control sequences must not be emitted if the terminal isn't interactive. Otherwise they may clutter redirection into a file and things like that.

src/client/refresh/gl1/gl1_model.c
```diff
@@ -166,14 +164,66 @@ Mod_Init(void)
...
+static void
+Mod_SP2Fixup(model_t *mod, const dsprite_t *sprout)
+{
+	mod->type = mod_sprite;
+
+	if (sprout)
+	{
+		int i;
+
+		/* byte swap everything */
```
DanielGibson, line 210: you're not byte-swapping anything here :-p

DanielGibson:
I wonder if this fixup-stuff can be centralized as well, with a few additional arguments for the `Mod_Load*()` function.
For example, for MD2 you would probably only have to pass `mod->skins`, `mod->mins`, `mod->maxs` and a function pointer for `R_LoadImage()` (to set the skins)? (You could also pass `mod->type`, but setting that in `Mod_ForName()` doesn't hurt)

src/client/refresh/gl1/gl1_model.c
0lvin: Yep, my miss....

src/common/clientserver.c
0lvin: I have only Linux, so I suppose it is useful for Linux only. 

I will move it to separate PR if i have some more technically correct solution.

0lvin:
Thank you!

> For example, for MD2 you would probably only have to pass mod->skins, mod->mins, mod->maxs and a function pointer for R_LoadImage() (to set the skins)? (You could also pass mod->type, but setting that in Mod_ForName() doesn't hurt)

Main drawback of function as i see that image type is different for all renders. so it will be some (*image_t) -> (void*) -> (*image_t). I have idea to use for image only image index as result we can reuse all image load code.

I try some ideas  :-)

DanielGibson:
they're all `struct image_s` (for compatibility with RegisterSkin() and DrawFindPic() that return an image to the client), so you can use `struct image_s**` as a type for the passed skins array

0lvin:
@DanielGibson @Yamagi Could you please review?

DanielGibson:
Looks good to me, but we'll have to test it before it can be merged :)

Yamagi:
I've played through the first 4 units with this applied. No problems so far. Let me finish the playthrough before I merge it, but  I think it's fine.

Yamagi:
I played another two units with the updated version. Seems to work fine. I'm merging this, thank you for your continued contributions! I'll have a look at the other PRs at the weekend.

0lvin:
Thank you, i will rebase my patches for shared image load and create PR. I have found that most of code can be shared :-).

0lvin:
I have  created PR for shared image load https://github.com/yquake2/yquake2/pull/924

PR #909, opened 2022-09-30
Prevent dead bodies from obstructing elevators

BjossiAlfreds:
This PR addresses the issue where entities can phase through solid geometry. Players mostly experience this as dead bodies stopping elevators and falling through them. There was a previous attempt to fix this in the code but it didn't work in practice.

The general problem is that startsolid traces treat different obstructing volumes as continuous solidity, even if the volumes have different content types. So for example when the player steps on a dead body in an elevator, now the player bbox and the elevator are treated as a continuous volume that startsolid tries to trace through looking for non-solid space.

A good solution is not possible without game API changes, but a simple work-around is to make a 2nd trace attempt if the trace is startsolid. This 2nd attempt will only collide with BSP. This means that when entities are obstructed, they will always collide with BSP (as long as they weren't inside BSP already). For the dead body example, this means the 2nd trace will make the elevator floor stop the dead body from falling.

I also made a small change in SV_TestEntityPosition. SVF_DEADMONSTER entities (dead bodies mostly) will use MASK_SOLID for the pos test. Since other entities freely pass through dead body bboxes, I feel it makes for a smoother experience to make them only collide with BSP while being pushmoved.

### DIFF
--- a/src/game/g_phys.c
+++ b/src/game/g_phys.c
@@ -54,15 +54,18 @@
 edict_t *
 SV_TestEntityPosition(edict_t *ent)
 {
+	trace_t trace;
+	int mask;
+
 	if (!ent)
 	{
 		return NULL;
 	}
 
-	trace_t trace;
-	int mask;
-
-	if (ent->clipmask)
+	/* dead bodies are supposed to not be solid so lets
+	   ensure they only collide with BSP during pushmoves
+	*/
+	if (ent->clipmask && !(ent->svflags & SVF_DEADMONSTER))
 	{
 		mask = ent->clipmask;
 	}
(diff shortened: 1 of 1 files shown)


Yamagi:
Note for me: This has some potential for unwanted side effects.

That said, it's still good work and better than my attempt (or did I merge it fro somewhere? I have forgotten.) Let's merge it and see if someone finds the one case were it breaks.

BjossiAlfreds:
The fix attempt was only in baseq2, not rogue or xatrix. I wasn't even sure what the extra code in PushEntity was trying to do, XOR'ing the deadsolid clip flag I mean.

But yes, Quake 2's collision system is... quirky, so I'm always on edge when I do stuff like this. This is not new code actually, I've used it for over a year on my side and a friend did an entire run without issues so I trust it for the most part. If there are any unwanted side effects, they will likely be discovered down the road thanks to a larger player base. If it comes to that I'll be happy to investigate and see what's going on.

PR #912, opened 2022-10-08
fix some unitialized vars and clarify signedness on a handful of bits…

devnexen:
…hift ops.


DanielGibson:
looks good to me, thanks!

PR #914, opened 2022-10-09
Mods ordered by priority in menu

### LINKED ISSUE #891: Player skins not ordered alphabetically
In vanilla Q2, the player skins in the menu are ordered alphabetically by name. For a while now, the order in Yamagi Q2 has been random and not predicated on any discernible criteria (model name or even file size of the skins). 

I've encountered this on Linux and FreeBSD (i.e. with case-sensitive filesystems), cannot confirm whether it is also the case on Mac or Windows. 
  
Since there is no other obvious (to me) cause in the code, I suspect the problem may be that the list is non-alphabetically ordered when retrieved from the filesystem. While not critical, for aesthetic reasons, it would be nice if this could be fixed.

protocultor:
With this change, mods are presented in the following order in its menu:

baseq2 (base game), always first
xatrix (first expansion)
rogue (second expansion)
ctf (official mod by id)
the rest, ordered alphabetically

I think it improves their presentation, since "xatrix" was always presented after "rogue", or "action" before even "baseq2".

Additionally, player skins in their menu are ordered alphabetically in *nix systems, closes #891 .

### DIFF
--- a/src/client/menu/menu.c
+++ b/src/client/menu/menu.c
@@ -5514,6 +5514,9 @@ PlayerModelList(void)
             }
         }
 
+        // sort skin names alphabetically
+        qsort(s_skinnames[mdl].data, s_skinnames[mdl].num, sizeof(char**), Q_sort_strcomp);
+
         s_skinnames[mdl].num++;         // guard pointer
 
         // at this point we have a valid player model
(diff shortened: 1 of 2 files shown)


Yamagi:
Thank you! :)

PR #916, opened 2022-10-10
Unix: Add color output for console

0lvin:
Usecase: show map name on level load

### DIFF
--- a/src/backends/unix/system.c
+++ b/src/backends/unix/system.c
@@ -201,18 +201,18 @@ Sys_ConsoleOutput(char *string)
 			if (string[0] == 0x01)
 			{
 				/* red */
-				fputs("\033[40m\033[31;1m", stdout);
+				fputs("\033[31;1m", stdout);
 			}
 			else
 			{
 				/* green */
-				fputs("\033[40m\033[32;1m", stdout);
+				fputs("\033[32;1m", stdout);
 			}
 
 			fputs(string + 1, stdout);
 
-			/* back to black */
-			fputs("\033[40m\033[37;1m", stdout);
+			/* reset to default terminal settings */
+			fputs("\033[0m", stdout);
 			return;
 		}
 	}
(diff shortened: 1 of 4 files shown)


0lvin:
It should also work with MacOS console, i have not tested with MacOS.

devnexen:
works fine on macOs too, do you think as future improvement they can be CVar ?

src/backends/unix/system.c
```diff
@@ -107,6 +110,17 @@ Sys_Quit(void)
...
+
+	term = getenv("TERM");
+
+	if (!strcmp(term, "xterm-color") ||
+		!strcmp(term, "xterm-256color"))
+	{
+		color_active = true;
+
+		Sys_ConsoleOutput("\2Terminal supports colors.\n");
+	}
```
Yamagi, line 123: The idea is good. However I would do it a little bit different, because checking only `xterm` only will break anything which is not xterm. For example multiplexers like screen or tmux or alternative terminals like alacritty or kitty. What about: Check if $TERM includes the substring `color` and if it does not check if $COLORTERM is set?

0lvin: Thank you!

I have updated code with COLORTERM and examples what it can be in case of xfce4-terminal.

Yamagi:
The terminal detection logic looks good. Thank you for fixing this! However, something is still wrong with the `colorterminal` cvar. I'm starting with '+set colorterminal 0` and the game still messes with the terminal. And why is it printing everything in bold? Is that the wanted behavior or bug?

grafik

0lvin:
> I'm starting with '+set colorterminal 0` and the game still messes with the terminal.
Do you mean that with colorterminal = 0 console have changed font style? I suppose its because cvar is initialized after init console. I will look.

> And why is it printing everything in bold? Is that the wanted behavior or bug?
Its expected behavior, code has not checked current font style, check current console style requires full termcap implementation,  and as result new white color is different to old one. I will check what I have, maybe console has not so bright colors.

0lvin:
I have replaced set to white to just reset font to default and removed set background to black.

As result terminals with white background should work better.

Yamagi:
Now it looks good. Thank you for fixing it! :)

PR #919, opened 2022-10-11
Fix for crash with long player skin name

### LINKED ISSUE #918: When skin name's length is greater than 16, game crashes when entering "multiplayer" -> "player setup" menu
While testing the multiplayer menus, I changed to a skin with a long name: "crakhor/sumasyrinx" (len 18). The game allows to set it, but when trying to change it again, the game crashes.

What happens is that the string which stores the skin inside `PlayerConfig_MenuInit`, `mdlname`, has a total length of 16. For comparison, the equivalent string in Q2 3.21 (`currentdirectory`) has a length of 1024.
I'll send a PR addressing this soon.

protocultor:
Skin names for players can have a total length of 32: 15 chars for model + "/" + 15 for skin + "\0". Fixes #918 
Also fixes "playermodels" cmd, which showed "(null)" as a final skin for every model.
Finally, gamepad allowed to select skin in menu.

### DIFF
--- a/src/client/menu/menu.c
+++ b/src/client/menu/menu.c
@@ -5849,6 +5849,7 @@ PlayerConfig_MenuDraw(void)
 static const char *
 PlayerConfig_MenuKey(int key)
 {
+    key = Key_GetMenuKey(key);
     if (key == K_ESCAPE)
     {
         char skin[MAX_QPATH];
--- a/src/client/menu/menu.c
+++ b/src/client/menu/menu.c
@@ -5517,8 +5517,6 @@ PlayerModelList(void)
         // sort skin names alphabetically
         qsort(s_skinnames[mdl].data, s_skinnames[mdl].num, sizeof(char**), Q_sort_strcomp);
 
-        s_skinnames[mdl].num++;         // guard pointer
-
         // at this point we have a valid player model
         s = (char*)malloc(MAX_DISPLAYNAME);
         t = strrchr(s_directory.data[i], '/');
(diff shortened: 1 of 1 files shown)


protocultor:
> I would replace all calls to strncpy with Q_strlcpy. strncpy doesn't work how you might think it does - no null char is added if strlen(src) >= n.

We needed you two months ago!
I've corrected it in every instance inside menu.c.

BjossiAlfreds:
Ok now I'm curious what happened two months ago, haha.

src/client/menu/menu.c
```diff
@@ -5387,7 +5387,7 @@ PlayerDirectoryList(void)
 
         YQ2_COM_CHECK_OOM(s, "malloc()", MAX_QPATH * sizeof(char))
         
-        strncpy(s, t, MAX_QPATH - 1);         // MAX_QPATH - 1, gcc warns about truncation
+        Q_strlcpy(s, t, MAX_QPATH);
         s[strlen(s)] = 0;
```
devnexen, line 5391: wonder, if that s the case, are all of those still needed ?

protocultor: I'm not sure what you mean. Are you talking about doing a full replace from strncpy to Q_strlcpy in the entire project?
If that's the case, there's only one way to find out... but not for this PR. I was going for a particular bug in this.

devnexen: no I meant the line I pointed out.

DanielGibson: he probably means the `s[strlen(s)] = 0;` line

which to be honest has always been pointless, if you think about what strlen() does

devnexen: exactly

protocultor: Sorry about the misunderstanding. Indeed, they are redundant now; they have been deleted.

BjossiAlfreds: Yup, those null char assignments are needless. Even if n == 1, Q_strlcpy would use that 1 byte for the null char.

There is also something to be said about that kind of code being bad practice. If you are doing the same sequence over and over again, that's a hint you should wrap it into a function, otherwise it's only a matter of time before you mess up the sequence and cause bugs or crashes.

devnexen (approved):

protocultor:
Many thanks to everyone involved in this PR, including Yamagi. I've learned a lot with this, and I hope this becomes a guide for future PRs which involve the Q2 menu.

Issue #920, opened 2022-10-12
OGG music bugs/oddities


BjossiAlfreds:
I've encountered 3 weird things with the OGG music feature.

* ignoretrack0 is enabled even when disabled when ogg_curfile <= 0, like after restarting the game. The problem is in playTrack(), a return statement is missing after Ogg_Stop(). playTrack(0) should never continue past that point when ogg_ignoretrack0 is disabled.

Example to reproduce:
1. Make sure ogg_ignoretrack0 is set to 0.
2. Finish all goals in a level to make the music stop.
3. Save and restart the game.
4. Load the save.
5. Music starts playing even though it should be silent.

* When enabling ogg music in the menu, the client config string is being used unguarded, so if you enable it the game will start playing music even if no session is active (CS_CDTRACK is the empty string at that point in time). Checking cls.state should do the trick.

Example to reproduce:
1. Start the game
2. (Re-)enable ogg music
3. Music starts playing even though no CDTRACK is loaded yet 

* The third is more an opinion but I do find it strange that the shuffle menu option calls playTrack(). Have you ever used a music player that starts playing a song or changes the current track when you toggle shuffle on/off? I haven't. I'd recommend just removing all that code and make it purely a shuffle on/off toggle.

I could do a PR to adress these.

Yamagi:
> I could do a PR to adress these.

That would nice. Thank you!

PR #921, opened 2022-10-12
Fixed ignoretrack0 and ogg menu option bugs

BjossiAlfreds:
This PR addresses issues in https://github.com/yquake2/yquake2/issues/920

Added the missing return statement to playTrack() so ignoretrack0 works as intended.
Added a safety check to ogg music menu option.
Removed playTrack() code from shuffle menu option.


Yamagi:
Thank you :)

PR #922, opened 2022-10-17
Fix savegame backward compatibility after cfb86ac.

### LINKED ISSUE #917: cbe501a – Bug – Audio: SFX sound either reversed, out-of-sync, or play for the wrong entity – or completely missing
[Video](https://youtu.be/_Fe1zx_sqOk)

I suppose it's related to the shared model loading PR that's mentioned in #910?

Yamagi:
The steps when loading a savegame are:

1. The server loads the map.
2. Loading the map spawns all entities, i.e. their spawn func is called. This loads the models, precaches the sounds, etc.
3. The savegame is loaded and all the entities in the edict are replaced by the entities read from game.ssv.

When the monster footstep sound were added in cfb86ac, new sounds were added to the spawn functions of most monster entities. This alters the sound indices of all sounds precached at a later time. When a savegame from an older version is loaded, the entities read from game.svv still reference the old sound indices. This leads to the wrong sound getting played.

Work around this by lazy loading the footstep sounds. Instead of loading (precaching) them in the spawn function, load them at their first use. This has the nice side effect of preventing the "missing sound" messages when running without the footsteps.pkz file in baseq2/. It might lead to short stuttering when the sound is played for the first time, though.

Closes #917.

### DIFF
--- a/src/client/menu/menu.c
+++ b/src/client/menu/menu.c
@@ -2113,23 +2113,6 @@ static void
 OGGShuffleFunc(void *unused)
 {
     Cvar_SetValue("ogg_shuffle", s_options_oggshuffle_box.curvalue);
-
-    cvar_t *ogg_enable= Cvar_Get("ogg_enable", "1", CVAR_ARCHIVE);
-	int track = (int)strtol(cl.configstrings[CS_CDTRACK], (char **)NULL, 10);
-
-    if (s_options_oggshuffle_box.curvalue)
-    {
-        Cvar_Set("ogg_shuffle", "1");
-    }
-    else
-    {
-        Cvar_Set("ogg_shuffle", "0");
-    }
-
-	if (ogg_enable->value)
-	{
-		OGG_PlayTrack(track);
-	}
 }
 
 static void
@@ -2143,8 +2126,11 @@ EnableOGGMusic(void *unused)
         OGG_InitTrackList();
         OGG_Stop();
 
-        int track = (int)strtol(cl.configstrings[CS_CDTRACK], (char **)NULL, 10);
-        OGG_PlayTrack(track);
+		if (cls.state == ca_active)
+		{
+	        int track = (int)strtol(cl.configstrings[CS_CDTRACK], (char **)NULL, 10);
+	        OGG_PlayTrack(track);
+		}
     }
     else
     {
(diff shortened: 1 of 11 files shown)


Yamagi:
@0lvin, @DanielGibson: Can you have a look?

Yamagi:
A Windows build is here: https://deponie.yamagi.org/quake2/windows/testbuilds/quake2-20221017-80bd398-fix_savegame_compat.zip

Yamagi:
Mkay, this doesn't work for the second level.

Yamagi:
Mkay again. I was too fast. It should be enough to invalidate the cached sound indices when spawning a new monster.

0lvin:
I like your idea. 

What do you think about use for group load sounds? 
e.g.:
```c
if (!sound_step ||  !sound_step2)
{
     sound_step = gi.soundindex("berserk/step1.wav");
     sound_step2 = gi.soundindex("berserk/step2.wav");
}
```

0lvin (approved):

BjossiAlfreds:
Storing a lazy loaded sound index in a static variable won't work unless you reset the values to 0 in the spawn functions. Otherwise the outdated index value will stay through level transitions and cause wrong sounds to play instead of footteps.

DanielGibson:
You're not wrong, but that's also what the current version does, see for example https://github.com/yquake2/yquake2/pull/922/files#diff-e0a1fcb0b98115edf75ff39ec6cfbacc888fc10015eeee849b3bd7d95053c8b4R548-R552

BjossiAlfreds:
Yeah I was looking at the outdated code, never mind what I said.

Yamagi:
> What do you think about use for group load sounds?

Good idea. I did that. Makes the code more readable and won't matter from a performance point of view. The user either has the pak with all footsteps sounds or nothing. And even if they are stupid, calling `gi.soundindex()` too often shouldn't have measureable impact.

I won't fix the corner case when a user sets `g_monsterfoodsteps 1` on an old savegame. Too much hassle for too little gain. I will mention that case in the CHANGELOG.

Pushed it to master. Thank you to everyone involved. :)

Yamagi:
Forgot to mention: This will break savegames created between the monster footstep change was merged on October 8 and the fix yesterday. This is caused by the fix reordering the order in which the sounds are loaded.

PR #924, opened 2022-10-20
Shared image load

0lvin:
After apply changes {gl1,gl3,soft} could implement only `R_FindImage` for search space for image or reuse already used image and `R_LoadPic` for convert image data to internal structure,   and use `FindPic`/`GetTexImage`/`GetSkyImage` everywhere where need to load picture in menu/get wall image/get sky image.


0lvin:
Additionally added m32 format support. Its same as m8 but with 24bit color.

Yamagi:
This looks good :) A little problem, though: You can't use `LoadImage()` as function name, it collides with a WinAPI macro with the same name. Can you rename it to something like `R_LoadImage()`?

The compiler error is: src/client/refresh/gl3/header/../../ref_shared.h:86:24: error: conflicting types for 'LoadImageA'

I'll test the PR in the next few days by playing through xatrix.

Yamagi:
A new Windows build with this PR included can be found here: https://deponie.yamagi.org/quake2/windows/testbuilds/quake2-20221022-2c81e25.zip

I hacked around the problem mentioned above with `#unifdef LoadImage()`...

0lvin:
Thank you!

I have renamed LoadImage and fixed bug with allocate memory in model (used wrong struct for get size), and  moves same structure types from {soft,gl1,gl3}/header/model.h. 

One more structure is similar but min/max has short instead float type and i have added comment about. It can be that its reason of FAN bug for soft render in mine level.

Yamagi:
I just wanted to say, that I haven't forgotten about this. I didn't have the time until now to play at least some units with and without the various retexturing packs applied and this PR.

0lvin:
Thank you!

Yamagi:
I am back from the dead. I have spend the last week playing through The Reckoning with this branch and tested several combinations of renderers and retexturing packs. No obvious problems. I'm merging this, let's see if any problems show up. Thank you and keep up the good work!

0lvin:
Thank you, i have created new release https://github.com/yquake2/ref_vk/releases/tag/v1.0.4

PR #722, opened 2021-06-13
Sync haptic feedback with sound track

0lvin:
Sync haptic feedback with original delay/attack/fade/volume/length of sound and remove effect hard coded haptic effects.

PR #927, opened 2022-10-23
Controller rumble support

### LINKED ISSUE #858: Haptic/Force feedback doesn't work for xbox wireless elite controller 2 on Windows
Hey! First, thanks for the awesome project. 
I've tried Windows builds 8.01 and 8.10 on Windows 11 and found out that Haptic/Force feedback doesn't work for xbox wireless elite controller 2 though controller itself works good, all buttons are mapped correctly too.
It would be great if you could look into it this issue. I could help with testing, if needed.
Many thanks!

protocultor:
This PR replaces the current haptic support, "deprecated" (*) since SDL2.0.14 for gamepads, for a more traditional rumble, supported on a wide range of controllers by SDL2.

I must warn that this development makes visible some issues, but those were already present in the classic haptic work, as tested in a Raspberry Pi 3 B+, the only platform I have an old version of SDL available (2.0.10), where gamepads were still supported by haptic. These are:

- Difficulty to properly identify which sounds come from the player and which come from the environment; the most evident example is that every player's shooting and pain sound will count for feedback in multiplayer. Likewise, a few monster's shots are identified as "from the player"; at least the Phalanx shots from the Gladiator Beta of the xatrix expansion count for this.
- Weapons from the rogue expansion behave differently, so they don't have rumble. The Plasma Beam uses the "laser hit" sound, common in the game to represent missed enemy laser fire (e.g. from the Flyer), but it doesn't come from the player, it comes from where the weapon hits, making it unable to distinguish itself from the mentioned case. On the other hand, the Chainfist doesn't play its sounds from S_StartSound() like the rest, so it's not calling the rumble function.
- OpenAL's cache length is in milliseconds, which properly sets the duration of the rumble. However, the SDL sound system provides this length in... another format (?) which is 40 to 50 times bigger, and it's not clear which "unit of time" uses (if that's a unit of time at all). Sidestepped this by dividing the length by 45, but it still doesn't provide the real length of the sound played, so rumble doesn't have an accurate duration. Sticking to OpenAL is the true solution for this.

Even with these problems (which, again, share with "haptic" implementation), I must say this improves quite a bit the single player experience for controller people.
Tested with a DualShock 4 (PS4) and a Switch Pro Controller, and in both rumble works wired and wireless, in Ubuntu 20.04, Windows 7, and even the very same Raspberry Pi 3 B+.

It's worth mentioning this is not a completely straight conversion "from haptic to rumble"; I've "re-filtered" which sounds provide feedback or not. For example, eliminated the "menu" sounds, and added the "quake" sounds, so earthquakes will have the proper feeling :) . Another example: ride a platform, then get off it.

This (hopefully) closes #858 
If this is accepted, I wanted to ask for a testbuild.

(*) Classic haptic SDL2 support has been reduced to target racing wheels and flightsticks, while gamepads must use the rumble function used in this PR:
https://github.com/libsdl-org/SDL/issues/4435#issuecomment-896432954
At this point, I don't even know if force feedback mice work with haptic.

### DIFF
--- a/src/client/header/keyboard.h
+++ b/src/client/header/keyboard.h
@@ -285,10 +285,8 @@ void Key_ReadConsoleHistory();
 void Key_WriteConsoleHistory();
 void Key_SetBinding(int keynum, char *binding);
 void Key_MarkAllUp(void);
-void Haptic_Feedback(char *name, int effect_volume, int effect_duration,
-				   int effect_begin, int effect_end,
-				   int effect_attack, int effect_fade,
-				   int effect_x, int effect_y, int effect_z);
+void Controller_Rumble(const char *name, vec3_t source, qboolean from_player,
+				unsigned int duration, unsigned short int volume);
 int Key_GetMenuKey(int key);
 
 #endif
--- a/src/client/input/sdl.c
+++ b/src/client/input/sdl.c
@@ -806,6 +806,7 @@ IN_Update(void)
 					break;
 				}
 				if (event.cdevice.which == SDL_JoystickInstanceID(SDL_GameControllerGetJoystick(controller))) {
+					Cvar_SetValue("paused", 1);
 					IN_Controller_Shutdown(true);
 					IN_Controller_Init(false);
 				}
@@ -819,6 +820,23 @@ IN_Update(void)
 					countdown_reason = REASON_CONTROLLERINIT;
 				}
 				break;
+
+#if SDL_VERSION_ATLEAST(2, 24, 0)	// support for battery status changes
+			case SDL_JOYBATTERYUPDATED:
+				if (!controller || event.jbattery.which != SDL_JoystickInstanceID(SDL_GameControllerGetJoystick(controller)))
+				{
+					break;
+				}
+				if (event.jbattery.level == SDL_JOYSTICK_POWER_LOW)
+				{
+					Com_Printf("WARNING: Gamepad battery Low, it is recommended to connect it by cable.\n");
+				}
+				else if (event.jbattery.level == SDL_JOYSTICK_POWER_EMPTY)
+				{
+					SCR_CenterPrint("ALERT: Gamepad battery almost Empty, will disconnect anytime.\n");
+				}
+				break;
+#endif	// SDL_VERSION_ATLEAST(2, 24, 0)
 
 			case SDL_QUIT:
 				Com_Quit();
--- a/src/client/input/sdl.c
+++ b/src/client/input/sdl.c
@@ -812,7 +812,7 @@ IN_Update(void)
 				}
 				else if (event.jbattery.level == SDL_JOYSTICK_POWER_EMPTY)
 				{
-					SCR_CenterPrint("ALERT: Gamepad battery almost Empty, will disconnect anytime.\n");
+					SCR_CenterPrint("ALERT: Gamepad battery almost Empty.\n");
 				}
 				break;
 #endif	// SDL_VERSION_ATLEAST(2, 24, 0)
@@ -1779,6 +1779,7 @@ IN_Controller_Shutdown(qboolean notify_user)
 		SDL_GameControllerClose(controller);
 		controller = NULL;
 		gyro_hardware = false;
+		show_haptic = false;
 		joystick_left_x = joystick_left_y = joystick_right_x = joystick_right_y = 0;
 		gyro_yaw = gyro_pitch = 0;
 		normalize_sdl_gyro = 1.0f / M_PI;
--- a/src/client/input/sdl.c
+++ b/src/client/input/sdl.c
@@ -108,30 +108,9 @@ static cvar_t *windowed_mouse;
 
 // ----
 
-struct hapric_effects_cache {
-	int effect_volume;
-	int effect_duration;
-	int effect_begin;
-	int effect_end;
-	int effect_attack;
-	int effect_fade;
-	int effect_id;
-	int effect_x;
-	int effect_y;
-	int effect_z;
-};
+qboolean show_haptic = false;
 
-qboolean show_haptic;
-
-static SDL_Haptic *joystick_haptic = NULL;
 static SDL_GameController *controller = NULL;
-
-#define HAPTIC_EFFECT_LIST_SIZE 16
-
-static int last_haptic_volume = 0;
-static int last_haptic_efffect_size = HAPTIC_EFFECT_LIST_SIZE;
-static int last_haptic_efffect_pos = 0;
-static struct hapric_effects_cache last_haptic_efffect[HAPTIC_EFFECT_LIST_SIZE];
 
 // Joystick sensitivity
 static cvar_t *joy_yawsensitivity;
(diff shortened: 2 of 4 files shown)


0lvin:
Could be used cvar  to make rumble/haptic support optional?

By SDL bug description it looks like it's windows xinuput specific regression. 

Code could make rumble default for Windows and haptic feedback default for Linux. In such case user can choose what is works better for him by change cvar.

Haptic feedback has advantages like direction, volume and wave form. As I know ps4/ps5 controllers have much wider functionality than just make noise with some frequency as described in bug for Windows drivers

0lvin:
About hard-coded sound names, different mods use different sound mane schema and sound names cold be used as filtering for unpleasant sounds instead of filtering what can produce feedback.

0lvin:
Looks as code already is filtering by sound name prefix.
E.g.:
```c
if (
		strstr(name, "misc/menu") ||
		strstr(name, "weapons/") ||
		/* detect pain for any player model */
		((
			strstr(name, "player/") ||
			strstr(name, "players/")
		) && (
			strstr(name, "/pain")
		)) ||
		strstr(name, "player/step") ||
		strstr(name, "player/land")
	)
```
Could be replaced with filter by list of prefixes, like   'player/ players/ misc/menu/ weapon/ ...' and so on.

protocultor:
> Could be used cvar to make rumble/haptic support optional?

Maybe, but didn't find it worthy, unless there are some hidden community of force feedback mouse users.

> By SDL bug description it looks like it's windows xinuput specific regression.
> 
> Code could make rumble default for Windows and haptic feedback default for Linux. In such case user can choose what is works better for him by change cvar.

It is not a "windows specific issue". It has been the new normal since 8 SDL releases, **for any OS**, with no sign of getting back. I have a PS4 controller that I use with Ubuntu, Windows, RPi and MacOS (forgot to mention the last one in the first post), and the only platform that recognizes it as a "haptic device" is the Raspberry Pi... because it's stuck with version 2.0.10 of SDL.

> Haptic feedback has advantages like direction, volume and wave form. As I know ps4/ps5 controllers have much wider functionality than just make noise with some frequency as described in bug for Windows drivers

If those advantages really exist, it's hard to notice, at least when trying the PS4 controller in SDL 2.0.10, because the motors sometimes would get crazy for no reason at all. Maybe that's the reason they dropped "haptic" support for it, because it doesn't behave as advertised. Everything I say is probably meaningless until you get a PS4 controller and compare this branch with the current master, in both SDL 2.0.10 and 2.24.0.

protocultor:
> About hard-coded sound names, different mods use different sound mane schema and sound names cold be used as filtering for unpleasant sounds instead of filtering what can produce feedback.
> 
> What do you think about to have variable with list of sounds that should not produce feedback?

That would be nice, but wouldn't be enough. Apart from that, you would need a list of behaviors, and a way to associate sounds with those behaviors. The most evident example is the blaster/hyperblaster duo, which in the "haptic" version they rumble like the world is going to end, forcing me to dampen them greatly.

> Looks as code already is filtering by sound name prefix.
> Could be replaced with filter by list of prefixes, like 'player/ players/ misc/menu/ weapon/ ...' and so on.

Prefixes are nice for a first filter, but as I say, even within "weapons/" there are a lot of specific cases you would want to treat individually; see my code for that. Of course the code can be improved, but the objective of this PR is "resurrecting" a feature that was left behind for a good while.

This is a good criticism, but honestly I'm more worried about the first problem I talked about in the PR presentation: having the sounds of all players in a multiplayer match count for haptic/rumble is not good.

0lvin:
Haptic is not deprecated by release logs, just rumble is added for simplified configuration for 2.0.9. it looks as both functions are supported for some devices list.

```
Added SDL_GameControllerRumble() and SDL_JoystickRumble() which allow simple rumble without using the haptics API
```
https://github.com/libsdl-org/SDL/releases

protocultor:
The only documentation that's worth in SDL is the code itself, and the comments from slouken.
I can tell two devices that are NOT supported for haptic in any OS since 2.0.14: PS4 and Switch Pro controllers. Then, as most of the complaints point out, there's the Xbox controller, in Windows at least. This trio is the current golden standard of controllers, so it's better to ask "what is supported by haptic". Not controllers, that's all I can tell.

Again, I need you to test live with a gamepad and tell me what you see.

Some interesting links:
https://github.com/libsdl-org/SDL/issues/3937
https://stackoverflow.com/questions/65189813/sdl2-haptics-methods-behaving-not-as-expected
https://github.com/veandco/go-sdl2/issues/315
https://forum.rvgl.org/viewtopic.php?t=1367#p6052

0lvin:
As filter for sound names you could use existed pre https://github.com/yquake2/yquake2/pull/722/files, it was tested with all three official mods of quake2. 

I don't have controller now, I have checked code last January with latest sdl2 available in debian testing. Haptic worked without issues at that moment. It was some cheap noname controller.

protocultor:
> As filter for sound names you could use existed pre #722 code with some previously defined values of duration and volume, it was tested with all three official mods of quake2.

That testing work has been done, I'm used to "eat my own dog food". Look at the `Controller_Rumble()` function, it sets some defaults and then changes them depending on the name of the sound played; at least that comparison of the names of the sounds is pretty much based on your `Haptic_Feedback()` code you posted in your 3rd message.
A difference is that I set `intens(ity)` as a factor of the real volume that comes as a parameter, instead of hardcoding it like before #722. `duration` can also change by a factor.
`low_freq(uency)` and `hi(gh)_freq(uency)` are factors for each of the motors of the controller, which can also change depending of the sound played.
Something I've added to your work is that the name of the sound can even change the maximum distance where that sound is allowed to be felt, something super useful in the mentioned "moving platform" effect, and with explosions and ambient sounds.
Many thanks for your work in the "direction of the sound" calculation, it's still being used as you can see in sound.c.

> I don't have controller now, I have checked code last January with latest sdl2 available in debian testing. Haptic worked without any issues at that moment. It was some cheap noname controller.

If Debian's package manager is anything like Ubuntu's, and depending on your platform, `libsdl2-dev` may have been 2.0.10. I have cloned the SDL source and installed it manually.
Or you just had luck and your controller was detected as haptic, something very unusual at this point. In any case, it's a shame you don't have that controller anymore, because those are one of the reasons `SDL_GameController` was invented, and I would have liked you to see what it's capable of.

Yamagi:
As always: Very good work! Thank you!

0lvin:
@Yamagi could we have haptic feedback as optional functionality under some cvars? 

I don't have access to game controller now, I could restore when I have access to some controller with haptic feedback. What do you think?

Yamagi:
I don't have really an opinion, because I have never played Quake II with some kind of controller or high-tech gamer mouse. However, having (in a broader sense) controversial features as optional features is never a bad idea. So I'm happy to merge a pull request.

protocultor:
FYI, haptic will be deprecated in a couple of SDL releases:
https://github.com/libsdl-org/SDL/issues/4435#issuecomment-1304867524

0lvin:
Do you mean SDL3?

protocultor:
It's a reminder that _haptic_ is supported just by SDL 2.0.0 through 2.0.12, so it not only doesn't have a present, it's also confirmed not to have a future. Still, I also don't want YQ2 to abandon SDL 2.0.10; one of the reasons of my involvement with this project is because how well it runs on Raspberry Pi.

Just wanted to mention that haptic lived (and died) on a theoretical "superiority" that wasn't actually there in practice, according to my experience playing YQ2 8.10 on a controller. Before you begin coding, I encourage you to try and play some maps with the current code's state, because that will be your new goal to overcome.

PR #930, opened 2022-10-25
update gl1 renderer menu similarly as gl3's

devnexen:


0lvin:
What do you think about rename gl1_colorlight and gl3_colorlight to r_colorlight in https://github.com/yquake2/yquake2/blob/master/src/common/cvar.c#L39, and reuse r_colorlight in all renders except soft render? Soft render is little bit different story because it has color light disabled by default if compare to other renders where it enabled by default.

src/client/menu/videomenu.c
```diff
@@ -503,29 +505,52 @@ VID_MenuInit(void)
...
+		s_intensity_slider.maxvalue = 10.0f;
+
+		s_overbrightbits_slider.generic.type = MTYPE_SLIDER;
+		s_overbrightbits_slider.generic.name = "overbrights";
+		s_overbrightbits_slider.generic.x = 0;
+		s_overbrightbits_slider.generic.y = (y += 10);
+		s_overbrightbits_slider.cvar = "r_overbrightbits";
+		s_overbrightbits_slider.minvalue = 0;
+		s_overbrightbits_slider.maxvalue = 3;
+		s_overbrightbits_slider.slidestep = 1;
```
protocultor, line 545: Using the very same slider structure for both GL1 and GL3 "overbrightbits" produces some undesired behavior: if you begin with GL1 and switch to GL3, the "slidestep" keeps being `1`, so you lose the capability of changing its value by steps of `0.1` until you restart the game.
Also, since "overbright(bits)" is an integer value for GL1, maybe it's convenient to print it without decimals (`.printformat = "%.0f";`).
See the FOV slider for an example.

src/common/cvar.c
```diff
@@ -89,7 +89,11 @@ replacement_t replacements[] = {
 	{"gl_shadows", "r_shadows"},
 	{"gl_anisotropic", "r_anisotropic"},
 	{"gl_lightmap", "r_lighmap"},
-	{"intensity", "gl1_intensity"}
+	{"intensity", "gl1_intensity"},
```
0lvin, line 92: Intensity -> r_intensity
Gl1_intensity -> r_intensity
Vk_intensity -> r_intensity

As know, name of variable should be also changed in {gl1,gl3}_main.c, e.g. https://github.com/yquake2/yquake2/blob/master/src/client/refresh/gl1/gl1_main.c#L1230 

Vk_ref change could be separate PR, if/when current PR is approved.

And check of current render will be not required as all variables will be same.

DanielGibson:
Please don't rename gl3_intensity and gl3_overbrightbits.
The meaning (or visual effect) of those CVars are slightly different between GL1 and GL3, which is why I gave them different names (so GL1 and GL3 can be configured independently), see also descriptions at https://github.com/yquake2/yquake2/blob/master/doc/040_cvarlist.md#graphics-opengl-14-only

0lvin:
@DanielGibson What do you think about backport gl3 logic to gl1 and only after that reuse? Will it be possible, or it will be to difficult to be worth?

DanielGibson:
I don't think that's feasible, GL3 uses shaders, GL1 doesn't

I really think they should use different CVars so users can set different values for GL1 vs GL3.

I don't really see why those cvars should be unified - in the menu code, it's different code paths already (`if (strcmp(vid_renderer->string, "gl3") == 0 || strcmp(vid_renderer->string, "gles3") == 0)` vs `else if (strcmp(vid_renderer->string, "gl1") == 0)`) so using different cvar names should be straightforward?

devnexen:
that was my initial approach

0lvin:
I am afraid that with different cvars aligned to visually same option in menu will user bugs, that something is not applied and has different values to things expected by user. Different cvars names in console provides confidence what is really changed, and when different variables changed by menu in visually same place it can produce additional bug reports. :-(

DanielGibson:
> I am afraid that with different cvars aligned to visually same option in menu will user bugs, that something is not applied and has different values to things expected by user.

I don't really get this argument?
No matter if we use the same CVar names or not, the options in the menu will be at the same position
But that doesn't change the fact that the meaning of the CVars (how what values looks in the game) depends on the renderer. And note that those CVars even have different default values: gl1_intensity looks good with 2.0, while gl3_intensity looks good better 1.5

And for the overbrightbits, GL1 only accepts **integer** values 0,1,2,3, while gl3_overbrightbits accepts any floating point value (and we found 1.3 to look best - a value that GL1 doesn't even support) - so ideally the menu should behave differently for them (only offer those integer values for GL1, and for GL3 a slider, probably between 0.0 and 5.0 or something and 0.1 steps?)

When using the same names on all renderers, people might complain that they configured something for GL1 and then switched to GL3 and it looks different, even though the setting is still the same in the menu.
When using different names, at least users will see when going to the video menu after changing the renderer, that the value is different, and can be changed without affecting how the other renderer will look.

src/common/cvar.c
```diff
@@ -89,7 +89,12 @@ replacement_t replacements[] = {
 	{"gl_shadows", "r_shadows"},
 	{"gl_anisotropic", "r_anisotropic"},
 	{"gl_lightmap", "r_lighmap"},
-	{"intensity", "gl1_intensity"}
+	{"intensity", "r_intensity"},
+	{"gl1_intensity", "r_intensity"},
+	{"gl1_colorlight", "r_colorlight"},
```
DanielGibson, line 94: gl1_colorlight doesn't exist

0lvin:
I see

> When using different names, at least users will see when going to the video menu after changing the renderer, that the value is different.

I am fully agree, its exactly my point. Or use same variables and logic behind it, some different names in menu.

0lvin (approved):

Yamagi:
@DanielGibson, are you okay with this?

src/client/menu/videomenu.c
```diff
@@ -499,33 +508,58 @@ VID_MenuInit(void)
...
-		s_intensity_slider.maxvalue = 256.0f;
-
-		s_overbrightbits_slider.generic.type = MTYPE_SLIDER;
-		s_overbrightbits_slider.generic.name = "overbrights";
-		s_overbrightbits_slider.generic.x = 0;
-		s_overbrightbits_slider.generic.y = (y += 10);
-		s_overbrightbits_slider.cvar = "gl3_overbrightbits";
-		s_overbrightbits_slider.minvalue = 0.1f;
-		s_overbrightbits_slider.maxvalue = 256.0f;
+		gl1_colorlight = 0;
```
DanielGibson, line 511: it's a pointer, so this should use NULL instead of 0 (same below in the else case)

src/client/menu/videomenu.c
```diff
@@ -499,33 +508,58 @@ VID_MenuInit(void)
...
+		s_gl3_intensity_slider.minvalue = 0.1f;
+		s_gl3_intensity_slider.maxvalue = 256.0f;
+
+		s_gl3_overbrightbits_slider.generic.type = MTYPE_SLIDER;
+		s_gl3_overbrightbits_slider.generic.name = "overbrights";
+		s_gl3_overbrightbits_slider.generic.x = 0;
+		s_gl3_overbrightbits_slider.generic.y = (y += 10);
+		s_gl3_overbrightbits_slider.cvar = "gl3_overbrightbits";
+		s_gl3_overbrightbits_slider.minvalue = 0.1f;
+		s_gl3_overbrightbits_slider.maxvalue = 256.0f;
```
Yamagi, line 526: For `gl3_intensity` and `gl3_overbrightbits` a maximum value of 256 means that all sane values are within the first pixels of the slider. What about limiting the menu to a maxumum value of 5? If anyone really wants more they can set it through the console.

src/client/menu/videomenu.c
```diff
@@ -499,33 +508,58 @@ VID_MenuInit(void)
...
+		s_gl1_intensity_slider.maxvalue = 10.0f;
+
+		s_gl1_overbrightbits_slider.generic.type = MTYPE_SLIDER;
+		s_gl1_overbrightbits_slider.generic.name = "overbrights";
+		s_gl1_overbrightbits_slider.generic.x = 0;
+		s_gl1_overbrightbits_slider.generic.y = (y += 10);
+		s_gl1_overbrightbits_slider.cvar = "gl1_overbrightbits";
+		s_gl1_overbrightbits_slider.minvalue = 0;
+		s_gl1_overbrightbits_slider.maxvalue = 3;
+		s_gl1_overbrightbits_slider.printformat = "%.0f";
```
Yamagi, line 554: Enforcing a printformat / enforcing it this way messes up the scale. Just give `minvalue` and `maxvalue` as floats, remove the `printformat` and it should be fine.

protocultor: > Enforcing a printformat / enforcing it this way messes up the scale. Just give `minvalue` and `maxvalue` as floats, remove the `printformat` and it should be fine.

Better than that, he should give `slidestep` a value of 1, since gl_overbrightbits is an integer value, and it should change in steps of '1' instead of the default of  '0.1'. See the FOV slider for a fully working example.

Yamagi: Ah yes, you are right. Of course.

Yamagi:
Merged it and changed the GL1 overbright bits to integers in a separate commit. Thank you! :)

PR #936, opened 2022-11-18
OGG_LoadAsWav, release as long it had been allocated.

devnexen:


DanielGibson:
Hmm probably only relevant in the unlikely (maybe impossible?) case that either `res != 0` or `ogg2wav_file->channels <= 0` but `ogg2wav_file` itself *not* being NULL, but shouldn't hurt at least ;)

Yamagi:
Thank you :)

PR #941, opened 2022-11-23
Game Controller code refactor

protocultor:
* #849 included a "#ifndef _WIN32" to fix a difference between Linux/Mac and Windows that wasn't supposed to exist. As of SDL 2.26.0, this difference has been fixed, so the "if" and the variable that it changed have been removed.
* Joysticks open with error checking, to avoid possible game crashes.
* Slight optimization for gyro aiming: some calculations only are made when needed.
* Console messages improved for game controller init: less useful text has been minimized, while more useful info is more verbose.
* Turns out, controller rumble support can be evaluated with SDL version < 2.0.18, by checking the return value of `SDL_GameControllerRumble()`: 0 rumbles, -1 doesn't.
* Speaking of `SDL_GameControllerRumble()`, now there's a check to see if it can be called, since it doesn't exist with SDL < 2.0.9; this allows the game to be compiled with any SDL2 version.


Yamagi:
Thank you!

PR #944, opened 2022-12-03
Restore haptic feedback functionality.

0lvin:
Rumble feedback will be default.
Checked in both modes with:
* LibSDL2: 2.24.2+dfsg-1 (debian testing)
* Microntek USB Joystick (DragonRise hid-dr.ko)

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -547,11 +547,11 @@ Set `0` by default.
   'calibrate' in the 'gamepad' -> 'gyro' menu to set them.
 
 * **joy_haptic_magnitude**: Haptic magnitude value, By default this cvar
-  is `0.0` or disabled. Valid values are 0..1.0.
+  is `0.0` or disabled. Valid values are positive, e.g. 0..2.0.
 
 * **joy_haptic_distance**: Haptic maximum effect distance value, By default
-  this cvar is `100.0`. Any positive value is valid. E.g. effect of shot
-  near barrel to barrel has 58 points.
+  this cvar is `100.0`. Any positive value is valid. E.g. effect of shoot
+  to a barrel has 58 points when player stay near the barrel.
 
 * **s_feedback_kind**: Select kind of controller feedback to use. By default
   this cvar is `0`. Possible values:
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -549,6 +549,11 @@ Set `0` by default.
 * **joy_haptic_magnitude**: Haptic magnitude value, By default this cvar
   is `0.0` or disabled. Valid values are positive, e.g. 0..2.0.
 
+* **joy_haptic_filter**: List of sound file names produced haptic feedback
+  separated by space. `*` could be used for replace part of file name as
+  regular expression. `!` at the beginning of value could be used for skip
+  file name equal to value.
+
 * **joy_haptic_distance**: Haptic maximum effect distance value, By default
   this cvar is `100.0`. Any positive value is valid. E.g. effect of shoot
   to a barrel has 58 points when player stay near the barrel.
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -547,7 +547,11 @@ Set `0` by default.
   'calibrate' in the 'gamepad' -> 'gyro' menu to set them.
 
 * **joy_haptic_magnitude**: Haptic magnitude value, By default this cvar
-  is `0.0` or disabled. Valid values are 0..1.0;
+  is `0.0` or disabled. Valid values are 0..1.0.
+
+* **joy_haptic_distance**: Haptic maximum effect distance value, By default
+  this cvar is `100.0`. Any positive value is valid. E.g. effect of shot
+  near barrel to barrel has 58 points.
 
 * **s_feedback_kind**: Select kind of controller feedback to use. By default
   this cvar is `0`. Possible values:
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -270,15 +270,10 @@ Set `0` by default.
   default this cvar is disabled (set to 0).
 
 * **s_reverb_preset**: Enable reverb effect. By default this cvar is disabled
-  (set to `-1`). Possibe values:
+  (set to `-1`). Possible values:
   `-2`: Auto reverb effect select,
   `-1`: Disable reverb effect,
   `>=0`: select predefined effect.
-
-* **s_feedback_kind**: Select kind of controller feedback to use. By default
-  this cvar is `0`. Possibe values:
-  `0`: Rumble feedback,
-  `1`: Haptic feedback.
 
 ## Graphics (all renderers)
 
@@ -551,6 +546,13 @@ Set `0` by default.
   the controller alone. As these vary by device, it's better to use
   'calibrate' in the 'gamepad' -> 'gyro' menu to set them.
 
+* **joy_haptic_magnitude**: Haptic magnitude value, By default this cvar
+  is `0.0` or disabled. Valid values are 0..1.0;
+
+* **s_feedback_kind**: Select kind of controller feedback to use. By default
+  this cvar is `0`. Possible values:
+  `0`: Rumble feedback,
+  `1`: Haptic feedback.
 
 ## cvar operations
 
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -275,6 +275,11 @@ Set `0` by default.
   `-1`: Disable reverb effect,
   `>=0`: select predefined effect.
 
+* **s_feedback_kind**: Select kind of controller feedback to use. By default
+  this cvar is `0`. Possibe values:
+  `0`: Rumble feedback,
+  `1`: Haptic feedback.
+
 ## Graphics (all renderers)
 
 * **cin_force43**: If set to `1` (the default) cinematics are displayed
--- a/src/client/input/sdl.c
+++ b/src/client/input/sdl.c
@@ -1768,6 +1768,12 @@ Controller_Rumble(const char *name, vec3_t source, qboolean from_player,
 #if SDL_VERSION_ATLEAST(2, 0, 9)
 	if (SDL_GameControllerRumble(controller, low_freq, hi_freq, duration) == -1)
 	{
+		if (!joystick_haptic)
+		{
+			/* no haptic, some other reason of error */
+			return;
+		}
+
 		/* All haptic/force feedback slots are busy, try to clean up little bit. */
 		IN_Haptic_Effects_Shutdown();
 	}
--- a/src/client/input/sdl.c
+++ b/src/client/input/sdl.c
@@ -108,9 +108,30 @@ static cvar_t *windowed_mouse;
 
 // ----
 
+struct hapric_effects_cache {
+	int effect_volume;
+	int effect_duration;
+	int effect_begin;
+	int effect_end;
+	int effect_attack;
+	int effect_fade;
+	int effect_id;
+	int effect_x;
+	int effect_y;
+	int effect_z;
+};
+
 qboolean show_gamepad = false, show_haptic = false, show_gyro = false;
 
+static SDL_Haptic *joystick_haptic = NULL;
 static SDL_GameController *controller = NULL;
+
+#define HAPTIC_EFFECT_LIST_SIZE 16
+
+static int last_haptic_volume = 0;
+static int last_haptic_efffect_size = HAPTIC_EFFECT_LIST_SIZE;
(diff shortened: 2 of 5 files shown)


protocultor:
It's too bad that you just "resurrected" your implementation without taking the time to improve it.
In `Haptic_Feedback()`, `name` should be a `const char *`.
Original typos still exist: "hap**r**ic", "eff**f**ect".
Did you take the time to play with rumble, and try again with haptic? Can you describe the difference?
What is your take on having haptic effects with events that are not related to the player, like grenades bouncing on walls, and not having them with events that should be felt, like earthquakes?
Just some questions from a controller player.

Yamagi:
I'll wait for @protocultor to approve this before merging.

0lvin:
Thank you, I will look.

Rumble feedback is targeted for controllers with two motors with different frequencies (low + hi),  Microntek USB Joystick in my case has two fully same motors and I can't taste difference in frequency of feedback. And both of feedback methods are produced fully fine effects as for me.

0lvin:
@protocultor @Yamagi could you please look? 

I have fixed small bug with internal haptic/rumble/force feedback. 

SDL uses one slot in haptic effect table, as result sometimes we can use only one less haptic effects in same time, I have added clean up of whole queue of known effects and mark that should be used one effect less than device reported. 

I have only one difference between rumble and haptic feedback for my case, rumble can produce only one effect in same time and new effect will stop previous, haptic can use up to 16 effects in same time.it will be less than 16 in real world as code does not look what is state of effect(is played now or not) and just remove oldest effect from list.

protocultor:
> Rumble feedback is targeted for controllers with two motors with different frequencies (low + hi), Microntek USB Joystick in my case has two fully same motors and I can't taste difference in frequency of feedback.

You just described every single controller in existence since the PlayStation 2 one, with the exception of the last generation ones, which add motors in the triggers, and have a different treatment using SDL.

As usual since 6 or 7 SDL releases, I'm not able to test this in my dev machine (Ubuntu 20.04, SDL 2.26.1) because "_the PS4 controller is not a haptic device_", so I had to test this in a Raspberry Pi 3B+ (SDL 2.0.10).
I like how unintrusive this is, and I couldn't notice any performance loss in the RPi; this is always a positive. I also like how by binding `"toggle s_feedback_kind"` to a key, I could switch between both feedback modes on the fly, without issues. I still don't like the feeling this provides (seriously, fire the ETF Rifle of the rogue expansion with this on xD), but I'm not in the target audience, and again, I couldn't obtain any conflict with the current rumble implementation.
I gave my minor gripes in the code itself.

0lvin:
Thank you!

protocultor:
...it still doesn't mean I approve of this. Please check the reviews I left.

0lvin:
Could you please review? Both of kind of feedback have used same code for calculate effect volume, difference is only in usage hi_freq/low_freq. Haptic feedback discard values of *_freq.

I have ported step feedback to shared logic. Menu sound feedback removed for now, it had used for  fine tuning of feedback in menu slider, I would propose to discuss implementation it in some other way as separate patch, maybe directly call feedback in menu code.

protocultor:
> I have ported step feedback to shared logic. Menu sound feedback removed for now, it had used for fine tuning of feedback in menu slider, I would propose to discuss implementation it in some other way as separate patch, maybe directly call feedback in menu code.

The whole point about haptic feedback is having... feedback about the world around you. The steps you make are not external to you, so it doesn't make any sense to make them vibrate, unless you want to communicate the player something, like "you are walking on quick sands", "you are in control of a giant robot", or like I did, "you are moving in a lift". Having an excess of rumble "because it can be done" is what I didn't like about the haptic implementation here, and I don't approve of this. You are not selling me that you will be using this.

0lvin:
Step is fully same story as jump and land.

protocultor:
The "land" sound plays after falling from a high altitude, not after a simple jump.
Running you do it all the time, "landing" is just a signal that you touched the ground without taking damage. It doesn't play constantly by just walking, like "step".

0lvin:
Have you tried to walk and check when step is happened?

0lvin:
I have replaced last commits with name filtering logic, player could filter sound files for haptic feedback, and if player likes to have some file or don't like annoying effect could add or remove such from list without code changes.

It has all effects from rumble list(e.g. quake and doors are also in list).

@protocultor @Yamagi could you please review?

0lvin:
I have changed filtering logic to use same space separator as in the image logic and with support of '!'. As result filter should behave fully same as rumble.

protocultor:
While this filter is much, much better than what you presented before, you are not going to get the same behaviour as rumble with name filtering alone. Which is OK, by the way.

One of the things that is missing from _rumble_ is that _rumble_ changes the maximum distance where a sound can be felt, depending on the "name" of the sound. This way, many `world/*` sounds are treated as "ambient", so can be felt almost anywhere on the map (`quake`, `rumble`), while others like `world/train2`, `doors/*` and specially `plats/*` can only be felt when very near from what produces it. On the contrary, _haptic_ establishes an universal maximum distance of 16, no more or less, for **any** sound, with no further filter other than the names in the cvar. It's very weird to feel plaforms, and specially doors, away from you making the controller vibrate so much.

What's more, _rumble_ uses a default maximum distance of 4, extremely close to the player, which allows to filter out a lot of sounds that otherwise would be played. The most evident example: `weapons/lashit.wav`, the sound of missed laser hits. Test the following: get in any situation against flyers (like the very beginning of the _rogue_ expansion), and dodge their fire. With _haptic_, you'll see that any shot they miss will make the controller vibrate, which repeats the situation of the bouncing grenades: it makes no sense to have haptics in those situations, the wall behind you is not a part of you. I assume you could just filter out this sound, but I don't know how many more sounds may have the same characteristics.

Still, it's very smart to have a filter for sounds for which you should NOT provide feedback. I have a question for you: can you see the reviews I left previously? If I check this page in another machine, where I'm not logged in in my Github account, I cannot see them. I'm asking because you haven't done anything about them, and I wondered if it was a fault of Github's.

DanielGibson:
IIRC when you do a review on Github (instead of posting single comments), you must somehow finish (or publish or whatever) that review for the comments to become visible (I forgot what the exact term is)

protocultor:
> IIRC when you do a review on Github (instead of posting single comments), you must somehow finish (or publish or whatever) that review for the comments to become visible (I forgot what the exact term is)

It was "submit review"! Thank you!

0lvin:
Initially code has 16 * 16 maximum effect distance. I have changed limit to use cvars and limit effect distance to 100. Engine generate effect of explode when player shout to barrel stay near same barrel with distance equal to 58 points.  So as result fire to barrel will have half of volume. And after two sizes of barrel will be no effect.

Could you please review?

src/client/input/sdl.c
```diff
@@ -1360,6 +1380,243 @@ In_FlushQueue(void)
...
+			effect_delay, effect_attack, effect_fade,
+			effect_x, effect_y, effect_z);
+
+		if (effect_id == -1)
+		{
+			/* have rumble used some slots in haptic effect list?,
+			 * ok, use little bit less haptic effects in same time*/
+			IN_Haptic_Effects_Shutdown();
+			last_haptic_effect_size --;
+			Com_Printf("%d haptic effects in same time\n", last_haptic_effect_size);
```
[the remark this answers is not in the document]
0lvin: Thank you. Fixed in both places.

src/client/input/sdl.c
```diff
@@ -1360,6 +1380,243 @@ In_FlushQueue(void)
...
+
+	if (effect_volume <= 0)
+	{
+		return;
+	}
+
+	if (effect_duration <= 0)
+	{
+		return;
+	}
```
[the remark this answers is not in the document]
0lvin: Thank you, merged to one operator.

src/client/input/sdl.c
```diff
@@ -1656,6 +1939,22 @@ IN_Controller_Init(qboolean notify_user)
...
+			}
+
+			if (joystick_haptic)
+			{
+				IN_Haptic_Effects_Info();
+				show_haptic = true;
+			}
+
 			show_gamepad = true;
 			Com_Printf("Enabled as Game Controller, settings:\n%s\n", SDL_GameControllerMapping(controller));
```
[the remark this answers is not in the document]
0lvin: Thank you, moved.

doc/040_cvarlist.md
```diff
@@ -275,6 +275,11 @@ Set `0` by default.
   `-1`: Disable reverb effect,
   `>=0`: select predefined effect.
 
+* **s_feedback_kind**: Select kind of controller feedback to use. By default
+  this cvar is `0`. Possibe values:
+  `0`: Rumble feedback,
+  `1`: Haptic feedback.
```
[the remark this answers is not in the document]
0lvin: Thank you, fixed in both places.

src/client/input/sdl.c
```diff
@@ -1360,6 +1382,379 @@ In_FlushQueue(void)
...
+	// "weapons/*x "        // explosions...
+	// "weapons/*r "        // reloads & ion ripper fire
+	// "weapons/*blastf "   // dampen blasters
+	// "weapons/*hyprbf "   // dampen blasters
+	// "weapons/*nail "     // nailgun's fire
+	// "weapons/*shotgf "   // shotgun
+	// "weapons/*rocklf "   // RL
+	// "weapons/*sshotf "   // for super shotgun
+	// "weapons/*machgf "   // machine gun & disruptor fire
+	// "weapons/*disint "   // machine gun & disruptor fire
```
protocultor, line 1561: I don't believe these comments serve a purpose here.
At line 1547, `"!weapons/*open "` should have the same comment as the line below: "rogue's items don't have feedback", since the `*open` sounds also come from rogue.

0lvin:
Thank you! @protocultor could please recheck?

doc/040_cvarlist.md
```diff
@@ -546,6 +546,17 @@ Set `0` by default.
   the controller alone. As these vary by device, it's better to use
   'calibrate' in the 'gamepad' -> 'gyro' menu to set them.
 
+* **joy_haptic_magnitude**: Haptic magnitude value, By default this cvar
+  is `0.0` or disabled. Valid values are 0..1.0.
```
protocultor, line 550: A valid value for this cvar can be over 1.0; before I changed the menu option (before rumble), the upper limit was 2.2, and today is 2.0, which of course continues to work for haptic. Maybe don't recommend an upper value? Or have 1.0 as a "recommended default" if you're going to use it?
It's not important, it's just for consistency; we can't tell the user "1.0 is the upper limit" if we offer 2.0/2.2 in the menu.

src/client/input/sdl.c
```diff
@@ -1360,6 +1382,371 @@ In_FlushQueue(void)
...
+	// any weapons that are not in previous list
+	"weapons/ "
+	// player{,s} effects
+	"player/*land "         // fall without injury
+	"player/*burn "
+	"player/*pain100 "
+	"player/*pain75 "
+	"player/*pain50 "
+	"player/*pain25 "
+	"player/*pain25 "
```
protocultor, line 1559: The last two are repeated. But you can leave this even more succint: by replacing all of these with `"player/*pain "`. You don't need to define every type of pain individually, since you are not specifying any difference between them.
A few lines below, in `players`, the same happens.

0lvin: Thank you, fixed

doc/040_cvarlist.md
0lvin: What do you think about updated sentence?

protocultor: I like it!

protocultor (approved):
I'm happy with the current state of things. Seems to have no negative impact on performance in a Raspberry Pi.

0lvin:
@protocultor Thank you!

src/client/input/sdl.c
```diff
@@ -1726,6 +2147,8 @@ IN_Init(void)
 	sensitivity = Cvar_Get("sensitivity", "3", CVAR_ARCHIVE);
 
 	joy_haptic_magnitude = Cvar_Get("joy_haptic_magnitude", "0.0", CVAR_ARCHIVE);
+	joy_haptic_distance = Cvar_Get("joy_haptic_distance", "100.0", CVAR_ARCHIVE);
+	haptic_feedback_filter = Cvar_Get("joy_haptic_filter", default_haptic_filter, CVAR_ARCHIVE);
```
Yamagi, line 2151: This cvar isn't documented. Or isn't it necessary at this point?

0lvin: Thank you, I have added description.

Yamagi:
Thank you for the very good work! :)

PR #947, opened 2022-12-11
Introduces new cvar cl_audiopaused to control if the audio loop

devnexen:
ought to go on or not when the game pauses.

### DIFF
--- a/src/client/sound/openal.c
+++ b/src/client/sound/openal.c
@@ -854,11 +854,12 @@ AL_AddLoopSounds(void)
 	int num;
 	entity_state_t *ent;
 
-	if ((cls.state != ca_active) || cl_paused->value || !s_ambient->value)
+	if ((cls.state != ca_active) || (cl_paused->value && cl_audiopaused->value) || !s_ambient->value)
 	{
 		return;
 	}
 
+	memset(&sounds, 0, sizeof(int) * MAX_EDICTS);
 	S_BuildSoundList(sounds);
 
 	for (i = 0; i < cl.frame.num_entities; i++)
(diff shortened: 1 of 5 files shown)


src/client/menu/menu.c
```diff
@@ -2365,6 +2373,13 @@ Options_MenuInit(void)
     s_options_quality_list.generic.callback = UpdateSoundQualityFunc;
     s_options_quality_list.itemnames = quality_items;
 
+    s_options_audiopause_box.generic.type = MTYPE_SPINCONTROL;
+    s_options_audiopause_box.generic.x = 0;
+    s_options_audiopause_box.generic.y = (y += 10);
+    s_options_audiopause_box.generic.name = "pause sound on pause";
```
Yamagi, line 2379: This overflows the menu on 320x280. And do we really want to use the last available line in the options menu for a rather obscure feature like this? I think having it as cvar only is good enough.

src/client/sound/openal.c
```diff
@@ -854,11 +854,12 @@ AL_AddLoopSounds(void)
 	int num;
 	entity_state_t *ent;
 
-	if ((cls.state != ca_active) || cl_paused->value || !s_ambient->value)
+	if ((cls.state != ca_active) || (cl_paused->value && cl_audiopaused->value) || !s_ambient->value)
 	{
 		return;
 	}
 
+	memset(&sounds, 0, sizeof(int) * MAX_EDICTS);
```
Yamagi, line 862: Why is this necessary?

Yamagi (changes requested):
I have added some comments.

src/client/sound/openal.c
devnexen: for matching the sdl backend mostly.

Yamagi:
Thank you :)

PR #952, opened 2023-01-05
Implement a simple stufftext filter

Yamagi:
I have received a report of ongoing attacks against various Quake II clients by malicious servers exploiting stufftexts. Stufftexts are random strings which are send by the server and directly fed into the clients command buffer. Even in 1997 this was a bad idea... It looks like so far only cvars getting altered, however given the state of Quake IIs command parser nearly everything seems possible.

This pull requests implements a simple stufftext filter. When `cl_stufftext` is set to `0` only known to be good stufftexts are processed. These are the basic commands required by the server at connection time and the `spectator` commands send by baseq2, rogue and xatrix. ctf doesn't send any stifftext. Since this will break all servers and mods which require random stufftext the default stays at `1`. Documented in the cvar list and the multiplayer setup guide.

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -146,6 +146,16 @@ Set `0` by default.
 * **cl_model_preview_end**: end frame value in multiplayer model preview.
   `-1` - don't show animation. Defaults to `94` for show salute animation.
 
+* **cl_stufftext**: If set to `1` (the default) any stufftext is
+  allowed. Since stufftext can inject arbitrary data into the client
+  command buffer this can be harmful. A malicious server may execute
+  any console command, alter any cvar or even hijack the client with
+  potentially grave consequences. If set to `0` only known to be good
+  messages are allowed. This reduces the attack surface significantly.
+  However this will break compatibility with any server and mod which
+  requires non-standard stufftext. And there is no guarantee that an
+  attacker can't bypass the filter.
+
 * **in_grab**: Defines how the mouse is grabbed by Yamagi Quake IIs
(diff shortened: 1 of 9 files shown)


maraakate:
There is more to just removing the "msg" CVAR.  You have to remove the check in SV_UserinfoChanged() and remove messagelevel from the client_t struct.  You need to keep the check for (cl->state != cs_spawned) however.

0lvin:
Just idea: commands could be filtered by logic from haptic feedback #944 or https://github.com/yquake2/ref_vk/commit/335b8855e20c24b795f07e86a191c443df8ecc04. What do you think?

Yamagi:
@DanielGibson and me had a longer discussion about this and decided to not merge the filter list at this point. I have cherry-picked the `msg` into master. The thing is, no matter from direction we are looking at the problem, there are are always aspects which are suboptimal: Filtering everything which isn't needed has a very high probability to break network play and foreign mods. Implementing a blacklist doesn't work, because everything is allowed and the only way to be save is a whitelist. I whitelist on the other hand requires a decision what should be allowed and what forbidden... Additionally there are a lot of special cases which aren't documented anywhere and must be taken into account do to it right.

And YQ2 isn't a focused on multiplayer. We always said that we are focusing on single player and a little bit of coop. Who wants to play multiplayer should use another client, q2pro is likely the best option. I will document that in out multiplayer guide, together with a hint that connecting to servers might be a security risk. If one of the multiplayer focused clients come up with a good(!) implementation I will merge it.

PR #970, opened 2023-01-30
silencing static analysis warnings.

devnexen:


0lvin:
Stb_* prefixed files are updated in sync with code version from https://github.com/nothings/stb as external libraries.

Yamagi:
Thank you :)

Issue #971, opened 2023-02-04
EFX reverb doesn't seem to work 


BielBdeLuna:
they work Corretcly on Dhewm3.

`s_reverb_preset` is set  to **-2** 
but if I test this or any of the presets they doesn't to engage on any reverb behaviour.

At the init I get the following:
<details><summary>click to toggle details</summary>
<p>

```
------- sound initialization -------
Loading library: libopenal.so.1
...opening OpenAL device: ok
...creating OpenAL context: ok
...making context current: ok

OpenAL settings:
AL_VENDOR: OpenAL Community
AL_RENDERER: OpenAL Soft
AL_VERSION: 1.1 ALSOFT 1.19.1
AL_EXTENSIONS: AL_EXT_ALAW AL_EXT_BFORMAT AL_EXT_DOUBLE AL_EXT_EXPONENT_DISTANCE AL_EXT_FLOAT32 AL_EXT_IMA4 AL_EXT_LINEAR_DISTANCE AL_EXT_MCFORMATS AL_EXT_MULAW AL_EXT_MULAW_BFORMAT AL_EXT_MULAW_MCFORMATS AL_EXT_OFFSET AL_EXT_source_distance_model AL_EXT_SOURCE_RADIUS AL_EXT_STEREO_ANGLES AL_LOKI_quadriphonic AL_SOFT_block_alignment AL_SOFT_deferred_updates AL_SOFT_direct_channels AL_SOFTX_events AL_SOFTX_filter_gain_ex AL_SOFT_gain_clamp_ex AL_SOFT_loop_points AL_SOFTX_map_buffer AL_SOFT_MSADPCM AL_SOFT_source_latency AL_SOFT_source_length AL_SOFT_source_resampler AL_SOFT_source_spatialize

Available OpenAL devices:
- Starship/Matisse HD Audio Controller Est  reo anal  gic
- TU106 High Definition Audio Controller Digital Stereo (HDMI 3)

Current OpenAL device:
- OpenAL Soft

Number of OpenAL sources: 31

Sound sampling rate: 0
------------------------------------
```

</p>
</details>

using last commit [2e939b10e9f3796d4d72e03e6b699c33c4d5579e]

Yamagi:
@0lvin: Can you have a look?

0lvin:
Could you please share whole log and os version?

>>  doesn't to engage on any reverb behaviour

Do you mean no sound difference on change effect, correct?

BielBdeLuna:
[...]

PR #972, opened 2023-02-05
openal: Fix reverb slot configuration

0lvin:
Fix effect slot configuration #971,
Call once set reverb for change slot configuration in underwater (not for each channel)


BielBdeLuna:
I've tested @0lvin reverb_fix branch and it works as intended! :+1:

Yamagi:
Very good. Thank you! :)

PR #979, opened 2023-03-16
further cmake build update using more cmake's heuristics instead

devnexen:


CMakeLists.txt
```diff
@@ -46,9 +46,9 @@ else() # GCC/clang/mingw
 #  -fwrapv              -> Make signed integer overflows defined
 #  -fvisibility=hidden	-> Force defaultsymbol visibility to hidden
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -fno-strict-aliasing -fwrapv -fvisibility=hidden")
-
 # Use -O2 as maximum optimization level. -O3 has it's problems with yquake2.
-string(REPLACE "-O3" "-O2" CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE}")
+set(CMAKE_BUILD_TYPE RelWithDebInfo)
```
Yamagi, line 50: While RelWithDebInfo defaults to `-O2` on most platforms it's not guaranteed. It isn't even guaranteed that RelWithDebInfo has lower optimization level than Release. Therefore we can't rely on RelWithDebInfo not setting -O3.

DanielGibson:
does this fix a specific problem?

otherwise I'd prefer not to mess with the build settings.

devnexen:
It catches up with [here](https://github.com/yquake2/yquake2/blob/master/Makefile#L166).

DanielGibson:
not sure why that's there, but let's not

devnexen:
what if I conceal only for unix build, seems your concern is about windows ?

Yamagi:
I have solved this this other way round. 3a9b8de2aba683184fdd9f3e20e8153354e882f5 removed the C99 enforcement from the Makefile.

Issue #729, opened 2021-07-18
[Feature] Sequential playback mode starting from default level music (like in PSX)


lafoxxx92:
When entering the level, its default song is played, which is good (The track is basically "level theme").
It's Repeated when Shuffle is disabled, which is not good (same music track all over gets too irritating).
When Shuffle is enabled, default song no longer plays when you enter the level, which is bad  (The default track is basically "level theme").

Suggestion:
Add new play mode ("Sequential") -- play default level track, but when it ends -- play Next track (for example -- Track05 after Track04, etc) (like in PSX Quake)
or
Always play default track when entering the leve, but Next track will be Random (Shuffle is enabled)

apartfromtime:
I've implemented this in pull request #986, and added an additional "play once" feature along with the requested "sequential" feature, "shuffle" remains unchanged except renamed to "random" in the menu control and now any two tracks should not repeat. I've tested all the features with the base quake2 soundtrack, I don't have the xatrix or rogue soundtracks so I'll have to defer testing of those to anyone that does.

protocultor:
Just in case, your PR does seem to work with Xatrix, with the soundtrack in GOG format (Track02.ogg to Track21.ogg in their own "music" directory).

Yamagi:
I have merged the pull request. I test build with the changes included can be found here: https://deponie.yamagi.org/quake2/windows/testbuilds/quake2-20230416-f67b65c.zip

apartfromtime:
Could probably close this issue now.

PR #986, opened 2023-03-26
Ogg Playback Features

apartfromtime:
Playback features outlined in issue #729.

Added shuffle playback parameters to the menu:

- default - Ogg track currently active repeats. This is the quake2 default behaviour.
- play once - Ogg track currently active plays once then stops.
- sequential - Ogg tracks play in numerical order, from the currently playing track and like default, repeats.
- random - Ogg tracks play randomly, but never the same track twice.

Ogg tracks can be played from a full-screen console and tracks played will adhere to the shuffle parameter.

Loading a game map the map cd-track takes precedence as the first played track then subsequent playback of tracks will adhere to the shuffle parameter.

Any currently playing track can be overridden from the console with "`ogg play <track>`" command and subsequent playback of tracks will adhere to the shuffle parameter.

If a sound restart occurs the ogg backend will attempt to save and recover the currently playing track, though some data in the audio queue may be lost in the process.

### DIFF
--- a/src/client/cl_view.c
+++ b/src/client/cl_view.c
@@ -363,8 +363,8 @@ CL_PrepRefresh(void)
 
 	/* start the cd track */
 	int track = (int)strtol(cl.configstrings[CS_CDTRACK], (char **)NULL, 10);
-
-	OGG_PlayTrack(track);
+	
+	OGG_PlayTrack(track, true, true);
 }
 
 float
--- a/src/client/sound/ogg.c
+++ b/src/client/sound/ogg.c
@@ -47,6 +47,7 @@ static cvar_t *ogg_volume;        /* Music volume. */
 static int ogg_curfile;           /* Index of currently played file. */
 static int ogg_numbufs;           /* Number of buffers for OpenAL */
 static int ogg_numsamples;        /* Number of sambles read from the current file */
+static int ogg_mapcdtrack;        /* Index of current map cdtrack */
 static ogg_status_t ogg_status;   /* Status indicator. */
 static stb_vorbis *ogg_file;      /* Ogg Vorbis file. */
 static qboolean ogg_started;      /* Initialization flag. */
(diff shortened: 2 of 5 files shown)


src/client/cl_view.c
```diff
@@ -363,8 +363,8 @@ CL_PrepRefresh(void)
 
 	/* start the cd track */
 	int track = (int)strtol(cl.configstrings[CS_CDTRACK], (char **)NULL, 10);
-
-	OGG_PlayTrack(track);
+
```
Yamagi, line 366: Here are some superfluous whitespaces.

src/client/sound/ogg.c
```diff
@@ -323,6 +324,24 @@ OGG_Stream(void)
...
+	{
+		// Shuffle was modified before last track finished.
+		ogg_shuffle->modified = false;
+	}
+	else if (ogg_status == STOP && ogg_shuffle->modified)
+	{
+		// If fullscreen console, clear the map cdtrack.
+		if (cls.state == ca_disconnected)
+		{
+			ogg_mapcdtrack = 0;			// Track 0 means "stop music".
```
Yamagi, line 338: Does this work as intended with `ogg_ignoretrack0 1`?

Yamagi:
I have added two smaller comments. Besides that this looks very good.

src/client/cl_view.c
apartfromtime: I'll try to find what is causing this in my text editor.

src/client/sound/ogg.c
apartfromtime: Yes. That statement is to clear the cd track if the user had played a map and disconnected, then changed any of the playback parameters through the console\menu after playback of the last track had finished.

Yamagi: Ah, okay.

PR #990, opened 2023-03-29
Gyro aiming for Switch controllers with SDL < 2.0.14

protocultor:
Since @Yamagi can have "controversial features as optional features", I wanted to do this controversial PR.

https://github.com/nicman23/dkms-hid-nintendo exposes the IMU sensors of a Switch controller as a "second joystick", which makes possible to use gyro aiming on a system without having SDL 2.0.14 available (the minimum required to read controller sensors like gyro and accelerometer). This PR allows to use gyro aiming in a Raspberry Pi, or any other system with an old SDL version, by reading this "second joystick".

"Sensor reading" is still the preferred method when available, and when the time for "SDL 3 only" comes, this still makes clear which method must remain.

### DIFF
--- a/src/client/input/sdl.c
+++ b/src/client/input/sdl.c
@@ -174,7 +174,7 @@ static cvar_t *gyro_calibration_z;
 #if SDL_VERSION_ATLEAST(2, 0, 14)	// support for controller sensors (gyro, accelerometer)
 
 static unsigned int num_samples;
-#define NATIVE_SDL_GYRO 1	// uses SDL_CONTROLLERSENSORUPDATE to read gyro
+#define NATIVE_SDL_GYRO	// uses SDL_CONTROLLERSENSORUPDATE to read gyro
 
 #else	// for SDL < 2.0.14, gyro can be read as a "secondary joystick" exposed by dkms-hid-nintendo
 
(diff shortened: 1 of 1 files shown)


src/client/input/sdl.c
```diff
@@ -167,12 +167,25 @@ static qboolean gyro_active = false;
...
-static unsigned int num_samples;
-
 static cvar_t *gyro_calibration_x;
 static cvar_t *gyro_calibration_y;
 static cvar_t *gyro_calibration_z;
 
+#if SDL_VERSION_ATLEAST(2, 0, 14)	// support for controller sensors (gyro, accelerometer)
+
+static unsigned int num_samples;
+#define NATIVE_SDL_GYRO 1	// uses SDL_CONTROLLERSENSORUPDATE to read gyro
```
0lvin, line 177: ''''c
#define NATIVE_SDL_GYRO
''''
Should be enough

protocultor: Thanks!

Yamagi:
> Since @Yamagi can have "controversial features as optional features", I wanted to do this controversial PR.

And I'll merge it. ;) I can't test it, because I don't have Switch controllers but the code looks good, compiles fine and has little impact. Keep up the good work!

Issue #995, opened 2023-04-19
[Original bug] Stand ground gladiators have a no-attack zone


BjossiAlfreds:
Discovered this one by accident, and it's in the original game as well. Gladiators have a small dead zone around them where they won't attack at all. This is intended to make fighting chasing gladiators a bit more fair for players, but for stand ground gladiators this is an AI glitch since the player can choose to stand still inside the dead zone and have a stare competition with the gladiator.

I can provide a PR where I add an AI_STAND_GROUND exception to the relevant code.

Yamagi:
> I can provide a PR where I add an AI_STAND_GROUND exception to the relevant code.

Another nice find that probably no one else has made in the last 25 years. :) Please do a PR for it.

PR #997, opened 2023-04-23
Fixed stand-ground gladiators not attacking within a certain range

BjossiAlfreds:
Fixes issue https://github.com/yquake2/yquake2/issues/995.

### DIFF
--- a/src/game/monster/gladiator/gladiator.c
+++ b/src/game/monster/gladiator/gladiator.c
@@ -325,13 +325,19 @@ gladiator_attack(edict_t *self)
 		return;
 	}
 
-	/* a small safe zone */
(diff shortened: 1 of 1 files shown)


abalfoort:
Shouldn't this be implemented in [Zaero](https://github.com/yquake2/zaero/blob/master/src/monster/gladiator/gladiator.c) as well?

BjossiAlfreds:
> Shouldn't this be implemented in [Zaero](https://github.com/yquake2/zaero/blob/master/src/monster/gladiator/gladiator.c) as well?

There's a lot of stuff that hasn't been ported there yet, but since you asked I'll do a Zaero version of this.

PR #998, opened 2023-04-23
Fix implicit declaration of SDL_GameControllerGetSensorDataRate avail…

apartfromtime:
…able since SDL 2.0.16 not 2.0.14

### DIFF
--- a/src/client/input/sdl.c
+++ b/src/client/input/sdl.c
@@ -171,12 +171,12 @@ static cvar_t *gyro_calibration_x;
 static cvar_t *gyro_calibration_y;
 static cvar_t *gyro_calibration_z;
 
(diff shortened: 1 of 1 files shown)


protocultor:
That's a good catch, I completely forgot about it. However, I don't agree on the way you fixed it: you deny functionality available on 2.0.14 to users because of the use of a single function, that is only used to show (not-that-important) info to the user. I know that's what I did originally, but I believe there's a better way:
```
#if SDL_VERSION_ATLEAST(2, 0, 16)
			Com_Printf( "Gyro sensor enabled at %.2f Hz\n",
					SDL_GameControllerGetSensorDataRate(controller, SDL_SENSOR_GYRO) );
#else
			Com_Printf( "Gyro sensor enabled." );
#endif
```

PR #999, opened 2023-04-23
addressing few gcc 13 analyser warnings.

devnexen:


src/backends/unix/system.c
```diff
@@ -575,18 +575,22 @@ void
...
-			Sys_Remove(filepath);
-		}
+			while ((file = readdir(directory)) != NULL)
+			{
+				snprintf(filepath, MAX_OSPATH, "%s/%s", path, file->d_name);
+				Sys_Remove(filepath);
+			}
 
-		closedir(directory);
+			closedir(directory);
```
Yamagi, line 592: This one should be in the `if (directory)` scope. If we can't open the directory, we likely can't remove it. And without having a look into the directory we cannot be sure that it's empty.

Issue #889, opened 2022-09-07
Makron cannot be gibbed


ghost:
The two halves of Makron's corpse cannot be gibbed because:

1) The torso is spawned in as a separate entity that cannot take damage.
2) `gib_health` is set to -2000, but `Killed()` contains a check that prevents health values below -999. 

Because `gib_health` is set for the Makron, I assume that this is a bug, not a feature.

The solution for (1) (borrowed from kmquake2) is to make the torso able to take damage and add a function `makron_torso_die` for gibbing the torso. (2) can be solved by adding an exception based on `targ->classname` in `Killed()`. Both worked for me.

If you want I can do a PR for this.

BjossiAlfreds:
Another thing to think about with (1) is that it's spawned at a fixed origin relative to the legs, so there is no bound checking. Maybe increase the size of the hitbox (but not bigger than the Makron itself) and trace from the legs opposite the facing direction?

As for (2) I'd just remove that check entirely. I did that over 10 years ago in my mod and never looked back. You can then add a -99 lower cap to the health value displayed by the HUD so that -100 and above don't get cut off.

widmak:
Would 60x60x30 be large enough for the hitbox? Also, my understanding of the issue is a bit fuzzy, but doesn't `makron_die()` already spawn the torso at a fixed distance opposite the facing direction of the legs?

If that check is indeed only there to ensure that negative health values fit on the HUD, removing it and changing the HUD health display code as you suggest would make more sense than adding an extra check.

BjossiAlfreds:
I use -24x-24x0 to 24x24x16, I find that works really well.

The origin offset is axis aligned, it's not calculated from the makron's forward vector. And also, there is no bounds checking, so the torso will happily spawn inside a wall if the destination happens to be that.

My code looks like this, feel free to use it if you want.

static void set_torso_origin (const edict_t *self, edict_t *torso)
{
	vec3_t	v;
	trace_t	tr;

	AngleVectors (self->s.angles, v, NULL, NULL);
	VectorMA (self->s.origin, -84.0f, v, v);

	tr = gi.trace (self->s.origin, torso->mins, torso->maxs, v, self, MASK_SOLID);

	VectorCopy (tr.endpos, torso->s.origin);
}

Yamagi:
> If you want I can do a PR for this.
Yes, please do! :)

PR #1005, opened 2023-05-03
Fixed ungibbable makron parts

BjossiAlfreds:
This PR addresses issue https://github.com/yquake2/yquake2/issues/889.

The following changes are made:
* Removed the -999 check in Killed function to make the makron legs gibbable. Its purpose was for HUD display. Instead I added a -99 cap in the HUD code, which also fixes confusing display of negative health that is between -100 and -989, where playrs would only see the first 2 digits.

* Made the makron torso shootable and gibbable

* Makron torso is no longer spawned in a fixed location relative to the legs, now it gets traced. This prevents it from spawning inside walls.

* Makron torso is affected by gravity, so if it spawns in the air it will fall down.

* Adjusted makron legs' obscenely large hitbox to match its visual appearance better

I have tested the following scenarios:
1. Killing the makron and gibbing the parts
2. Killing the makron, saving, loading the save and gibbing the parts
3. Killing the makron right up against a wall to test the new trace code
4. Walk into a laser to test the -99 HUD display cap

All seems to work for me, but I invite others to verify.

### DIFF
--- a/src/game/g_combat.c
+++ b/src/game/g_combat.c
@@ -127,11 +127,6 @@ Killed(edict_t *targ, edict_t *inflictor, edict_t *attacker,
 		return;
 	}
 
-	if (targ->health < -999)
-	{
-		targ->health = -999;
-	}
-
 	targ->enemy = attacker;
 
 	if ((targ->svflags & SVF_MONSTER) && (targ->deadflag != DEAD_DEAD))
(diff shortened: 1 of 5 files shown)


Yamagi:
I gave it a short try and as far as I can tell it works fine.

For this and all the other PRs: Thank you for them. It's very appreciated. I'm always amazed what strange corner cases you find in the game. :)

Issue #1009, opened 2023-05-11
[Original bug] Makron's jump out of the jorg suit is broken


BjossiAlfreds:
This one is probably well known to most seasoned Q2 players and has a number of things that are broken.

1. The makron does his jump move but there is no actual velocity with it
2. The makron forgets all about his enemy while leaving the suit
3. The makron does the jump move upon seeing the player, even if the suit is long gone

I also ran into a 4th issue specific to YQ2:

4. The jorg suit's gibs block the makron's jump velocity

I have a working jump sequence so I'll do a PR soon that you guys can try out, and will explain the bugs and the fixes there in more detail.

BjossiAlfreds:
No problem, I'll replicate this in the other games soon-ish.

PR #1010, opened 2023-05-12
Various fixes to makron jump sequence

BjossiAlfreds:
This PR addresses issue https://github.com/yquake2/yquake2/issues/1009.

Makron had no velocity applied because id's code didn't add 1 to the origin, the physics code decided the makron was touching the ground regardless of his upward velocity (might be a bug in its own right?).

The 2nd one was easily fix by copying enemy and oldenemy to the makron entity in MakronToss, and then set up the chase in MakronSpawn.

The 3rd problem was fixed by moving makron_move_sight from makron_sight to MakronSpawn. It doesn't make any sense for him to do this jump except when he's jumping out of the jorg suit.

I'm still unsure why YQ2 changed gibs to SOLID_BBOX as it causes insta-despawning gibs like with quadded RL, but it was blocking the makron's jump velocity so as a compromise fix I set gibs to SVF_DEADMONSTER.

I did make 3 notable changes to id's code:
* I normalize the upward velocity with sv_gravity. This is because the jump frameset is fixed length, and maps are generally designed for him to always have the same jump height regardless of gravity. Plus, Q2 definitely doesn't have the infrastructure for complex frame control, would just get confusing and cause further edge case issues.

* The makron was using level.sight_client as his reference for angle of the jump. Doesn't really make any sense so instead I have him jump towards his enemy if that enemy is still alive and visible, otherwise he jumps in the direction he was facing upon death.

* Gave him 1 second of pain resistance at the start of the jump, just so the jump move is allowed to mostly complete.

By all means give this branch a try and give your thoughts. Just know that the jump frames are easy to send out of sync, like if you shoot him with the railgun while he's airborne. I prefer not freezing the frames while in the air because it can cause the AI to soft-lock if he lands in a weird spot or the player can keep him airborne. So I opted to keep things crude but functional rather than try something too fancy.

### DIFF
--- a/src/game/g_misc.c
+++ b/src/game/g_misc.c
@@ -218,6 +218,7 @@ ThrowGib(edict_t *self, char *gibname, int damage, int type)
 
 	gi.setmodel(gib, gibname);
 	gib->solid = SOLID_BBOX;
+	gib->svflags = SVF_DEADMONSTER;
 	gib->s.effects |= EF_GIB;
 	gib->flags |= FL_NO_KNOCKBACK;
 	gib->takedamage = DAMAGE_YES;
--- a/src/game/monster/boss3/boss32.c
+++ b/src/game/monster/boss3/boss32.c
@@ -788,12 +788,6 @@ makron_pain(edict_t *self, edict_t *other /* unused */,
 void
 makron_sight(edict_t *self, edict_t *other /* unused */)
 {
-	if (!self)
-	{
-		return;
-	}
-
-	self->monsterinfo.currentmove = &makron_move_sight;
 }
 
 void
@@ -1170,7 +1164,7 @@ SP_monster_makron(edict_t *self)
 	self->monsterinfo.dodge = NULL;
 	self->monsterinfo.attack = makron_attack;
 	self->monsterinfo.melee = NULL;
-	self->monsterinfo.sight = makron_sight;
+	self->monsterinfo.sight = NULL;
 	self->monsterinfo.checkattack = Makron_CheckAttack;
 
 	gi.linkentity(self);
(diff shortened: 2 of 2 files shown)


Yamagi:
> I'm still unsure why YQ2 changed gibs to SOLID_BBOX as it causes insta-despawning gibs like with quadded RL, but it was blocking the makron's jump velocity so as a compromise fix I set gibs to SVF_DEADMONSTER.

That was a long time ago. If I remember right it was an attempt to fix gibs not being transported by conveyor belts and everything abusing the `func_conveyor` for fancy effects. Especially notable in the factory. And yes, looking back it was bad idea because it introduced a number of new problems.

I'll give it a try tomorrow. I'm currently at my laptop and this needs a real keyboard + mouse setup to test.

BjossiAlfreds:
Hmm that's interesting. I never really paid much attention to the conveyors.

As far as I can tell func_conveyor is unused in the game. If you look at its code it doesn't even do anything, just sets a BSP model and that's it. It messes around with self->speed and self->count but does nothing with those values, which are both private to the game module.

I'm guessing the conveyor effect is being done through the content flags of the conveyor surface?

Yamagi:
Hmm... The original commits were  07477e0f75707ebcdbadf243ae7e88555f0065b2 and b4d16ab6b3e040adf1764b7d0c87a57c48616f31. The rational behind them was that entities marked `SOLID_NOT` aren't moved by whatever is used to implement the conveyors under the meat grinders in fact2 and some other places in the game. Looking at the entity file for fact2 I take an educated guess and speculate that the conveyors are `func_trains`.

Yamagi:
> By all means give this branch a try and give your thoughts. Just know that the jump frames are easy to send out of sync, like if you shoot him with the railgun while he's airborne. I prefer not freezing the frames while in the air because it can cause the AI to soft-lock if he lands in a weird spot or the player can keep him airborne. So I opted to keep things crude but functional rather than try something too fancy.

I gave it a try. I would say that there are two ways looking at this issue. Rather inexperienced players will likely never encounter the problems and aren't able to exploit them. For more seasoned players the fix breaks the 'kill jork -> hide behind the walls -> hit macro once or even twice while he's jumping' strategy or makes it somewhat more challenging.

Nevertheless, I would say we should merge this change. Because Macron "freezing" until he sees the player for the first time is illogical and I always found the initial jump confusing. And it makes Jork / Macron at least a little bit more challenging. He's way to easy to kill, especially when considering that he is the final boss and (the|one) evil mastermind behind the Strogg.

abalfoort:
Can this also be implemented in zaero?

BjossiAlfreds:
> Can this also be implemented in zaero?

Here you go: https://github.com/yquake2/zaero/pull/35

PR #1011, opened 2023-05-14
fix potential overflow on env vars settings.

devnexen:
changing a bit Q_strlcpy flow as it can report wrong bytes copied if src > buffer size.

### DIFF
--- a/src/client/cl_main.c
+++ b/src/client/cl_main.c
@@ -270,13 +270,13 @@ CL_Setenv_f(void)
 		char buffer[1000];
 		int i;
 
-		strcpy(buffer, Cmd_Argv(1));
-		strcat(buffer, "=");
+		Q_strlcpy(buffer, Cmd_Argv(1), sizeof(buffer));
+		Q_strlcat(buffer, "=", sizeof(buffer));
 
 		for (i = 2; i < argc; i++)
 		{
-			strcat(buffer, Cmd_Argv(i));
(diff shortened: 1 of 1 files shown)


src/common/shared/shared.c
```diff
@@ -1120,14 +1120,10 @@ Q_strlcpy(char *dst, const char *src, int size)
 {
 	const char *s = src;
 
-	while (*s)
+	while (--size > 0)
```
DanielGibson, line 1123: fails for size=0

DanielGibson: (what's wrong with the original implementation anyway?)

devnexen: > fails for size=0

true even tough only used with compile time buffers it seems.

devnexen: > (what's wrong with the original implementation anyway?)

I realise ... is those functions trying to detect truncations ?

e.g.

`
char buf[5];
Q_strlcpy(buf, "abcdef", sizeof(buf));
`

will return 6. 

if that s the case I will revert the change.

DanielGibson: they're supposed to behave like https://www.freebsd.org/cgi/man.cgi?query=strlcpy&sektion=3

devnexen: clear enough

PR #1004, opened 2023-04-29
Sound backend and ogg pause features

apartfromtime:
Playback features outlined in issue #715 .

Reintroduce activate\deactivate sound backend when window focus in changed. Added "windowed_pauseonfocuslost" cvar to allowing user control of pause when the window changes focus. Added "ogg_pausewithgame" cvar to allow user control of ogg playback whenever the game is paused. Introduced ogg playback command "mute" as an additional playback option for the console, command buffer and key binds.

If the window focus is changed all sound playback will stop and resume as expected. Setting "window_pauseonfocuslost" to 1 will force the game to pause when the window focus is lost, "cl_paused" is ignored in multiplayer. Setting "ogg_pausewithgame" to 1 will force ogg playback to pause whenever the game is paused. Toggling ogg mute through "ogg mute" sets the volume to 0 while playback resumes.

PR #1017, opened 2023-05-24
Pause on Minimized/Focus Lost is now in the options menu

### LINKED ISSUE #951: game pauses when  changing  between vid_fullscreen  0  <-> 1
probably introduced  by commit  https://github.com/yquake2/yquake2/commit/d93916d6ba86ff9f226cfe6394d1f96bce0d1dde and https://github.com/yquake2/yquake2/issues/911 

Please unpause when window/screen is visible again, or introduce a  CVAR  to disable  completely, Default  pausing disabled. :)

Rykita:
Hopefully closes #951 , ~~alternatively you could add a third option that unpauses when the window regains focus but I decided to just make it an override for simplicity's sake.~~ Added with [cd351a0](https://github.com/yquake2/yquake2/pull/1017/commits/cd351a06023ecbbbb7e2a99d40684a65688a0a3f) 

There are three options:
Unpause on re-focus
Pause
Don't pause at all

All are available in the regular "Options" menu.

`windowed_pauseonfocuslost` is now ~~`w_pauseonfocuslost`~~ ``vid_pauseonfocuslost`` to make it easier/faster/to type by the player and more consistent with other similar options.

### DIFF
--- a/src/client/input/sdl.c
+++ b/src/client/input/sdl.c
@@ -693,7 +693,7 @@ IN_Update(void)
 					{
 						S_Activate(false);
 
-						if (windowed_pauseonfocuslost->value != 2)
+						if (windowed_pauseonfocuslost->value != 1)
 						{
 						    Cvar_SetValue("paused", 1);
 						}
@@ -710,7 +710,7 @@ IN_Update(void)
 					{
 						S_Activate(true);
 
-						if (windowed_pauseonfocuslost->value == 0)
+						if (windowed_pauseonfocuslost->value == 2)
 						{
 						    Cvar_SetValue("paused", 0);
 						}
@@ -744,12 +744,6 @@ IN_Update(void)
 					{
 					    Cbuf_AddText("ogg toggle\n");
 					}
-
-				}
-				else if (event.window.event == SDL_WINDOWEVENT_MINIMIZED ||
-					event.window.event == SDL_WINDOWEVENT_HIDDEN)
-				{
-					Cvar_SetValue("paused", 1);
 				}
 				break;
 
@@ -2336,7 +2330,7 @@ IN_Init(void)
 		gyro_active = true;
 	}
 
-	windowed_pauseonfocuslost = Cvar_Get("w_pauseonfocuslost", "0", CVAR_USERINFO | CVAR_ARCHIVE);
+	windowed_pauseonfocuslost = Cvar_Get("vid_pauseonfocuslost", "0", CVAR_USERINFO | CVAR_ARCHIVE);
 	windowed_mouse = Cvar_Get("windowed_mouse", "1", CVAR_USERINFO | CVAR_ARCHIVE);
 
 	Cmd_AddCommand("+mlook", IN_MLookDown);
--- a/src/client/input/sdl.c
+++ b/src/client/input/sdl.c
@@ -732,8 +732,8 @@ IN_Update(void)
 						Cvar_SetValue("paused", 0);
 					}
 				}
-				else if (event.window.event == SDL_WINDOWEVENT_MINIMIZED ||
-					event.window.event == SDL_WINDOWEVENT_HIDDEN)
+				else if (event.window.event == SDL_WINDOWEVENT_MINIMIZED && windowed_pauseonfocuslost->value > 0 ||
+					event.window.event == SDL_WINDOWEVENT_HIDDEN && windowed_pauseonfocuslost->value > 0)
 				{
 					Cvar_SetValue("paused", 1);
 				}
@@ -2322,7 +2322,7 @@ IN_Init(void)
 		gyro_active = true;
 	}
 
-	windowed_pauseonfocuslost = Cvar_Get("windowed_pauseonfocuslost", "0", CVAR_USERINFO | CVAR_ARCHIVE);
+	windowed_pauseonfocuslost = Cvar_Get("w_pauseonfocuslost", "0", CVAR_USERINFO | CVAR_ARCHIVE);
 	windowed_mouse = Cvar_Get("windowed_mouse", "1", CVAR_USERINFO | CVAR_ARCHIVE);
 
 	Cmd_AddCommand("+mlook", IN_MLookDown);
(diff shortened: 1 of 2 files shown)


apartfromtime:
The cvars probably should come under "vid_" but the abbreviated "w_" is acceptable.

You shouldn't need to track if it is minimised or not, only if it has gained or lost focus. Minimise pushes a focus lost event.

I've pushed a fix to my branch, and fixed some ogg playback features that I missed/made in #1004, that also have to be updated to work with the window pause features. If you could review those changes and add them to your code changes.

be1e6e1743d9d5ad7fb7fbf395df6155435bd58f

Yamagi:
Let's not waste the last remaining slot in the options menu for a feature most users will likely never change. Just set a sane default (unpause on re-focus?) and document the cvars in https://github.com/yquake2/yquake2/blob/master/doc/040_cvarlist.md. Maybe add a sentence or two to the documentation guide under https://github.com/yquake2/yquake2/blob/master/doc/030_configuration.md, but that might not be necessary.

Also, how was this tested? Especially Linux with things like tiling window managers and interesting implementations of virtual desktops everything which is related to window events has a high probability to break or not to work in ways the user expects.

apartfromtime:
So I went ahead and tested this on Windows and Linux. I tested all combinations of exclusive full-screen, windowed full-screen and window, with `pause on minimized: yes|no|unpause`, and `cl_unpaused_scvis: 0|1`, in both the intro loop and game. Of course these results will need to be verified by someone else, but there are some differences.

Linux works as expected except in exclusive full-screen with `pause on minimized: no` and `cl_unpaused_scvis: 0`, the expected result should be un-paused on return but the intro and game remain paused,
`pause on minimized: no` should override `cl_unpaused_scvis: 0` in this case. Of note is Linux always minimizes the game window (at least in my testing) whenever alt-tabbing.

Windows is different in several more cases, in exclusive full-screen and windowed full-screen both exhibit the same behaviour, with `pause on minimized: yes` and `cl_unpaused_scvis: 1`, the expected result is un-paused on return but the result is paused in in both intro and game. With `pause on minimized: no` and `cl_unpaused_scvis: 0 or 1` the expected result is unpaused on return but the result is paused. In windowed, only `pause on minimized: unpaused` and `cl_unpaused_scvis: 0 or 1` returns the expected result in the intro. In a game `pause on minimized: yes|no|unpause`, and `cl_unpaused_scvis: 0|1` work as expected.

A couple of things to note, I recommended changing the variable name for `windowed_pauseonfocuslost` to something like `vid_pauseonfocuslost` or `wnd_pauseonfocuslost`, these should also be renamed wherever they occur in the source so that the variable gets updated whenever it changes value and any maintainers can find all occurrences with a search of the variable name, at present they may not be getting updated properly and `unpause on re-focus` could just be `unpause` and convey the same information in the context of `pause on minimized`.

Rykita:
> Let's not waste the last remaining slot in the options menu for a feature most users will likely never change. Just set a sane default (unpause on re-focus?) and document the cvars in https://github.com/yquake2/yquake2/blob/master/doc/040_cvarlist.md. Maybe add a sentence or two to the documentation guide under https://github.com/yquake2/yquake2/blob/master/doc/030_configuration.md, but that might not be necessary.
> 
> Also, how was this tested? Especially Linux with things like tiling window managers and interesting implementations of virtual desktops everything which is related to window events has a high probability to break or not to work in ways the user expects.

Sorry, I was away for a while; I haven't tested it on Linux unfortunately; but it should work fine as apartfromtime commented, on another note I'll make a commit that removes it from the menu and documents it in the docs instead per request.

src/client/input/sdl.c
```diff
@@ -2322,7 +2330,7 @@ IN_Init(void)
 		gyro_active = true;
 	}
 
-	windowed_pauseonfocuslost = Cvar_Get("windowed_pauseonfocuslost", "0", CVAR_USERINFO | CVAR_ARCHIVE);
+	windowed_pauseonfocuslost = Cvar_Get("vid_pauseonfocuslost", "0", CVAR_USERINFO | CVAR_ARCHIVE);
```
Yamagi, line 2333: `vid_pauseonfocuslost` is still missing documentation in the General section in https://github.com/yquake2/yquake2/blob/master/doc/040_cvarlist.md

src/client/menu/menu.c
```diff
@@ -146,6 +139,12 @@ M_PopMenu(void)
     if (!m_menudepth)
     {
         M_ForceMenuOff();
+	    /* play music */
+	    if (Cvar_VariableValue("ogg_pausewithgame") == 1 &&
+	        OGG_Status() == PAUSE && cl.attractloop == false)
+	    {
+	        Cbuf_AddText("ogg toggle\n");
+	    }
```
Yamagi, line 147: The indention is wrong.

src/client/menu/menu.c
```diff
@@ -2432,6 +2447,13 @@ Options_MenuInit(void)
     s_options_crosshair_box.generic.name = "crosshair";
     s_options_crosshair_box.generic.callback = CrosshairFunc;
     s_options_crosshair_box.itemnames = crosshair_names;
+	
+	s_options_pauseonfocus_box.generic.type = MTYPE_SPINCONTROL;
```
Yamagi, line 2451: Wrong indention

src/client/menu/menu.c
```diff
@@ -2475,6 +2497,7 @@ Options_MenuInit(void)
     Menu_AddItem(&s_options_menu, (void *)&s_options_lookstrafe_box);
     Menu_AddItem(&s_options_menu, (void *)&s_options_freelook_box);
     Menu_AddItem(&s_options_menu, (void *)&s_options_crosshair_box);
+	Menu_AddItem(&s_options_menu, (void*)&s_options_pauseonfocus_box);
```
Yamagi, line 2500: Wrong indention.

Yamagi (commented):
I have added come comments inline. Besides them it's okay.

It would still be nice if someone can test these changes with a tiling window manager, but if no one steps up I'm going to merge it anyways. They can write bug reports if something breaks. ;)

Yamagi:
Ping?

Yamagi:
I have merged this by hand and corrected the indention problems. I have also documented the `vid_pauseonfocuslost` cvar.

PR #1029, opened 2023-06-10
hunk api enrichment and micro optimisation on R_SetFrustum

devnexen:


DanielGibson:
all this code to get rid of *one* memset() (or, actually, move it to another function)?

devnexen:
> all this code to get rid of _one_ memset() (or, actually, move it to another function)?

right, a bit overengineered indeed, removed.

turol:
Typo in commit message

Yamagi:
Thank you :)

PR #1030, opened 2023-06-17
adding compile time assertion support whenever possible

devnexen:
and convert some runtime which deserve to be evaluated earlier.


src/common/header/shared.h
```diff
@@ -77,6 +78,8 @@ typedef unsigned char byte;
 	#define YQ2_ATTR_NORETURN       __attribute__ ((noreturn))
 	#define YQ2_ATTR_MALLOC         __attribute__ ((__malloc__))
 	#define YQ2_ATTR_INLINE         __attribute__((always_inline)) inline
+    	// GCC supports this extension since 4.6
```
DanielGibson, line 81: indenting is inconsistent with surrounding lines

DanielGibson:
good idea generally, though I'd call it `YQ2_STATIC_ASSERT`

devnexen:
I know but I tried to avoid long names :-) but I ll change it no pb.

Yamagi:
Thank you :)

PR #1031, opened 2023-06-28
Cheat to speed up "weapon change" animations

protocultor:
I know gameplay related changes are questioned here, but I think this is a very popular issue with Quake 2, because one of the most criticized aspects of it (even among Quake series' fans) is the speed which you can change weapons. And since Quake 2 Remastered is near ([leaked by the Korean Game Rating and Administration Committee, just like with Q1 Remastered in 2021](https://www.gematsu.com/2023/06/quake-ii-remastered-rated-in-korea)), I bet that a similar option to speed up weapon switching will be included there.

Code is based on [Lithium II](https://github.com/mattayres/li2mod) and [Q2 25th Anniversary](https://github.com/glhrmfrts/q25_game). Although I say so myself, this PR has a less disruptive, better organized implementation of "fast weapon switching" than these 2 mods.

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -222,6 +222,13 @@ Set `0` by default.
   single player, the same way as in multiplayer.
   This cvar only works if the game.dll implements this behaviour.
 
+* **g_swap_speed**: Sets the speed of the "changing weapon" animation.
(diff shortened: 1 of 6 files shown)


Yamagi:
That's a very good idea. Thank you for implementing it.

PR #1033, opened 2023-07-08
Fixed VectorNormalize2 not writing 0-length vectors to out

BjossiAlfreds:
Noticed that VecorNormalize2 does not output a vector if the length is 0 which is just asking for undefined behavior (my T_Damage fix made the false assumption in fact, that out would always be modified).

Its code was also redundant so I made it call the 1-arg variant to carry out the normalization.

I added these changes into the other 3 games through the T_Damage fixes.

Btw, optimization thought: Couldn't we move the sqrt call behind the length != 0 check? Would avoid sqrt for blank vectors like vec3_origin.

### DIFF
--- a/src/common/shared/shared.c
+++ b/src/common/shared/shared.c
@@ -521,20 +521,9 @@ VectorNormalize(vec3_t v)
 vec_t
 VectorNormalize2(vec3_t v, vec3_t out)
 {
-	float length, ilength;
+	VectorCopy(v, out);
 
-	length = v[0] * v[0] + v[1] * v[1] + v[2] * v[2];
-	length = (float)sqrt(length);
-
-	if (length)
-	{
-		ilength = 1 / length;
-		out[0] = v[0] * ilength;
-		out[1] = v[1] * ilength;
-		out[2] = v[2] * ilength;
-	}
-
-	return length;
+	return VectorNormalize(out);
 }
 
 void


apartfromtime:
> Btw, optimization thought: Couldn't we move the sqrt call behind the length != 0 check? Would avoid sqrt for blank vectors like vec3_origin.

I check for zero in the vector, rather than checking for zero length, if one is non-zero the result will be non-zero. But I've not looked at the assembly produced or run any tests with two compares (potentially one compare) of "`x != 0 || y != 0`" then calculate versus calculate then one compare of "`length !=0`".

BjossiAlfreds:
My gut feeling tells me zero tests are very fast, and the vectors are cache friendly (continuous in memory and a compact 12 bytes).

How much these tests save in CPU time really depends on how often blank vectors go through there versus non-blank ones. There is also the question of short circuit effect, if most vectors are non-blank then only 1 test would be performed.

Probably thinking way too hard about this haha. Premature optimization and all that.

Yamagi:
> Probably thinking way too hard about this haha. Premature optimization and all that.

It's not like Quake II is CPU limited in any way on modern hardware...


> Btw, optimization thought: Couldn't we move the sqrt call behind the length != 0 check? Would avoid sqrt for blank vectors like vec3_origin.

It should be okay. I gave it a short try. The game still seems to work fine and there's no measurable performance difference on my Ryzen 3700X. There might be a small gain on older hardware, though.

I'm merging this and the T_Damage() changes in the addons. Thank you :)

PR #1037, opened 2023-08-05
fix cmake assumption on x86 arch.

devnexen:
set wrongly on i386 even tough the cpu is x86`_64`.


Yamagi:
This might break some savegames. If that's a case a new game must be started or the ARCH field in the savegame file can be rewritten with a hex editor.

Thank you for this bugfix :)

Issue #1039, opened 2023-08-10
[Feature Request] Remaster Assets/Pak


d10sfan:
Quake II recently was updated with a remaster (https://store.steampowered.com/news/app/2320/view/3644027577415996831?l=english), that includes a new expansion among other things.

Is it possible for this project to add support for the new content? Inside the steam/gog release, there's a rerelease/baseq2/pak0.pak file (that's it for pak files), trying to launch from this directory, yquake2 complains about `Error: Couldn't load pics/conchars.pcx`

Yamagi:
Supporting the remastered game data would be nice. However on very, very first look it seems like the changed a lot of things. Including gamecode changes and alteration to the assets. This might not be easy to implement so don't get your hopes too high. It would be easier if they release their source code...

And there's the question how compatible Vanilla Quake II and Quake II Remastered are in terms of network protocol and mods. Since YQ2 need to stay compatible with Vanilla, it might be impossible to dully support Quake II Remaster.

0lvin:
```
Custom maps designed for the PC versions of the original Quake II should still work with the enhanced release of the game.
```
It could be that in opposite directions it will be possible, new game library could have same API as old one, and yquake2 could load new game.dll on windows.

But in same time I don't see mention of Linux and freebsd in release notes.

OpenRift412:
> It would be easier if they release their source code...

Well we do have this: https://github.com/id-Software/quake2-rerelease-dll

Yamagi:
That's really cool. Even better is that they finally released the unaltered source code of the addons under GPL2.  Until now the addons were under the crappy SDK License, which wasn't really an open source license and made it impossible to distribute them together with the client. I need to ponder if relicensing our versions is possible.

> The rerelease of Quake II uses a new version of the API to communicate between the server & the game module. It also introduces a very thin "client game" module, akin to Quake III Arena's cgame module, to allow for extended modding opportunities that change previously hardcoded client behavior. It also has a new network protocol, version 2023.

It won't "just work", because YQ2 doesn't support this new API version. But with the extensible documentation (many thanks to Nightdive Studios, that's much better than just a source code dump!) it should be possible to implement the changes in YQ2 or maybe separate YQ2 Remastered branch.

I'm currently moving to a new home and don't have much time. This could be a nice project for long winter nights.

d10sfan:
[...]

PR #1040, opened 2023-08-12
Support Quake2 ReRelease

0lvin:
Fixed for support ReRelease path changes #1039 

Use PNG as replace of pcx automatically, generate colormap and 16to8 if information found(mostly required for soft render)

Rework of nolerp filtering, logic are same as for haptic feedback

Vulkan support https://github.com/yquake2/ref_vk/pull/46

### DIFF
--- a/src/client/refresh/soft/sw_image.c
+++ b/src/client/refresh/soft/sw_image.c
@@ -691,14 +691,17 @@ R_InitImages (void)
 	registration_sequence = 1;
 	image_max = 0;
 
+	R_Printf(PRINT_ALL, "%s() Load colormap\n", __func__);
+
+	GetPCXPalette (&vid_colormap, (unsigned *)d_8to24table);
+	vid_alphamap = vid_colormap + 64*256;
+
 	d_16to8table = NULL;
 	ri.FS_LoadFile("pics/16to8.dat", (void **)&table16to8);
 
 	if ( !table16to8 )
 	{
-		ri.Sys_Error(ERR_FATAL, "%s: Couldn't load pics/16to8.dat", __func__);
-		// code never returns after ERR_FATAL
-		return;
+		R_Printf(PRINT_ALL, "%s: Couldn't load pics/16to8.dat", __func__);
 	}
 
 	d_16to8table = malloc(0x10000);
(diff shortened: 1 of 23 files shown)


src/client/refresh/soft/sw_image.c
```diff
@@ -708,8 +711,57 @@ R_InitImages (void)
...
+					for (i = 0; i < 256; i ++)
+					{
+						int r_c, g_c, b_c;
+						float curr_diff;
+
+						r_c = d_8to24table[i * 4 + 0] - (r << 3);
+						g_c = d_8to24table[i * 4 + 1] - (g << 2);
+						b_c = d_8to24table[i * 4 + 2] - (b << 3);
+
+						curr_diff = sqrt(
```
turol, line 743: There's no need to sqrt here since you only care about their relative values and not the actual result.

0lvin: Thank you, fixed

0lvin:
Looks like change is enough to run without critical errors.

Known bugs in render code:
* dark/no light in soft render
* bright lighting in GL1/GL3/VK renders (looks like no shadows at all)
* Slow rotating of sky

@Yamagi could you please review?

0lvin:
I will port lights fixes from https://github.com/yquake2/ref_vk/pull/47 for map when I will found way to fix light for models.

0lvin:
I have added additional small fix for palette [r_retexturing: use rgb with pcx if image has builtin palette](https://github.com/yquake2/yquake2/pull/1040/commits/eb89968211876c9b4c805c169778562a38d7e58f).

It resolved case when image is palette based but has different palette than default one:

Before with Daikatana colormap:
daikatana_colormap

Before with Quake2 colormap:
quake2_colormap

After fix:
reuse_colormap

0lvin:
@Yamagi @DanielGibson What do you think about add Daikatana image format? It will take ~ 211 lines of code change. Look to https://github.com/0lvin/yquake2/tree/rerelease-combined as example.

Yamagi:
Moving home is done and I'm back in the game :) I will look into this into this in the next few days.

0lvin:
Thank you, I have created one additional #1055. It's not finished.

Code has only GL1 implementations, I have tested code with base1 and mguhub. It has no regression for old maps as I know.

Yamagi:
I have created a new repository for the remastered version: https://github.com/yquake2/yquake2remaster I think it's better to keep the remastered support in a fork until we either have developed it to a point were we can decide if it can be merged back into YQ2 itself or should stay a fork. This way we don't need to be too overcautious with breaking things.

I have already merged this PR into the new repo. @0lvin, I'll give you write access to the new repository.

0lvin:
Thank you! 

I will merge ref_vk to it also, just to have single source repository without additional branch in ref_vk for ReRelease with autorotate parameter in API.

PR #1049, opened 2023-08-27
Fix for g_swap_speed behaviour with extreme values

protocultor:
The cvar added in https://github.com/yquake2/yquake2/pull/1031 presents undesirable behavior, whether the value of the cvar is between 0 and 1, or if it exceeds 65535.
In both cases, internally it is treated as if its real value is 0, which breaks the weapon animation, making it stuck.
The present PR solves this.

In addition to this:

- Zoid Kirsch's name in credits is now consistent with his credit in the recent "enhanced" version ([source](https://twitter.com/ZoidCTF/status/1542699504041750528)).

- Replacement cvar for "gl_lightmap" is now correct. Also, I took the trouble to include typo corrections.


Yamagi:
Thank you :)

PR #1052, opened 2023-09-02
CURl client fix UAF download queue on removal.

devnexen:


DanielGibson:
Good find!

I think the fix is a bit hard to follow though; counterproposal: https://github.com/yquake2/yquake2/pull/1061

Yamagi:
I have merged #1061.

PR #1057, opened 2023-09-15
Fixed Linux compile error by adding missing header

thefakeryker:
Can someone test on windows to see if any problems occur?

### DIFF
--- a/src/game/player/weapon.c
+++ b/src/game/player/weapon.c
@@ -26,6 +26,7 @@
 
 #include "../header/local.h"
 #include "../monster/misc/player.h"
+#include "limits.h"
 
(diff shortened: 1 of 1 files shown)


turol:
limits.h is a system header and so should be included with angle brackets like so:
```
#include <limits.h>
```

protocultor:
> limits.h is a system header and so should be included with angle brackets like so:
> 
> ```
> #include <limits.h>
> ```

Yep, this is the actual solution. My apologies, should have tested my PR in more platforms.

Yamagi:
I have merged #1059 with the same fix. Thank you.

PR #1058, opened 2023-09-15
macOS: Fix compilation of weapon.c

atsb:
On the current master branch, weapon.c fails to compile on macOS due to the following:

```
===> CC src/game/player/weapon.c
src/game/player/weapon.c:609:26: error: use of undeclared identifier 'USHRT_MAX'
                (g_swap_speed->value < USHRT_MAX)? (unsigned short int)g_swap_speed->value : 1
                                       ^
1 error generated.
```

This is because on macOS, the include file 'limits.h' must be present for the compiler to find this definition.

Tested on both an Intel Mac running BigSur and Apple Silicon running the latest.  This PR fixes the compilation.

### DIFF
--- a/src/game/player/weapon.c
+++ b/src/game/player/weapon.c
@@ -24,9 +24,7 @@
  * =======================================================================
  */
 
-#ifdef __APPLE__
 #include <limits.h>
-#endif
 
 #include "../header/local.h"
 #include "../monster/misc/player.h"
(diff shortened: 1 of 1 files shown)


protocultor:
The problem also occurs in Linux: https://github.com/yquake2/yquake2/pull/1058
...so you can take out the #ifdef and it will work in all platforms.
My apologies.

atsb:
> The problem also occurs in Linux: #1058
> ...so you can take out the #ifdef and it will work in all platforms.
> My apologies.

Awesome!  Done, nice and simple 👍

Yamagi:
I have merged #1059. Thank you.

PR #1060, opened 2023-09-15
"constifying" where underlying C calls expect compile time constants.

devnexen:


Yamagi:
You need to also constify the function pointer in the game and renderer APIs, otherwise it will give warnings like this:

```
src/server/sv_game.c:397:17: warning: incompatible function pointer types assigning to 'void (*)(int, char *, ...)' from 'void (int, const char *, ...)' [-Wincompatible-function-pointer-types]
        import.bprintf = SV_BroadcastPrintf;
```

Yamagi:
Thank you :)

Issue #1065, opened 2023-10-13
How is "gl1_overbrightbits" supposed to work?


protocultor:
[Documentation](https://github.com/yquake2/yquake2/blob/master/doc/040_cvarlist.md#graphics-opengl-14-only) says its value can go from 0 to 3, with '2' and '3' being factors for the lightmaps, but from the maps with lava I've tested ("city1" and "boss1", at least), '3' looks more like '1', with the difference that your own weapon, enemies, and other non-static brushes are brighter than your surroundings. '2' is a lot brighter than both, and '4' is almost ridiculous, with everything looking near fullbright. Numbers beyond that make everything look like '3'.

My question comes from this bit on [RI_BeginFrame](https://github.com/yquake2/yquake2/blob/master/src/client/refresh/gl1/gl1_main.c#L1616):
```        // Clamp overbrightbits
        if (gl1_overbrightbits->modified)
        {
                if (gl1_overbrightbits->value > 2 && gl1_overbrightbits->value < 4)
                {
                        ri.Cvar_Set("r_overbrightbits", "2");
                }
                else if (gl1_overbrightbits->value > 4)
                {
                        ri.Cvar_Set("r_overbrightbits", "4");
                }

                gl1_overbrightbits->modified = false;
        }
```
Since `"r_overbrightbits"` doesn't exist, I assume it's an old name for `gl1_overbrightbits`, and this must be corrected. But doing this removes "3" as a possible value, which both contradicts the documentation and takes out an useful value. On the other hand, using "3" may be considered "an advantage" to some, thanks to the contrast it provides, and that's why it wasn't originally available.

What do you think?

Yamagi:
`r_overbrightbits` is likely an oversight from splitting the overbright bits implementation between GL1 and GL3. The overbright bits in GL1 are a compromise, because implementing them without shaders is hard. I would say: Fix the logic in a way that the values 0, 1, 2 and 3 are possible. One could argue that 3 is too bright, but supporting it doesn't hurt and maybe it helps visual impaired players.

protocultor:
Just in case: the one that is too bright is 4, which isn't even mentioned in documentation as a possible value, and is clearly something that `"r_overbrightbits"` allowed.
In order of "less bright" to "fullbright", values are: **0**, **1**, **3**, **2**, **4**.
[...]

PR #1067, opened 2023-10-18
Fixed gl1_overbrightbits limiter

protocultor:
As discussed in #1065, allowed values for this cvar are between 0 and 4, but now the game actually enforces these limits. Documentation was updated to reflect this, and the difference between `3` and `4`.

P.D.: sorry about the last commit ;)


Yamagi:
Thank you :)

PR #1076, opened 2023-12-08
Faster "weapprev" and "weapnext" behavior

protocultor:
These commands can now "count" how many times they have been called. This allows for changing to different weapons when quickly tapping their keys (or scrolling the mouse wheel), instead of going just "one down" or "one up" at a time. This is similar to how these commands operate on the recent rerelease of the game, or the old PS1 version of Q2.
The code for this is based on #865, where I did the same for `cycleweap`.

A new cvar, `g_quick_weap`, allows to enable or disable this behavior. By default, it is disabled.

This has been tested on Windows and Raspberry Pi OS.

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -222,6 +222,12 @@ Set `0` by default.
   single player, the same way as in multiplayer.
   This cvar only works if the game.dll implements this behaviour.
 
+* **g_quick_weap**: If set to `1`, both *weapprev* and *weapnext*
+  commands will "count" how many times they have been called, making
+  possible to skip weapons by quickly tapping one of these keys.
+  By default this cvar is set to `0`, and will only work if the
+  game.dll implements this behaviour.
+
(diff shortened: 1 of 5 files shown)


Yamagi:
Hi, that's a cool idea. The weapon switch behavior has always been annoying. What about expending this a little and switch the ammunition icon in the HUD everytime the weapon switch count increases? This way the user can see in the hud what weapon would be selected if they stop incrementing the counter.

protocultor:
Good idea. Just give me a few days to implement that.

protocultor:
So much for the "_few days..._", it was pretty easy to implement :) .
Weapon icon and name stay in the HUD for 0.9 seconds, I believe is visible and practical enough. If you don't think so, just tell.

Yamagi:
> Weapon icon and name stay in the HUD for 0.9 seconds, I believe is visible and practical enough. If you don't think so, just tell.

It's perfekt. Better than my suggestion to abuse the ammo icon.

PR #1077, opened 2023-12-12
Fix duplicate menu key keyboard entries

apartfromtime:
Fixes a bug where duplicate keyboard entries would occur in menu fields when using the keypad keys with num-lock on.
Menu fields now mirror behaviour of the console field.
Removed keypad left arrow case from console key ignore list.

### DIFF
--- a/src/client/menu/menu.c
+++ b/src/client/menu/menu.c
@@ -242,31 +242,31 @@ Key_GetMenuKey(int key)
 	switch (key)
 	{
 		case K_KP_UPARROW:
-                    if (IN_NumpadIsOn() == true) { break; }
+		    if (IN_NumpadIsOn() == true) { break; }
 		case K_UPARROW:
 		case K_DPAD_UP:
-			return K_UPARROW;
+		    return K_UPARROW;
 
 		case K_TAB:
 		case K_KP_DOWNARROW:
-                    if (IN_NumpadIsOn() == true) { break; }
+		    if (IN_NumpadIsOn() == true) { break; }
 		case K_DOWNARROW:
 		case K_DPAD_DOWN:
-			return K_DOWNARROW;
+		    return K_DOWNARROW;
 
 		case K_KP_LEFTARROW:
                     if (IN_NumpadIsOn() == true) { break; }
 		case K_LEFTARROW:
 		case K_DPAD_LEFT:
 		case K_SHOULDER_LEFT:
-			return K_LEFTARROW;
+		    return K_LEFTARROW;
 
 		case K_KP_RIGHTARROW:
-                    if (IN_NumpadIsOn() == true) { break; }
+		    if (IN_NumpadIsOn() == true) { break; }
 		case K_RIGHTARROW:
 		case K_DPAD_RIGHT:
 		case K_SHOULDER_RIGHT:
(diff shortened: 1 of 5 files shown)


Yamagi:
Hi,

that is nice find :) This solution has some unwanted implications, though. First, the code distinguishes between character events and key events because in SDL 1.2 and most likely in the Vanilla Quake 2 backends for Win9x and X11 these were different kind of events. SDL2 got away with that and has only key events. In a good world YQ2 would use only key events and no character events, text input would be implemented through SDL2s input events. Input events are an abstraction over key events. In practice no one ever had the motivation to modernize the code accordingly and most likely it wouldn't be a good idea either. The chance to introduce new bugs or uncover hidden corner cases is extremely high.

Your approach alters the behavior of the numpad for the whole game. The key bindings are now dependent on the state of the num key which at least messes with binding and may break some cases like numblocks embedded into the normal part of the keyboard or some strange keyboard layout.

And the console behavior gets inconsistent. Numpad left is now allowed in the console, numpad up and down aren't. I would prefer not to mess with the console because the console code is broken, fragile and the change has potential to uncover hidden bugs or design limitations.

What about another approach: Add a function `Sys_NumpadIsOn()` or something like that to the input backend and leave it otherwise alone. Use that function to fix the menu inside the menu code.

apartfromtime:
`Key_GetMenuKey()` introduced a subtle bug by remapping inputs. I've implemented your suggestion, this fixes the duplicate entries, the keypad directional keys work in the menu as one would expect and any key-binds, bind the keypad key not the numpad character.

src/client/menu/menu.c
```diff
@@ -283,12 +287,18 @@ Key_GetMenuKey(int key)
 		case K_BACKSPACE:
 		case K_DEL:
 		case K_KP_DEL:
+                    if (IN_NumpadIsOn() == true) { break; }
```
Yamagi, line 290: Not 100% sure but it seems like the indention is wrong.

src/client/menu/qmenu.c
```diff
@@ -223,51 +223,31 @@ extern int keydown[];
...
+        case K_KP_5:
+        case K_KP_RIGHTARROW:
+        case K_KP_END:
+        case K_KP_DOWNARROW:
+        case K_KP_PGDN:
+        case K_KP_INS:
+        case K_KP_DEL:
+        default:
+            break;
+    }
```
Yamagi, line 250: The switch-case statement is now superfluous and can be deleted.

Yamagi (changes requested):
This looks now much better. Thank you for your changes. I have still two smaller points. I think that it it's ready for merging after these have been resolved.

src/client/menu/menu.c
apartfromtime: Weird one, displays correctly in the editors, Notepad, Sublime Text, Visual Studio, but tabs are broken when posting to GitHub. Very annoying. Vim resolved this.

src/client/menu/qmenu.c
apartfromtime: Deleted.

Yamagi (approved):

Yamagi:
Now it perfekt. Thank you for your hard work!

PR #1079, opened 2023-12-22
fix frags player stats uninitialized bytes.

devnexen:


BjossiAlfreds:
While this doesn't hurt, I don't see where the uninitialized part of the array is being accessed.

devnexen:
true, I misread, qsort affects only a subset of..

PR #1081, opened 2024-01-06
Fixed gl1_overbrightbits limiter (for real now)

### LINKED ISSUE #1080: `gl1_overbrightbits` allows for invaild value
Setting `gl1_overbrightbits` to `3` causes opengl errors in calls to

`glTexEnvi(GL_TEXTURE_ENV, GL_RGB_SCALE_EXT, gl1_overbrightbits->value);`

> According to docs:
> If the pName parameter is GL_RGB_SCALE_EXT, or GL_ALPHA_SCALE,
> the Parameter(s) parameter is (or points to) a floating-point scale factor.
> Only three such scale factors are valid: 1.0, 2.0, and 4.0. The default value is 1.0.

Possible fix is to replace this line https://github.com/yquake2/yquake2/blob/e40f5d14804b40ae38a2972a5638954103735483/src/client/refresh/gl1/gl1_main.c#L1623 with `else if (gl1_overbrightbits->value > 2)`

Edit: Tested on NVIDIA GTX1650, Windows 10

protocultor:
Allowed values are now `0`, `1`, `2` and `4` only, since values like "3" and "0.5" generate OpenGL errors.

About the latter, these errors can finally be visible, by using the only supported method on old OpenGL: by calling `glGetError()` manually.
We do not have available something like `glDebugMessageCallbackARB` here, so by using some preprocessor magic we call `glGetError()` after every OpenGL function. Because of the ridiculous overhead this produces, this is only available when compiling with `DEBUG=1`.
At least, now there's a practical way to see if something's wrong with OpenGL 1.

This is based on the first method described here:
https://learnopengl.com/In-Practice/Debugging
You can test it by commenting the `gl1_overbrightbits` validator of the second commit, and set it to "3", "0.7", "8" or any other wrong value.

Now, the only issue remaining is that the GL1 "overbrightbits" slider in the video options is limited from 0 to 2.
That's because the new validator will avoid setting it to `3`, which means its maximum value of `4` can never be reached.
Typing it in console (or a script) is the only way to get it to `4` now.

This fixes #1080.

P.S.: just as an extra, gamepad buttons are read in a more succinct, efficient manner now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,7 +144,7 @@ endif
 # Highest supported optimizations are -O2, higher levels
 # will likely break this crappy code.
 ifdef DEBUG
-CFLAGS ?= -O0 -g -Wall -pipe
+CFLAGS ?= -O0 -g -Wall -pipe -DDEBUG
 ifdef ASAN
 override CFLAGS += -fsanitize=address -DUSE_SANITIZER
 endif
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -453,9 +453,8 @@ Set `0` by default.
 * **gl1_overbrightbits**: Enables overbright bits, brightness scaling of
   lightmaps and models. Higher values make shadows less dark. Possible
   values are `0` (no overbright bits), `1` (more correct lighting for
-  water), `2` (scale by factor 2), `3` (scale lighting by 3 only for the
-  dynamic meshes, like enemies and items), and `4` (scale lighting of
-  everything by 4). Applied in realtime, does not need `vid_restart`.
+  liquids), `2` (scale lighting by factor 2), and `4` (scale by factor
+  4). Applied in realtime, does not need `vid_restart`.
 
 * **gl1_particle_square**: If set to `1` particles are rendered as
   squares.
--- a/src/client/header/keyboard.h
+++ b/src/client/header/keyboard.h
@@ -220,8 +220,6 @@ enum QKEYS {
 	K_STICK_RIGHT,
 	K_SHOULDER_LEFT,
 	K_SHOULDER_RIGHT,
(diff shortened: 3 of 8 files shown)


Yamagi:
I gave it a try on various GPUs, all on Linux with more or less the latest driver:

* As before, on my Radeon setting `3` with the validator commented out doesn't trigger an OpenGL. Neither into the console, nor when build with DEBUG through `glGetError()`. Since `3` looks the same as `2` radeonsi seem to clamp the value to the next lower supported value.
* On the crappy GeForce in my Thinkpad setting `3` with the validator commented out corrupts the OpenGL state, when build with DEBUG an error is reported through `glGetError()`.
* On the Intel GPU in my Thinkpad setting `3` with the validator commented out doesn't do anything, when build with DEBUG an error is reported through `glGetError()`.

Experience shows that OpenGL behavior greatly differs between GPUs, even from the same vendor, driver versions and platforms. So this may vary if anything is changed.

I played through the first unit with DEBUG enabled, there don't seem any other OpenGL errors. At least not with more or less standard settings. While we'll hopefully never need it, having a way to call `glGetError()` after every API call is nice and may help with feature bugs. Or if somebody is crazy enough to implement new features in GL1.

> Now, the only issue remaining is that the GL1 "overbrightbits" slider in the video options is limited from 0 to 2.
> That's because the new validator will avoid setting it to 3, which means its maximum value of 4 can never be reached.

I think that is okay. The menu is limited and we already have options which work only partly through the menu. As an example YQ2 supports wullscreen windows with resolution != desktop resolution. That can only be configured through the console.

> just as an extra, gamepad buttons are read in a more succinct, efficient manner now.

Nice :)

PR #1083, opened 2024-01-21
Case insensitive sort for player skins

### LINKED ISSUE #1082: Case insensitive sort for player skins
Right now, in "Player Setup", lowercase player skins are separated from uppercase.
Instead, they should be sorted case insensitive.
This has always been a problem for Quake II since it was new and it'd be nice to finally fix it.

apartfromtime:
Case insensitive compare in quicksort compare function for player names.

Closes #1082

### DIFF
--- a/src/client/menu/menu.c
+++ b/src/client/menu/menu.c
@@ -5650,7 +5650,7 @@ PlayerModelList(void)
         }
 
         // sort skin names alphabetically
-        qsort(s_skinnames[mdl].data, s_skinnames[mdl].num, sizeof(char**), Q_sort_strcomp);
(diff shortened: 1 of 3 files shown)


Yamagi:
We can't change `Q_sort_strcomp()` globally, because it's also used in the command parser. What about adding another compare function `Q_sort_stricmp()` instead?

apartfromtime:
Amended the commit.

Perhaps mods should also be ordered in this way?

Yamagi:
> Perhaps mods should also be ordered in this way?

That's a doof idea. Done in 5933b64f778a9388cd89d0132f3924914e5a034d,

And many thanks for this PR. :)

PR #1085, opened 2024-02-11
Texture filtering option

### LINKED ISSUE #753: [Feature] Texture filtering option
It would be nice to have a toggle in video settings which lets you switch between GL_NEAREST_MIPMAP_LINEAR/VK_MIPMAP_NEAREST and GL_LINEAR_MIPMAP_LINEAR/VK_MIPMAP_LINEAR. Could be labelled "Texture Filtering", with the options named "Linear" and "Trilinear" (or even simpler: "Pixelated" and "Smooth").

apartfromtime:
Adds texture filtering option to the video menu.

Closes #753

### DIFF
--- a/src/client/menu/videomenu.c
+++ b/src/client/menu/videomenu.c
@@ -68,6 +68,7 @@ static menulist_s s_fs_box;
 static menulist_s s_vsync_list;
 static menulist_s s_af_list;
 static menulist_s s_msaa_list;
+static menulist_s s_filter_list;
 static menuaction_s s_defaults_action;
 static menuaction_s s_apply_action;
 
@@ -177,6 +178,23 @@ ResetDefaults(void *unused)
 
 #define CUSTOM_MODE_NAME "[Custom    ]"
 #define AUTO_MODE_NAME   "[Auto      ]"
+
+static void
+ApplyFilter(void* unused)
+{
+	if (s_filter_list.curvalue == 0)
+	{
+		Cvar_Set("gl_texturemode", "GL_NEAREST");
+	}
+	else if (s_filter_list.curvalue == 1)
+	{
+		Cvar_Set("gl_texturemode", "GL_LINEAR_MIPMAP_NEAREST");
+	}
+	else if (s_filter_list.curvalue == 2)
+	{
+		Cvar_Set("gl_texturemode", "GL_LINEAR_MIPMAP_LINEAR");
+	}
+}
 
 static void
 ApplyChanges(void *unused)
@@ -398,6 +416,14 @@ VID_MenuInit(void)
 		"4x",
 		"8x",
 		"16x",
+		0
+	};
+
+	static const char *filter_names[] = {
+		"pixelated",
+		"standard",
+		"trilinear",
+		"custom",
 		0
 	};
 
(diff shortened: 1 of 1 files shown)


protocultor:
I have yet to test your branch, but reading the code, I'm not convinced on your "ratings", the name you put on every type of texture filtering option. From what I read, you have:

_pixelated_ = GL_NEAREST
_balanced_ = GL_NEAREST_MIPMAP_LINEAR
_smooth_ = GL_LINEAR_MIPMAP_LINEAR

And according to https://learnopengl.com/Getting-started/Textures ("Mipmaps" section), these mean:

**GL_NEAREST**: no filtering, no mipmaps.
**GL_NEAREST_MIPMAP_LINEAR**: linearly interpolates between the two mipmaps that most closely match the size of a pixel and samples the interpolated level with no filtering.
**GL_LINEAR_MIPMAP_LINEAR**: same as above but the interpolated level is sampled linearly filtered.

Basically, the first two levels will look the same on most cases; people will have to look at the same texture at a distance to see the difference. Also, there's no option for the default for `gl_texturemode`: **GL_LINEAR_MIPMAP_NEAREST**, what was called "bilinear filtering" back in the day, which takes the nearest mipmap level for the texture, but samples that level using linear interpolation.

I would do this:

_pixelated_ (or _unfiltered_) = GL_NEAREST_MIPMAP_NEAREST
_balanced_ (or _bilinear_) = GL_LINEAR_MIPMAP_NEAREST	(will appear as the default)
_smooth_ (or _trilinear_) = GL_LINEAR_MIPMAP_LINEAR

...and people who don't want to use mipmaps can change `gl_texturemode` manually.
If you want to test all options possible, I recommend to disable anisotropic filtering, to be able to see the differences more clearly.

apartfromtime:
> GL_NEAREST_MIPMAP_LINEAR: linearly interpolates between the two mipmaps that most closely match the size of a pixel and samples the interpolated level with no filtering.

That's not correct.

From the OpenGL specification:

**GL_NEAREST_MIPMAP_LINEAR**:
Chooses the two mipmaps that most closely match the size of the pixel being textured and uses the GL_NEAREST criterion (the texture element nearest to the center of the pixel) to produce a texture value from each mipmap. The final texture value is a weighted average of those two values.

The returned pixel is the weighted average (bilinear filtered) of the two texels.

Also you're only considering the min filter in your naming convention, and never the mag filter. The min filter GL_LINEAR_MIPMAP_NEAREST and mag filter of GL_LINEAR (as implemented in quake2) produces visible banding between texture detail. Where as a min filter of GL_NEAREST_MIPMAP_LINEAR and mag filter of GL_NEAREST, produces a balanced result, with no visible banding, which is the reason behind naming it 'balanced'.

The re-master implements this as texture smoothing, nearest or trilinear and and it is either on or off.

I haven't changed any defaults, so whatever users current `'gl_texturemode'` value is, then gets mapped to one of the 3 strings "pixelated", "balanced" or "smooth" and then a `'custom'` string is set for those who manually change `'gl_texturemode'`.

I tested all combinations available in quake2 (and some that aren't) with anisotropic filtering off.

I'll wait until you and others have tested this, before making any further changes.

Yamagi:
I like this change. I pixelated look is one of the more often requested things, some players even use the software renderer for it.

The default is still ` GL_LINEAR_MIPMAP_NEAREST`. We never changed it, however I set it to `GL_LINEAR_MIPMAP_LINEAR` since forever. Testing the modes on my 32" 4k display, I'm unable to see a difference between `GL_NEAREST` and `GL_NEAREST_MIPMAP_LINEAR`. The difference between `GL_LINEAR_MIPMAP_NEAREST` and `GL_LINEAR_MIPMAP_LINEAR` is minimal. If any.

I don't see a point in providing two pixelated modes. I think we gain more by providing one pixelated mode, bilinear filtering (the default) and optional trilinear filtering.

What about:
* Pixelated (`GL_NEAREST`)
* Standard (`GL_LINEAR_MIPMAP_NEAREST`)
* Trilinear (`GL_LINEAR_MIPMAP_LINEAR`)

apartfromtime:
Amended the commit.

Yamagi:
I'm sorry for the delay. Thank your for the changes. :)

PR #1088, opened 2024-03-02
Added a player speed display that prints the players speed and ground speed at the top right corner. The function can be toggled on using cl_showspeed 1.

ConHuevosGuey:
I wasn't quite sure how to implement it without adding GetPlayerSpeed() to pmove.c. I'm unsure if there is a cleaner way to do this without accessing pmove.c while keeping it inside cl_screen.c. I also wasn't sure if the SCR_AddDirtyPoint() calls were necessary inside of SCR_DrawSpeed(), it seemed to work fine without them but decided to err on the side of caution in the case that it aids performance.

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -142,7 +142,12 @@ Set `0` by default.
 
 * **cl_showfps**: Shows the framecounter. Set to `2` for more and to
   `3` for even more informations.
-
+  
+* **cl_showspeed**:  Shows the players speed.  Set to `1` to display both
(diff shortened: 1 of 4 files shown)


Yamagi:
Thank you for the pull request. The code looks very good :) Especially the idea to print the seed under the crosshair is nice. Can you add a short description of `cl_showspeed` to https://github.com/yquake2/yquake2/blob/master/doc/040_cvarlist.md? Afterwards I can merge it.

ConHuevosGuey:
Sure no problem, done.

PR #1091, opened 2024-03-18
Texture filtering options for OpenGL and Software video modes

apartfromtime:
Add renderer specific menu options.

### DIFF
--- a/src/client/menu/videomenu.c
+++ b/src/client/menu/videomenu.c
@@ -182,17 +182,33 @@ ResetDefaults(void *unused)
 static void
 ApplyFilter(void* unused)
 {
-	if (s_filter_list.curvalue == 0)
+	if (Q_stricmp(vid_renderer->string, "gl3") == 0 || Q_stricmp(vid_renderer->string, "gles3") == 0 ||
+		Q_stricmp(vid_renderer->string, "gl1") == 0)
 	{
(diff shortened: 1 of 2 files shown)


apartfromtime:
The software texture filtering appears unstable, I would recommend leaving that off. Vulkan users will have to set their options manually.

0lvin:
Software render does not support a texture filtering, code always down sample image to the original size of texture because code itself can't render walls and models with non default resolution. Maybe in future it could be fixed in yquake2remaster repository, it's not goal for now 

Images used for UI is downsampled to power of two size of original file resolution. Code has some bug with non 2x textures in UI, so sampled to 2x resolution only. 

If retexturing value is bigger than one, original texture could be upsampled 2x or 3x times that depends on screen resolution.

Software render has different approach of filtering, if image on the screen is not changed or position of player is unchanged e.g player is not running - wall texture could be rendered with [unreal engine hack](https://github.com/yquake2/yquake2/commit/184ff91453fea23b4b79168092f1fd3fe77d8576) or that more useful whole image made smooth if colors in palette are near. Real texture filtering is hugely limited as soft render supports only 256 color palette.

Soft render/Vulkan render could mimic little bit to GL renders, like convert setting from GL cvar to some similar settings specific to render.

0lvin (approved):

Yamagi:
Thank you :)

Issue #1093, opened 2024-03-23
Function args hide global variable `r_dlightframecount`


BraXi:
Some functions hide global variable `r_dlightframecount` by using argument with the same name

```C
void GL3_MarkSurfaceLights(dlight_t *light, int bit, mnode_t *node, int r_dlightframecount)
void R_MarkSurfaceLights(dlight_t *light, int bit, mnode_t *node, int r_dlightframecount)
void R_MarkLights (dlight_t *light, int bit, mnode_t *node, int r_dlightframecount, marksurfacelights_t mark_surface_lights);
```

0lvin:
Yes, it's bad code style. Shared code does not use global variables and receive all such values as parameter.Better to rename parameter name to something else, are you prefer to create pull request to yquake repository, or I will create pull request with fix and mark you as author of fix?

I will create fix for remaster in any case.

BraXi:
The change is trival and since you offer to PR yourself its okay

Yamagi:
Fixed by merging #1094.

PR #1094, opened 2024-03-23
Rename functions r_dlightframecount parameter to lightframecount

0lvin:
Renders has global variable with the such name that used as parameters for R_MarkLights shared function.

Found by @BraXi 

Fixes:
 * https://github.com/yquake2/yquake2remaster/issues/17
 * https://github.com/yquake2/yquake2/issues/1093


Yamagi:
Thank You :)

Issue #1098, opened 2024-03-30
[Suggestion] cl_laseralpha cvar


BjossiAlfreds:
I've always found it really hard to see red and blue lasers (I'm visually impaired), at least outside the software renderer. There could be other players with the same problem so I figured I'd suggest making the laser alpha value customizable, instead of hard-coding it to 0.3.

I can do a PR if you are interested.

Yamagi:
That sounds like a good idea. Go ahead :)

PR #1099, opened 2024-03-31
Added cvar cl_laseralpha and documentation in cvarlist

BjossiAlfreds:
Adds the cvar suggested in https://github.com/yquake2/yquake2/issues/1098.


Yamagi:
Thank you :)

PR #1100, opened 2024-04-01
Some improvements to the SW renderer

dkoreshkov:
1. Fixed some instances of bmodel clipping. E.g. the cracked explodable floor in base1 because btree side detection used dist > 0 instead of dist >= 0. The infamous fan in Borehole (drawn crudely by the way and having a concave polygon) now looks ok.
2. Cleanup of alias model to scanline conversion. Basically works the same except some left-clipped triangles now have correct texture positioning (the original, duplicated code had sfrac = tfrac = 0 in one place). I wonder why no one has fixed it in so many years of better-than-320x200 displays. Also added an integer floored division routine, which is faster on x86.

### DIFF
--- a/src/client/refresh/soft/sw_polyset.c
+++ b/src/client/refresh/soft/sw_polyset.c
@@ -291,13 +291,13 @@ quotient must fit in 32 bits.
 FIXME: GET RID OF THIS! (FloorDivMod)
 ====================
 */
-#if defined(__i386__) || defined(__amd64__)
-// just exclude ARM32 with FPU, e.g. Pi 1
 static void
 FloorDivMod(int numer, int denom, int *quo, int *rem)
 {
 	int q, r;
 
+// just exclude ARM32 with FPU, e.g. Pi 1
+#if defined(__i386__) || defined(__amd64__)
 	q = numer / denom;
 	r = numer - q * denom;
 	if (-1/2 || 1/-2 || -1/-2) {
@@ -313,7 +313,6 @@ FloorDivMod(int numer, int denom, int *quo, int *rem)
 			r += denom;
 		}
 	}
-	assert(r == numer - q * denom);
 	if ((numer < 0) ^ (denom < 0))
 		assert(q <= 0);
 	else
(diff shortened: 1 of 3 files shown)


src/client/refresh/soft/sw_polyset.c
```diff
@@ -299,6 +291,41 @@ quotient must fit in 32 bits.
 FIXME: GET RID OF THIS! (FloorDivMod)
 ====================
 */
+#if defined(__i386__) || defined(__amd64__)
+// just exclude ARM32 with FPU, e.g. Pi 1
+static void
+FloorDivMod(int numer, int denom, int *quo, int *rem)
```
0lvin, line 297: Could you please move `#ifdef` inside of function declaration? It should keep function definition in case if code will change in future and definition have changed only in one place by mistake.

src/client/refresh/soft/sw_polyset.c
```diff
@@ -299,6 +291,41 @@ quotient must fit in 32 bits.
...
+			r -= denom;
+		}
+	} else {
+		// C99 always truncates to 0
+		if ((numer ^ denom) < 0 && r != 0) {
+			q -= 1;
+			r += denom;
+		}
+	}
+	assert(r == numer - q * denom);
```
0lvin, line 316: Next code looks like just for testing code until code is developed. I suppose it could be removed if everything works as expected.

src/client/refresh/soft/sw_polyset.c
```diff
@@ -621,7 +645,7 @@ R_PolysetDrawSpans8_66(const entity_t *currententity, spanpackage_t *pspanpackag
 		}
 
 		pspanpackage++;
-	} while ((pspanpackage < triangles_max) && (pspanpackage->count != INT_MIN));
+	} while (pspanpackage->count != INT_MIN);
```
0lvin, line 648: I am expect that it will improve performance, but it looks as good as you have added additional `INT_MIN`, as for me it's good idea.

0lvin (commented):
It's really promising fix. Need to check whole game, I hope it will also fix glitches with ReRelease levels. Thank you!

src/client/refresh/soft/sw_polyset.c
dkoreshkov: I think the first part never evaluated to true. Anyway it was badly placed because we must check for overflow when adding spans, not when drawing them.

0lvin (approved):

Yamagi:
Thank you :)

0lvin:
> It's really promising fix. I need to check whole game and I hope it will also fix glitches with ReRelease levels. Thank you!

Thank you, I have merged code to remaster repository. Unfortunately, glitches with ReRelease maps are still exist, I suppose it's some different issue 😕

Issue #1097, opened 2024-03-27
Port to SDL3


Yamagi:
SDL 3 is the next major version of libsdl, the platform abstraction library used by Yamagi Quake II and many other games. It brings many improvements above SDL 2, but with an incompatible API. While YQ2 must support SDL 2 for the years to come, a SDL 3 would be nice.

I did a very proof of concept port within 90 minutes: https://github.com/yquake2/yquake2/tree/sdl3_poc A high quality port is much more work, though. The ongoing work can be found here: https://github.com/yquake2/yquake2/tree/sdl3

While the conceptional changes are smaller than during the transition from SDL 1.2 to SDL 3, the ABI churn is gigantic. Most function were renamed, their arguments reordered, return types changed, memory is now allocated by SDL, etc. This makes it near impossible to support SDL 2 and 3 side by side through #ifdef. Another challenge is, that SDL 2 are mutual exclusive. A substancial subset of YQ2 must be ported before the game can be build and tested with SDL 3.

This is a high level plan, aiming to archive a build and runnable version as soon as possible. Porting will be done on Linux, it will likely cover all other unixoid platforms. Windows will be done last.

To get a runnable version:

- [x] Makefile changes, support SDL 3 alongside SDL 2.
- [x] Port GL1.
- [x] Port GL3.
- [x] Fork *src/client/vid/glimp_vid.c* into a SDL 3 version and port it.
- [x] Fork *src/client/input/sdl.c* into a SDL 3 version and port the keyboard / mouse handling. Joystick / controller support, haptics, window events, etc. can be done later.
- [x] Fork *src/client/sound/sdl.c* into a SDL 3 version and comment enough code to get it to build. Sound handling is the only area were SDL 2 and 3 have great conceptional differences, porting the code may be rather difficult and can be done at a later time. OpenAL has been the default sound backend for >10 years anyways.

After that the rest:

- [x] Port the remaining parts of *src/client/input/sdl.c*.
- [x] Port *src/client/sound/sdl.c* (or may remove it?).
- [x] Port the software renderer.
- [x] Port external renderers.
- [x] CMakeFile changes.
- [x] Port the Windows platform backend to SDL 3.

SDL provides a semantic patch for Coccinelle to help with the port: https://github.com/libsdl-org/SDL/blob/main/build-scripts/SDL_migration.cocci

A general list of changes between SDL 2 and 3 can be found here: https://github.com/libsdl-org/SDL/blob/main/docs/README-migration.md

Open questions:

[...]

PR #1102, opened 2024-04-07
client: fix gles3 name

0lvin:
Fix regression introduced in e61d9444c

https://github.com/yquake2/yquake2/issues/1097


Yamagi:
Crap. :( I had tested it, but didn't realize the game directly jumped to gl1. Thank you for the quick fix!

Issue #147, opened 2016-08-02
Massive FPS drops during explosions with cl_lights 1 


iButane:
I get significant FPS drops during explosions with cl_lights 1. It's mostly noticeable during DM matches where you've got multiple rockets being fired etc. I know Yamagi is primarily a single player port but this issue is also in SP mode even though it's more of a concern in multiplayer. 
Also, there have been some user maps I've played which also have massive drops in FPS without any explosions.  I don't get these FPS drops in any other Q2 port I've used (q2pro and unofficial 3.24 patch for example).

Anyway, I'm currently running through the SP campaign and if I find a specific area where these FPS drops happen significantly without any action/explosions I'll post it here.  So far I can show you the FPS drops at the very beginning of the game, base1 explosions:

cl_lights 1: 
![quake2 2016-08-01 21-34-40-07](https://cloud.githubusercontent.com/assets/20784457/17314537/33728cdc-5833-11e6-8b05-c2099c15f7bf.jpg)
![quake2 2016-08-01 21-34-51-55](https://cloud.githubusercontent.com/assets/20784457/17314538/36207ade-5833-11e6-9496-51f565de5472.jpg)
![quake2 2016-08-01 21-35-48-56](https://cloud.githubusercontent.com/assets/20784457/17314539/3772d990-5833-11e6-9a2f-7e6997df24d7.jpg)

and cl_lights 0:

![quake2 2016-08-01 21-45-07-81](https://cloud.githubusercontent.com/assets/20784457/17314549/44020d34-5833-11e6-8833-3ae594d310ea.jpg)
![quake2 2016-08-01 21-45-08-51](https://cloud.githubusercontent.com/assets/20784457/17314552/4529d8d6-5833-11e6-98a1-5e70e7fa7b4e.jpg)
![quake2 2016-08-01 21-45-09-23](https://cloud.githubusercontent.com/assets/20784457/17314553/468a9440-5833-11e6-83e3-8da09cda1001.jpg)

My system specs are: Windows 7 64 bit, i5-2500 3.3ghz, 12GB RAM, AMD r9 270x

maraakate:
I've had this problem with integrated video cards.  Are you using an 
intel integrated card?  They seem to have massive slowdowns with dynamic 
lighting and multitexturing.


Frank

DanielGibson:
Holy shit, 54fps instead of 250 is a massive drop.

[...]

PR #1103, opened 2024-04-23
GL1 multitexture

protocultor:
While it is somewhat based in the one deleted almost 8 years ago by this project, this _mtex_ implementation is a very heavy revision on it, borrowing some properties from _Knightmare's KMQuake2_ source port, like keeping all lightmaps in memory for easier manipulation.

Features:

1. It is effectively apart from the standard rendering path. One of the reasons old multitexture was so slow is that it applied lightmaps again in a second pass (`R_BlendLightmaps()` function), which is redundant, lightmaps were already bound to the second TMU in the first (supposedly only) rendering pass. That was also one of the reasons [textures appeared so bright](https://github.com/yquake2/yquake2/issues/147), and rendering was so slow. This implementation is 100% visually consistent with the "no multitexture" rendering, doing both texture and lightmap in 1 pass.

2. Follows the standard texture chain, which avoids changing bound textures so much. "Old multitexture" rendered brushes without order, which bound to the same texture multiple times within a frame, again being redundant and reducing performance. The only heavy bind remaining here is the lightmap texture in the second TMU, which doesn't get re-bound that much compared to the color texture anyway.

3. In a step previous to rendering, dynamic lights are applied to the static lightmaps in memory, and then uploaded to the GPU, as a batch. Think `R_BlendLightmaps()` but in reverse. Avoids doing too many calls to `glTexSubImage2D()`, which are ridiculously slow.

4. Presents no lightmap generation errors, like the lava sectors in the `boss1` map with old mtex, which used to show fumbled textures.

5. `gl_showtris` works with this! It works even faster than classic rendering path.

There are a couple of issues:

1. Difference between max and min framerate is more pronounced, since the game goes almost 50% faster when nothing is happening, to just 5% faster when there's heavy lightmap generation. E.g. the exploding column in `base1` map, the very beginning of the game. Still, it must be mentioned that, unlike "old mtex", framerate is always above the standard "no mtex" rendering path, never below it.

2. `gl1_overbrightbits` works as intended for 1, 2 and 4, but jumping from 2 or 4 to 0 requires a `vid_restart`, or at least setting it to 1 first before going to 0.

I believe that, even with these issues, it's a huge help for people with old/small hardware.
Enabled by default, classic rendering path is nearly untouched if you want to go back.
I recommend testing this with the worst videocard/hardware you can find :)

Credits and thanks to _MH_ and _Knightmare_ for the dynamic lighting code (feature 3), which greatly helped the performance of this implementation.

Before finishing, I gotta mention a small but important fix included here: now the standard texture chain functions properly.

When adding a surface to the texture chain (in `R_RecursiveWorldNode()`), its current texture is identified by calling `R_TextureAnimation()`. Then, when browsing the chain in `R_DrawTextureChains()`, at the moment of drawing (`R_RenderBrushPoly()`), `R_TextureAnimation()` is called again to identify and bind the texture... which way too many times returns a different one, forcing GL1 to bind back and forth the same textures. You can apply this patch in the current master to see for yourself: [texture_chains_issue.patch](https://github.com/yquake2/yquake2/files/15080028/texture_chains_issue.patch)

Now the chain is respected, which helps rendering whether you are using multitexture or not.

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -464,6 +464,12 @@ Set `0` by default.
   value, at least `1.0` - default is `2.0`. Applied when textures are
   loaded, so it needs a `vid_restart`.
 
+* **gl1_multitexture**: Enables (`1`) the blending of color and light
+  textures on a single drawing pass; disabling this (`0`) does one pass
+  for color and another for light. Default is `2`, which also enables
+  texture combine mode (`GL_ARB_texture_env_combine`) when supported.
+  Requires a `vid_restart` when changed.
+
 * **gl1_overbrightbits**: Enables overbright bits, brightness scaling of
   lightmaps and models. Higher values make shadows less dark. Possible
   values are `0` (no overbright bits), `1` (more correct lighting for
--- a/src/client/refresh/gl1/gl1_image.c
+++ b/src/client/refresh/gl1/gl1_image.c
@@ -145,13 +145,37 @@ R_SetTexturePalette(unsigned palette[256])
 }
 
 void
+R_SelectTexture(GLenum texture)
+{
+	int tmu;
+
+	if (!gl_config.multitexture)
+	{
+		return;
+	}
+
+	tmu = texture - GL_TEXTURE0;
+
+	if (tmu == gl_state.currenttmu)
+	{
+		return;
+	}
+
+	gl_state.currenttmu = tmu;
+	gl_state.currenttarget = texture;
+
+	qglActiveTexture(texture);
+	qglClientActiveTexture(texture);
+}
+
+void
 R_TexEnv(GLenum mode)
 {
 	static int lastmodes[2] = {-1, -1};
 
 	if (mode != lastmodes[gl_state.currenttmu])
 	{
-		glTexEnvf(GL_TEXTURE_ENV, GL_TEXTURE_ENV_MODE, mode); // FIXME: shouldn't this be glTexEnvi() ?
+		glTexEnvi(GL_TEXTURE_ENV, GL_TEXTURE_ENV_MODE, mode);
 		lastmodes[gl_state.currenttmu] = mode;
 	}
 }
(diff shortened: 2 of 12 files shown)


Yamagi:
Wow, that came unexpected :) I'm just back from vacation and will have a look in the next few days.

Yamagi:
And now "the next few days" are over :) I looked over the code and it looks good. My OpenGL skills are limited, though. And I checked some of the places were the old multitexturing code failed [1], they are rendering fine now. On my laptop with an Core i7-8550U IGP (the slowest GPU I have without tinkering) FPS increase from 195 with `gl1_multitexture 0` to 265 with `gl_multitexture 2`. Good work!

Yamagi:
1: A good place to check is the entity in city3 (a func_elevator?) where the security CD is placed upon. With the old multitexturing code it rendered only the lightmap and not the actual texture.

protocultor:
Haha, I saw some errors with "old multitexture", but not that one.
In the code for "old mtex", if you open `gl1_image.c`, search for the `R_EnableMultitexture()` function definition, and look for this section:
```
	if (enable)
        {
                R_SelectTexture(GL_TEXTURE1);
                glEnable(GL_TEXTURE_2D);
                R_TexEnv(GL_REPLACE);
        }
        else
	...
```
Just change the `R_TexEnv(GL_REPLACE)` to `R_TexEnv(GL_MODULATE)` and the column in _city3_ will appear correctly.

This should always have been like this, after all the TMU 1 is always used for lightmaps, and lightmaps always "modulate" the color texture in TMU 0. But for some reason, the selection between using normal _GL_MODULATE_ (today's `gl1_multitexture 1`), or the complex combination for the _GL_ARB_texture_env_combine_ extension (now `gl1_multitexture 2`), was in `R_DrawBrushModel()` instead of `R_EnableMultitexture()`, which never made sense.

The order in how things were done in old multitexture made these kind of errors possible, something I wanted to correct in this PR. Many thanks for merging it so quickly (I thought you would take more time with it).

PR #1106, opened 2024-05-10
update of tag allocator.

devnexen:
without changing the size of the zone head type.
introducing YA2_UNLIKELY to optimise rarely called edge cases.

### DIFF
--- a/src/common/header/common.h
+++ b/src/common/header/common.h
@@ -803,8 +803,8 @@ extern int time_before_ref;
 extern int time_after_ref;
 
 void Z_Free(void *ptr);
-void *Z_Malloc(int size);           /* returns 0 filled memory */
-void *Z_TagMalloc(int size, int tag);
+void *Z_Malloc(unsigned int size);           /* returns 0 filled memory */
+void *Z_TagMalloc(unsigned int size, int tag);
 void Z_FreeTags(int tag);
 
 void Qcommon_Init(int argc, char **argv);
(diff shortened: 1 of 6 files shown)


Yamagi:
Does this fix a bug or yield a measurable performance improvement? Using `size_t` and unsigned integers is formally more correct, nevertheless I'm hesitant to merged changes to the allocator. The code is fragile and lives from assumptions, we had bugs from well-mend changes in the past, etc.

devnexen:
It does not fix any bug I know, it s about correctness, the yq2_unlikely bit however brings micro optimisations for careful chosen cases.

BjossiAlfreds:
This changes the import API. Even if you update the API for the expansions as well, mods with custom gamex86 will continue to set up the function calls as if the 1st argument is a regular int. I guess this should be safe for compatibility as signed and unsigned are the same value as long as the sign bit is 0 (assuming two's complement) but the question is if that is standard across all platforms.

protocultor:
> but the question is if that is standard across all platforms, is it guaranteed that `sizeof(size_t) == sizeof(int)`?

Yeah, that's not gonna happen.
Since @turol cannot post here for whatever reason, I'm going to parrot his comments just to make them visible:

Changing `int` parameter to `size_t` is not safe on 64-bit windows:
https://godbolt.org/z/eeadY1qGE
https://godbolt.org/z/WPPeGTexb

Also the code as posted does not work correctly on 32-bit. In the overflow check in Z_TagMalloc the addition will overflow and the comparison with `UINT_MAX` will always be false.

DanielGibson:
> is it guaranteed that sizeof(size_t) == sizeof(int)

not at all. on all releveant 64bit platforms I know of `sizeof(int) == 4` and `sizeof(size_t) == 8`.

Issue #1095, opened 2024-03-23
[idea] pack 4 surface lightmaps into one?


BraXi:
Digging through the lighting code I've noticed that there are 4 lightmap textures being created for each lit surface, which is understandable given thats the maximum count of lightmaps per surface, but It could be improved to build a single texture containing 4 lightmaps which in return would result in less texture binds when drawing lightmapped world geometry and likely improve fps a little. I've only checked GL3 renderpath and I assume GL4 does the same.

I think the benefits are pretty noticeable
- one instead of four lightmap texture binds per lightmapped surface
- four times less lightmap textures (MAX_LIGHTMAPS vs MAX_LIGHTMAPS*4)
- no need to switch texture units (currently tmu1..4), lighting could be done on TMU 1 all the time

Yamagi:
```C
void
LM_UploadBlock(void)
{
	int map;

	// NOTE: we don't use the dynamic lightmap anymore - all lightmaps are loaded at level load
	//       and not changed after that. they're blended dynamically depending on light styles
	//       though, and dynamic lights are (will be) applied in shader, hopefully per fragment.

	GL3_BindLightmap(gl3_lms.current_lightmap_texture);

	// upload all 4 lightmaps
	for(map=0; map < MAX_LIGHTMAPS_PER_SURFACE; ++map)
	{
		GL3_SelectTMU(GL_TEXTURE1+map); // this relies on GL_TEXTURE2 being GL_TEXTURE1+1 etc
		glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
		glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);

		glTexImage2D(GL_TEXTURE_2D, 0, GL_LIGHTMAP_FORMAT,
		             BLOCK_WIDTH, BLOCK_HEIGHT, 0, GL_LIGHTMAP_FORMAT,
		             GL_UNSIGNED_BYTE, gl3_lms.lightmap_buffers[map]);
	}

	if (++gl3_lms.current_lightmap_texture == MAX_LIGHTMAPS)
	{
		Com_Error(ERR_DROP, "%s: MAX_LIGHTMAPS exceeded\n", __func__);
	}
}
```

For reference, in my own project I upload all 4 surface lightmaps into a single texture:
```C
static void R_LightMap_UploadBlock()
{
[...]

PR #1108, opened 2024-05-11
GL1 big lightmaps and scrap

protocultor:
Inspired by @BraXi's [idea](https://github.com/yquake2/yquake2/issues/1095), this PR allows the user to increase the lightmaps' size, having a more optimal distribution in GPU memory:
- Fewer lightmaps needed to cover the full map, which reduces the maximum number of lightmaps needed
- Less context switch change: a larger lightmap covers more surfaces, avoiding changing to a specific lightmap to start rendering.
- Fewer lightmaps mean fewer calls to `glTexSubImage2D()` when handling dynamic lights; this might be the most expensive OpenGL call the game has.

A new lightmap can reach a maximum size of 512x512, being equivalent to 16 lightmaps in the "classic" format (128x128). This, in turn, reduces the _MAX_LIGHTMAPS_ count to a minimum of 8, reducing the number of loops with some "for" statements.

This PR also increases the size of the "**scrap**", an atlas for 2D textures destined to HUD elements. Like lightmaps, originally has a size of 128x128 ([256x256 in Q2 3.21](https://github.com/id-Software/Quake-2/blob/master/ref_gl/gl_image.c#L348)), but now this also can increase to 384x384, which is almost double of what's needed to cover all HUD pictures in a single texture. This should benefit many low-end machines.
Note that scrap code has changed to add a transparent border of 1 pixel on all 4 directions, to avoid color bleeding between images. This bleeding is visible very rarely today, with the "128x128 scrap", in the Q2 "cursor logo" of the main menu, where in some of the frames of the spinning animation, white pixels could appear at the top or bottom of the logo.

Logically, a check is done at renderer init to verify hardware support for larger textures. This is optional and controlled by a new cvar, _gl1_biglightmaps_ (suggestions of a better name are welcome :) ).
One last, unrelated thing: fixed a cvar query after all cvars were freed in Windows, evaluating always to false and even potentially causing crashes at game exit.

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -460,7 +460,7 @@ Set `0` by default.
 
 ## Graphics (OpenGL 1.4 only)
 
-* **gl1_biglightmaps**: Enables lightmaps to use a bigger
+* **gl1_biglightmaps**: Enables lightmaps and scrap to use a bigger
   texture size, which means fewer texture switches, improving
   performance. Default is `1` (enabled). Requires a `vid_restart`.
 
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -460,6 +460,10 @@ Set `0` by default.
 
 ## Graphics (OpenGL 1.4 only)
 
+* **gl1_biglightmaps**: Enables lightmaps to use a bigger
+  texture size, which means fewer texture switches, improving
+  performance. Default is `1` (enabled). Requires a `vid_restart`.
+
 * **gl1_intensity**: Sets the color intensity. Must be a floating point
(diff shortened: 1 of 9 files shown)


Yamagi:
This is nearly the same approach @DanielGibson took in GL3 with 1024x512 lightmaps. Good to see something like that implemented in GL1. I gave it a try and played through the first unit, no visible rendering problems. In my Intel IGP it makes no speed difference if `gl1_biglightmaps` is `0` or `1`, but that GPU is already very fast.

> One last, unrelated thing: fixed a cvar query after all cvars were freed in Windows, evaluating always to false and even potentially causing crashes at game exit.

Maybe I need to test the game also under Windows with ASAN and UBSAN.

PR #1109, opened 2024-05-29
Fix potential memory corruption issue where cl_numparticles is 0 and CL_ClearParticles is called

starfrost013:
If CL_ClearParticles is called with zero particles, cl_particles[0-1] = -1 will have its "next" field set to NULL, potentially corrupting 4 bytes of memory at (cl_particles)-[80..76]. This may be user-manipulable but I'm not sure how bad this is in practice.

### DIFF
--- a/src/client/cl_particles.c
+++ b/src/client/cl_particles.c
@@ -33,6 +33,9 @@ int cl_numparticles = MAX_PARTICLES;
 void
 CL_ClearParticles(void)
 {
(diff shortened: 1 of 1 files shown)


Yamagi:
It doesn't matter, because `cl_numparticles` ist always set to `MAX_PARTICLES`, altering that define will likely break things. Nevertheless this is nice find and makes the code somewhat more robust in corner cases.

PR #1110, opened 2024-06-04
Gyro tightening

protocultor:
This is a way to counter involuntary hand movements, or just a noisy signal, when using gyro aiming. Consists in a _deceleration_ applied only under a small threshold, so small rotations are dampened, being "squeezed to zero", but not completely ignored. This provides a much better feeling than a simple hard deadzone, which cuts all player rotations... including voluntary ones, losing the point of using gyro as a mouse (c'mon _Nightdive Studios_, you can do better!). Credits again to @JibbSmart for creating this ([source](http://gyrowiki.jibbsmart.com/blog:good-gyro-controls-part-1:the-gyro-is-a-mouse#toc9)).

Again, extras are included, and there are a lot for the OpenGL 1 renderer:

- _Multitexture_ bugfix: when using `hand 2` (no player weapon visible), and there were no "_alias_" models in the player's sight, only the lightmap was visible on the map. Multitexture is now forced to be disabled when needed.
- `gl_showbbox 1` was producing OpenGL errors, visible only when compiling with `DEBUG=1`.
- Deleted `multitexture combine` mode. This was made in a time before the _programmable pipeline_, to create effects that could not be done in old style OpenGL. But in _Quake 2_'s case, it was being used for an effect that can be replicated, in a simpler and quicker way, with just a `GL_MODULATE`, so the amount of OpenGL calls to do the same were not justified. Explains its complete absence from _Quake 3 Arena_ code.
- OpenGL boolean flags are in the proper format.

Finally, and getting out of GL1: code refactor for menu sliders and gamepad buttons; just removing redundancy and improving clarity.

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -598,6 +598,10 @@ Set `0` by default.
   for people who hold the controller upright, or use a device with the
   controller attached to the screen, e.g. Steam Deck.
 
+* **gyro_tightening**: Threshold of rotation in degrees per second,
+  where gyro inputs below it will be dampened. Meant to counter a
+  noisy gyro and involuntary hand movements. Default `3.5`.
+
 * **gyro_calibration_(x/y/z)**: Offset values on each axis of the gyro
   which helps it reach true "zero movement", complete stillness. These
   values are wrong if you see your in-game view "drift" when leaving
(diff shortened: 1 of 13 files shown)


Yamagi:
As always: Thank you :) The code looks good and I gave it a short try without any problems. Of course there still might be rendering bugs. Let's see if you or someone else notices something.

protocultor:
I'm always available to fix anything I might have introduced to the GL1 renderer. Thank you.

PR #1111, opened 2024-06-08
Show weapon in player model preview

0lvin:
Additionally add long (QPATH) skin preview.

If some of multiplayer map is already loaded, weapon is changed each 9 seconds. In other case weapon.md2 is showed.


apartfromtime:
You could replace `0` with `NULL` in the other functions associated with the menu player model, I tend to use `intptr_t` for pointers and handles in my own code.

Yamagi:
Very nice idea. :) Thank you!

Issue #867, opened 2022-08-09
Recent Homebrew messes up compiling on M1 Mac


emarkham:
Seems that somewhat recently, Homebrew installs to `/opt/homebrew` instead of `/usr/local/`

This screws up openal-soft and, in a roundabout way, SDL2

for example, openal-soft includes were at:
`/usr/local/opt/openal-soft/include`

but now they're at:
`/opt/homebrew/Cellar/openal-soft/1.22.2/include`

also, SDL2 includes are living at:
`/opt/homebrew/include/SDL2`

but what makes the compiler happy is:
`-I/opt/homebrew/include/`

because most everything has` #include <SDL2/SDL.h>` 

I've hastily patched the Makefile to allow compilation on my machine and it's working but, um, I'm not much of a programmer and I've never contributed a patch. And I can't seem to create a branch. And I don't really know what I'd contribute other than just this thing

```
diff --git a/Makefile b/Makefile
index f381d56c..0a58fb57 100644
--- a/Makefile
+++ b/Makefile
@@ -281,11 +281,7 @@ endif
 # ----------
 
 # Extra CFLAGS for SDL.
-ifeq ($(YQ2_OSTYPE),Darwin)
-SDLCFLAGS := $(shell sdl2-config --cflags 2>&1 | sed 's%/SDL2%%')
-else
 SDLCFLAGS := $(shell sdl2-config --cflags)
-endif
 
 # ----------
 
@@ -362,8 +358,11 @@ endif
 # ----------
 
 # Extra LDFLAGS for SDL
-
+ifeq ($(YQ2_OSTYPE), Darwin)
+SDLLDFLAGS := -lSDL2
+else # not Darwin
 SDLLDFLAGS := $(shell sdl2-config --libs)
+endif # Darwin
 
 # The renderer libs don't need libSDL2main, libmingw32 or -mwindows.
 ifeq ($(YQ2_OSTYPE), Windows)
@@ -481,10 +480,8 @@ ifeq ($(WITH_OPENAL),yes)
 ifeq ($(YQ2_OSTYPE), OpenBSD)
 release/quake2 : CFLAGS += -DUSE_OPENAL -DDEFAULT_OPENAL_DRIVER='"libopenal.so"'
 else ifeq ($(YQ2_OSTYPE), Darwin)
[...]

PR #1115, opened 2024-06-14
Update Makefile to build on Apple Silicon

hbokh:
This will fix issue #867.

Should work on both macOS / Darwin with Intel CPUs and on Apple Silicon (tested).


Yamagi:
Do we need similar changes for the xatrix and rogue addons?

hbokh:
> Do we need similar changes for the xatrix and rogue addons?

Looking at the source code: yes.
But loading either of them as mods within the built `arm64` binary ("Apple Silicon") also makes them playable.

Yamagi:
Thank you :)

PR #1116, opened 2024-06-29
micro optimisations for previous security mitigations.

devnexen:
also strtok_r uses its own provided buffer instead of the static one even in a somewhat monothread context, it s still better.


Yamagi:
Can you rebase it onto the current master? I pushed some changes to the download filter and now this branch doesn't apply anymore.

Yamagi:
Thank you :)

PR #1118, opened 2024-06-30
fix few warning builds.

devnexen:


src/backends/unix/system.c
```diff
@@ -402,7 +402,8 @@ Sys_UnloadGame(void)
 void *
 Sys_GetGameAPI(void *parms)
 {
-	void *(*GetGameAPI)(void *);
+	typedef void *(*fnAPI)(void *);
+	fnAPI GetGameAPI;
```
Yamagi, line 406: What warning does this fix?

devnexen: This particular line does not however it s to fix the dlsym part below

```
warning: ISO C forbids assignment between function pointer and ‘void *’
482 |         GetGameAPI = (void *)dlsym(game_library, "GetGameAPI");

```

Yamagi: Ah, I see.

Yamagi:
Thank you :)

Issue #1117, opened 2024-06-30
Juggernaut The Civic Center map has no keys


igorpupkinable:
This may not be the right place to post Juggernaut issue, but I would appreciate if you could advice on how to fix the following.

Juggernaut **The Civic Center** map (jug13) is missing red and blue keys as well as security pass.
Keys and pass placeholders are empty from the start, so no chance to open doors and progress to the next stage.

Is it possible to fix this with _ent_ files or other tweaks without source code?

Steps to reproduce:
1. Either progress through the mission pack or just `map jug13` in the console.
2. `noclip` in the console after map loading and fly to the top center of the map where one of the keys supposed to be placed.
3. **Expected:** both key placeholders/podiums have keys. **Actual:** none of the podiums have keys and security pass.

Thanks.

Yamagi:
@BjossiAlfreds, can you have a look?

BjossiAlfreds:
The key explodes, getting crushed by the moving brush I guess. I tested in the original engine and the key is where it should be. So something in YQ2 is different that is making the collision detection a bit more sensitive. I have no idea where though, this could be in the trace code or the physics code.

Yamagi:
Maybe another floating point problem. YQ2 is compiled with 64 bit floats, the original quake2.exe used 80 bit floats. This has bitten us before, no matter which way we go - 64 bit or 80 bit - some things break. We have choosen 64 bit because 80 bit precision is bound to the x87 fpu only available on x86 CPUs.

BjossiAlfreds:
I played around with the triggering sequence of the keys a bit more. There is this func_door with targetname "overhead" that is supposed to open first, and then the brush the key is resting on lowers through the opening.

If I trigger the "overhead" event first, and then "g_key" to lower the key, it seems to work. So this is looking more like a timing issue. The level designers might have tried values until the key stopped exploding to get the timing right, but when that sensitive balance changes it can break it.

I'll see if I can make the timings more forgiving to avoid the collision.

BjossiAlfreds:
I've made an ent file that hopefully fixes this. See if it works.

The red key trigger lacked a delay of 1 second that the other key triggers had. I don't know why this setup works in the original, probably some sensitive balance that let this map bug slip by. I also saw another map bug where some targets were incorrectly naked r_cl instead of r_key_cl, so fixed that too.

[jug13ent.zip](https://github.com/user-attachments/files/16046300/jug13ent.zip)

igorpupkinable:
[...]

PR #1120, opened 2024-07-05
Mapfix for map jug13 (The Civic Center) in Juggernaut

BjossiAlfreds:
This PR addresses a few map issues found as a result of bug report https://github.com/yquake2/yquake2/issues/1117

* Fixed shootable button that reveals the red key being re-shootable. Should have had a wait time of -1.
* Removed unused 4096 spawnflag from a bunch of entities.
* Fixed a broken trigger chain that intended to pull the security pass holder back into the ceiling when the key gets used to unlock the door. The trigger_key had a target of r_cl while the targetname of the trigger_relays that pull up the key holder was r_key_cl which was never targeted. The targeted name is now consistent.

### DIFF
--- a/stuff/mapfixes/juggernaut/jug13.ent
+++ b/stuff/mapfixes/juggernaut/jug13.ent
@@ -1,0 +1,2759 @@
+// FIXED ENTITY STRING (by BjossiAlfreds)
+//
+// 1. Removed unused entities (b#1)
+//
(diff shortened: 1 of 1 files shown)


igorpupkinable:
I have tested these fixes by going through the map few times and everything seems to work fine.

Yamagi:
Thank you :)

Issue #1119, opened 2024-07-05
PCX Loader bug


protocultor:
Delivering this on behalf of certain people from Map-Center Discord channel.

The attached zip contains a couple of PCX files, part of a menu mod that is only a recolored dog tag, but Yamagi is not able to show properly:
![Yamagi bug 2](https://github.com/yquake2/yquake2/assets/32081604/52a92493-575b-45fd-886f-810e57573eb2)

This seems to be a legacy bug from R1Q2. Note that this does appear correctly in both Q2Pro and the game re-release.
[pics.zip](https://github.com/user-attachments/files/16109899/pics.zip)

0lvin:
Interesting, I will look. it could be just unexpected image format. Have you tried with disabled disabled re_texturing? 

I suppose it could be some regression as in some of commits, I have forced to convert all images with palette to use embedded to file palette, instead colormap.pcx file one.

Paril:
FWIW it's probably a bug in the RLE decoder that was brought over from vanilla; a bunch of older engines seem to have it, but Q2PRO and re-release don't; we used our own code in re-release, and Q2PRO has a rewritten loader.

0lvin:
@Paril thank you, I have compared code in q2pro, currently code has ignored bytes per line field. I will check little bit more and create pull request.

Yamagi:
Should be fix by merging #1121.

0lvin:
I have implemented support for all kinds of [pcx  version five files](https://github.com/yquake2/yquake2remaster/blob/4e61cbccc2b97817830c2a09d7f0aa4457b1772f/src/client/vid/image.c#L190), code is based on ffmpeg code and checked with samples from ffmpeg. Code has showed all images same as gimp showed it as I see.

PR #1121, opened 2024-07-06
Improve pcx support and cleanup soft render

0lvin:
Add use byte per line in pcx loader #1119
Fix apply skin with incorrect ratio to model in soft skin


Yamagi:
It's always interesting that these somewhat obvious bugs can go unnoticed for nearly 27 years... Thank you very much for fixing this :)

PR #1123, opened 2024-07-09
Workarounds for naggy help icons

BjossiAlfreds:
This PR is a general, code-based workaround version of PR https://github.com/yquake2/rogue/pull/115.

This PR has the following improvements:
* target_help that are activated during the level pre-load period (first 2 game frames) delay the effect until the 3rd frame. Savegame loads also run these pre-load frames so if target_help activates, its effect will leak, causing unwanted changes to help messages. This happens for example in rhangar2 in GroundZero and a number of Juggernaut maps.

* If a target_help has identical text as what is currently in the help computer, its activation is ignored. An example of this is the start of Guard House (jail5). It has a target_help triggered by a trigger_multiple, causing repeated activation of the same help message. This PR prevents spamming the help icon in such cases.

* Updated jug13 and jug19 mapfixes to add a bit of delay to the trigger_always that trigger target_help.

I have tested this and everything seems to work as expected but I welcome others to test and review the changes in case I overlooked something.

### DIFF
--- a/src/game/g_target.c
+++ b/src/game/g_target.c
@@ -26,6 +26,9 @@
 
 #include "header/local.h"
 
+#define TARGET_HELP_PRIMARY 1
+#define TARGET_HELP_THINK_DELAY 0.3f
+
 /*
  * QUAKED target_temp_entity (1 0 0) (-8 -8 -8) (8 8 8)
  * Fire an origin based temp entity event to the clients.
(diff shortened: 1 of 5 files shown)


Yamagi:
Thank you :)

> I have tested this and everything seems to work as expected but I welcome others to test and review the changes in case I overlooked something.

I'll merge it. Let's see if something comes up.

Yamagi:
A new Windows testbuild with this and all other recent changes can be found here: https://deponie.yamagi.org/quake2/windows/testbuilds/quake2-20240713-fd2ebb8.zip

Issue #949, opened 2022-12-20
Stuttering for no discernable reason


Technoturnovers:
Maybe this is just user error and I'm an idiot, but I just cannot figure out what the issue here is.

I've been trying to play at 4k resolution, but whenever I do the game will start lagging and stuttering despite my Nvidia FPS overlay consistently reporting 70+ FPS. This usually happens when there are enemies on screen. I would just put this up to my hardware not being up to the task, but I've played newer games at the same resolution with no issue, and the fact that this performance issue is seemingly untied to FPS suggests to me that something way weirder is going on. I've already tried manually setting my refresh rate to 60 as per #277 in case that was related to no effect, and the issue shows up with both GL1 and GL3 as the renderer.

For comparison, here's 2 clips, one showing me playing Quake II with yamagi's GL3 renderer, the latter showing me playing Return to Castle Wolfenstein via iortcw, both at 4k, just to prove that this isn't my computer being complete trash. Both of these clips were recorded via nvidia shadowplay, which has pretty minimal performance impact in my experience with the exception of starting and ending recording.
https://cdn.discordapp.com/attachments/524684262621052938/1054635511697920040/Yquake2.Exe_2022.12.20_-_00.40.54.02.webm
https://cdn.discordapp.com/attachments/524684262621052938/1054637581582729226/Iowolfsp.X64.Exe_2022.12.20_-_00.48.14.03.webm

If there's any more information that would help, I'll gladly provide it.

---

OS: Windows 11 Enterprise 21H2 - 22000.1335 (Insider build)
GPU: NVIDIA GeForce RTX 2070 with Max-Q Design (Driver version 516.40)
CPU: Intel Core i7-10750H @ 2.60GHz
RAM: 16GB DDR4 @ 2933MHz
Yamagi Quake II Version: 8.20

maraakate:
Are you certain no power saving modes are enabled in your video driver?  Try renaming quake2.exe to yq2.exe and see if that fools any app compatibility "helpers".

Yamagi:
If the config is clean (no cvars messing with the timings altered) this is properly something external messing with the game. Like @maraakate says, maybe a well meaning graphics driver or - often the case when playing on laptop - power management clocking the CPU or GPU too low.

protocultor:
Hello @Technoturnovers, I know a year and a half have passed, but I implemented a few optimizations in the GL1 renderer for Yamagi Quake II v8.40, and I'm interested in knowing if they have been of any help for your case. Regards.

PR #1124, opened 2024-07-15
Grouped draw calls for improved performance with GL1

protocultor:
_Quake III Arena_ was a revolution because of its use of "_shaders_", which doesn't refer about the "_programmable pipeline_" (the modern sense), but to the scripted features of a surface, which in backend meant the replacing of many small `glDrawArrays()` calls with a single big `glDrawElements()` call, which many OpenGL documents will agree is the preferred way to pursue an increase of performance.

The short version of the thinking behind this is: if there are a lot of `glDrawArrays()` with GL_TRIANGLE_FANs (or any other primitive) with the same texture, instead create indices that regenerate these FANs (or primitives) manually, and at the end do a `glDrawElements()` call, which use these indices to draw instead of the original vertices. May sound more expensive, but indices + vertices don't use much more space than the vertices alone, and as OpenGL uses a "client-server" architecture internally, 1 communication established between CPU and GPU is a lot less than the 6, or more, it uses to draw the same primitives in a normal scene.

Some Q2 ports like _Knightmare_ and [Thenesis Quake 2](https://github.com/thenesis-org/lp-public) use a similar technique for drawing, and here I'm presenting my version, which IMHO is more respectful of the original Q2 functions' organization, while being more readable, or at least simpler to follow in code. All this work reduces the OpenGL calls to around ~1500 per frame which, while still over the ~1000 Quake 3 uses, is better than the current ~8000 Yamagi / vanilla Q2 has. These numbers were obtained thanks to [apitrace](https://github.com/apitrace/apitrace).

By disabling vsync, and using `timedemo 1` with different demos (especially Xatrix's `xdemo1.dm2`), up to a ~45% speed improvement can be observed against the current master. As usual, the worse the hardware, the better the test :)

One major difference with other implementations is the reduction of calls to `glTexSubImage2D()`, which in this PR gets reduced to one call per lightmap, per frame. This is thanks to the pre-calculation of dynamic lights for `mod_brush` models before they are drawn, in the `R_GetBrushesLighting()` function.

Note: calls to `glTexSubImage2D()` are normally sluggish. But for big sized textures, those calls are even slower, and not many times dynamic lights cover more than two of the classic 128x128 area, so the advantage of grouping many surfaces under one lightmap was lost. Having big lightmaps used to be fast, but after this batching of draw calls, it became a nuisance, so I partially removed it and restored the classic 128x128 area. Original `gl1_biglightmaps` logic remains underneath, but it's preferable not to make it visible to the user. Let's see if there's a need to restore it in the future.

While I don't have the hardware to test it, I believe this PR can help a lot with issue #949, since this work's approach to rendering gets it a little bit closer to Q3A/RTCW.

@DanielGibson: if you still have it, can you test this PR with your Raspberry Pi 4? I believe the work done here, applied to the GLES3 renderer, can help a lot with its performance. Just a thought.

Credits and thanks to @Paril for his explanation of Q2's dynamic lighting in Discord.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -590,6 +590,7 @@ set(GL1-Source
 	${REF_SRC_DIR}/gl1/gl1_surf.c
 	${REF_SRC_DIR}/gl1/gl1_warp.c
 	${REF_SRC_DIR}/gl1/gl1_sdl.c
+	${REF_SRC_DIR}/gl1/gl1_buffer.c
 	${REF_SRC_DIR}/files/models.c
 	${REF_SRC_DIR}/files/pcx.c
 	${REF_SRC_DIR}/files/stb.c
--- a/Makefile
+++ b/Makefile
@@ -959,6 +959,7 @@ REFGL1_OBJS_ := \
 	src/client/refresh/gl1/gl1_surf.o \
 	src/client/refresh/gl1/gl1_warp.o \
 	src/client/refresh/gl1/gl1_sdl.o \
+	src/client/refresh/gl1/gl1_buffer.o \
 	src/client/refresh/files/surf.o \
 	src/client/refresh/files/models.o \
 	src/client/refresh/files/pcx.o \
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -471,10 +471,6 @@ it's `+set busywait 0` (setting the `busywait` cvar) and `-portable`
 
 ## Graphics (OpenGL 1.4 only)
 
-* **gl1_biglightmaps**: Enables lightmaps and scrap to use a bigger
-  texture size, which means fewer texture switches, improving
-  performance. Default is `1` (enabled). Requires a `vid_restart`.
-
 * **gl1_intensity**: Sets the color intensity. Must be a floating point
   value, at least `1.0` - default is `2.0`. Applied when textures are
   loaded, so it needs a `vid_restart`.
--- a/src/client/refresh/gl1/gl1_buffer.c
+++ b/src/client/refresh/gl1/gl1_buffer.c
@@ -1,0 +1,171 @@
+/*
+ * Copyright (C) 1997-2001 Id Software, Inc.
+ * Copyright (C) 2024      Jaime Moreira
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or (at
+ * your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
+ * See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA
+ * 02111-1307, USA.
+ *
+ * =======================================================================
+ *
+ * Drawing buffer: sort of a "Q3A shader" handler, allows to join multiple
+ * draw calls into one, by grouping those which share the same
+ * characteristics (mostly the same texture).
+ *
+ * =======================================================================
+ */
+
+#include "header/local.h"
+
+#define MAX_VERTICES	16384
+#define MAX_INDICES 	(MAX_VERTICES * 4)
+
+typedef struct	//	832k aprox.
+{
+	buffered_draw_t	type;
+
+	GLfloat
+		vtx[MAX_VERTICES * 3],	// vertexes
+		tex[MAX_TEXTURE_UNITS][MAX_VERTICES * 2],	// texture coords
+		clr[MAX_VERTICES * 4];	// color components
+
+	GLushort
+		idx[MAX_INDICES],	// indices
+		vtx_ptr, idx_ptr;	// pointers for array positions
+
+	int	texture[MAX_TEXTURE_UNITS];
+	int	flags;	// entity flags
+	float	alpha;
+} glbuffer_t;
+
+glbuffer_t gl_buf;
+
+GLuint vt, tx, cl;	// indices for arrays in gl_buf
+
+extern void R_MYgluPerspective(GLdouble fovy, GLdouble aspect, GLdouble zNear, GLdouble zFar);
+
+void
+R_ApplyGLBuffer(void)
+{
+	// Properties of batched draws here
+	GLint vtx_size;
+	qboolean texture;
+
+	if (gl_buf.vtx_ptr == 0 || gl_buf.idx_ptr == 0)
+	{
+		return;
+	}
+
+	// defaults for drawing
+	vtx_size = 3;
+	texture = true;
+
+	// choosing features by type
+	switch (gl_buf.type)
+	{
+		case buf_2d:
+			vtx_size = 2;
+			break;
+		default:
+			break;
+	}
+
+	glEnableClientState( GL_VERTEX_ARRAY );
+	glVertexPointer (vtx_size, GL_FLOAT, 0, gl_buf.vtx);
+
+	if (texture)
+	{
+		R_Bind(gl_buf.texture[0]);
+
+		glEnableClientState(GL_TEXTURE_COORD_ARRAY);
+		glTexCoordPointer(2, GL_FLOAT, 0, gl_buf.tex[0]);
+	}
+
+	// All set, we can finally draw
+	glDrawElements(GL_TRIANGLES, gl_buf.idx_ptr, GL_UNSIGNED_SHORT, gl_buf.idx);
+	// ... and now, turn back everything as it was
+
+	if (texture)
+	{
+		glDisableClientState( GL_TEXTURE_COORD_ARRAY );
+	}
+
+	glDisableClientState( GL_VERTEX_ARRAY );
+
+	gl_buf.vtx_ptr = gl_buf.idx_ptr = 0;
+}
+
+void
+R_UpdateGLBuffer(buffered_draw_t type, int colortex, int lighttex, int flags, float alpha)
+{
+	if ( gl_buf.type != type || gl_buf.texture[0] != colortex )
+	{
+		R_ApplyGLBuffer();
+
+		gl_buf.type = type;
+		gl_buf.texture[0] = colortex;
+		gl_buf.texture[1] = lighttex;
+		gl_buf.flags = flags;
+		gl_buf.alpha = alpha;
+	}
+}
+
+void
+R_Buffer2DQuad(GLfloat ul_vx, GLfloat ul_vy, GLfloat dr_vx, GLfloat dr_vy,
+	GLfloat ul_tx, GLfloat ul_ty, GLfloat dr_tx, GLfloat dr_ty)
+{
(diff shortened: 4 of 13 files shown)


0lvin:
Impressive change, arm panfrost are going from 75fps to 215fps with 640x480 window.

Yamagi:
Let's test with my RaspberryPI 3. That thing is the redefinition of slow. Build the game is painfull and the CPU is the better part of the SoC:

```
# time make -j4
...
===> LD release/quake2
make[1]: Verzeichnis „/home/yamagi/q2/src/gl1-buffers“ wird verlassen

real    4m19,126s
user    16m25,774s
sys     0m36,757s
```

1920x1080 is all the GPU can and its clearly beyond it's capabilities:

* master: 11 FPS
* gl1_buffer: 15 FPS

The numbers don't show how much smoother it is. master has extreme framedrops, gl1_buffers stutters but at least doesn't drop down to under 10 FPS.

Even with 1024x768 the GPU is still slower than my crappy Savage4 I bought in 1999. gl1_buffer yields a huge performance gain and the game becomes playable:

* master: 26 FPS
* gl1_buffer: 40 FPS

Again master has framedrops down to well under 10 FPS, gl1_buffer is smooth.

For comparison the soft renderer yields 12 FPS at 1024x768 without framedrops and is more playable than the GL1 renderer build from master. Despite the way lower average framerate..

I'll delay the 8.40 release for two weeks to get this in.

protocultor:
> I'll delay the 8.40 release for two weeks to get this in.
> 
That's awesome, thank you!

That you still have a Savage 4 at this point in time is awesome.

Now that you mention a Raspberry Pi 3, I also have one, and it's a Tile-Based Rendering system, meaning it doesn't behave well with `glTexSubImage2D()` (yeah, I'm gonna keep hating on it), since active manipulation of textures already being used in the last few frames can cause slowdown on these systems, and that's the cornerstone of dynamic lighting in Q2.

I'm gonna spoil my own surprise here, but in the last few months I've been preparing an OpenGL ES 1.0 renderer, specially meant for old Raspberries. One of the changes it includes is a "multiple copy" of every single lightmap, to avoid what I just mentioned in the last paragraph, by using a different lightmap copy every frame. It rotates the used copy every 3 frames, so it doesn't stumble itself; this helps tremendously with framerate.

I know you people hate adding renderers, but this is not "new" but an alteration of the GL1 one; it has a lot less differences than between GL3 and GLES3. You can test it right now:
https://github.com/protocultor/yquake2/tree/gles1_preview
This is an ugly overwrite of the GL1 renderer, so obviously needs a heavy refactor. And yes, `vid_renderer gl1`.

If you test this with the Pi 3, you'll notice it doesn't get over 30 FPS. That's because GL drivers for it are just not good, no matter the amount of optimizations you do. My GLES1 renderer is meant to be used with the Legacy driver; you can install it using `raspi-config` -> Advanced options -> GL Driver. The framerate increase... will keep it to myself, to not be considered a liar. I just want you to see it by yourself.

If you want to keep delaying 8.40, I'll leave it to you. I just don't think I'll have the next PR in less than two weeks. I'll give it a try, though.

0lvin:
>> While Vulkan might have been a good idea no one was really motivated to maintain it after it was merged...

Vulkan is still alive, it just has no new ground breaking features:-) and all improvements and changes are going to remaster repo. And i really hope that in some not so far future we will have real dynamic lights in Vulkan render :-)

Yamagi:
Did you find a solution to the problem that Vulkan can drop the frame at every API call while Q2 expects the frame to finish in one or the other way? I know that Q2 has this `longjmp()` construct, but that is unreliable and hacky. Especially now where several external libraries require threading.

0lvin:
No, I had not reproduced such errors before, as I remember it was related to Intel driver on windows. Radv or livapipe drivers have not had such errors under Linux.

0lvin:
> Did you find a solution to the problem that Vulkan can drop the frame at every API call while Q2 expects the frame to finish in one or the other way? I know that Q2 has this `longjmp()` construct, but that is unreliable and hacky. Especially now where several external libraries require threading.

Could you please add links to bugs or step to reproduce Vulkan crashes to https://github.com/yquake2/yquake2remaster/issues/26?

Yamagi:
@DanielGibson is still fighting with his RPi4. I'm merging it now at will upload a new Windows testbuild afterwards to get at least some test exposure before the release.

Yamagi:
Mkay, I don't have the SSH key for yamagi.org on my laptop. The testbuild is here: [quake2-20240720-824cd9c.zip](https://github.com/user-attachments/files/16319316/quake2-20240720-824cd9c.zip)

I will upload it to deponie.yamagi.org tomorrow.

protocultor:
> The testbuild is here: [quake2-20240720-824cd9c.zip](https://github.com/user-attachments/files/16319316/quake2-20240720-824cd9c.zip)
> 

This testbuild doesn't include changes from _gl1_buffer_. Not only keeps the "old speed" for GL1, it keeps `gl1_biglightmaps` alive, while _gl1_buffer_ removed it.

Yamagi:
Windows Defender. Again. Another try: [quake2-20240720-824cd9c.zip](https://github.com/user-attachments/files/16321237/quake2-20240720-824cd9c.zip)

protocultor:
Thank you! :)

DanielGibson:
yeah, sorry, my RPi4 somehow always freezes soon after boot, I'll have to reinstall it to see if it's a software problem

0lvin:
@protocultor could you please check model select in multiplayer menu? By some reason I have no menu options, only name of player and player model render visible.

Is it only me?

protocultor:
> @protocultor could you please check model select in multiplayer menu? By some reason I have no menu options, only name of player and player model render visible.
> 
> Is it only me?

No, it is not you. I even tested Zaero with this thing, with hand 0-1-2, with and without mtex, and thought it was ready. Didn't think about checking the simplest user operation.
I'll be looking for the fix between today and tomorrow; I'll do a PR as soon as I find the reason of failure.

protocultor:
All right, this menu in particular encompasses text + player avatar 2D image + the special border made with characters + the animated player model, and the whole mix doesn't exist anywhere else in Q2's code. Now the "GL buffer" is just flushing at any time a change of context occurs (stopped printing characters to show an image or viceversa), plus take into account the call to R_RenderFrame this menu has.

Here's the patch that fixes the issue: [gl1_buffer_fix.patch](https://github.com/user-attachments/files/16322116/gl1_buffer_fix.patch)
@Yamagi, I was embarrassed to make a pull request only for this, so I would appreciate it if you can apply this yourself.
Thanks for the report @0lvin.
Please anyone, tell me anything else you find with this.

DanielGibson:
I installed latest "Raspberry Pi OS", this time 64bit, and it works again.

At 1920x1200 in windowed mode, I get 38.7 fps instead of 23.6fps in demo1.dm2 and 39.9fps instead of 28.2fps in demo2.dm2

I too have the impression that it's much smoother overall, the extreme slowdowns that happened sometimes (due to dynamic lights, I think?) went away. Good job!

Yes, GL3 would also benefit from grouped draw calls, I've had vague plans for this since implementing the GL3 renderer, but so far never had the motivation to actually do this. Maybe one day.. ;)
(At least GL3 doesn't have to update the lightmap textures, because the dynamic lights are applied via shader, and I *did* implement batched drawcalls for models, where it's one drawcall per model instead of dozens or hundreds or whatever it was. Implementing that did really improve performance.)

protocultor:
> At 1920x1200 in windowed mode, I get 38.7 fps instead of 23.6fps in demo1.dm2 and 39.9fps instead of 28.2fps in demo2.dm2
> 
> I too have the impression that it's much smoother overall, the extreme slowdowns that happened sometimes (due to dynamic lights, I think?) went away. Good job!
 
Thanks! Yeah, as I repeteadly say, `glTexSubImage2D()` is the devil, and to me, one of the reasons for the programmable pipeline's existence. Having shaders in GL(ES)3 means one less worry.
Still, I believe the slowdowns can be reduced further. If you have the time and you're in the mood, check my [OpenGL ES 1.0 renderer](https://github.com/protocultor/yquake2/tree/gles1_preview); I gave [my explanation of what it does above](https://github.com/yquake2/yquake2/pull/1124#issuecomment-2231524374).
I almost have the "_multiple lightmap copies_" configurable in a cvar. My idea is having it available in both GL1 and GLES1; it may help people in GL1 with mobile hardware, although it would be obviously disabled by default.

> Yes, GL3 would also benefit from grouped draw calls, I've had vague plans for this since implementing the GL3 renderer, but so far never had the motivation to actually do this. Maybe one day.. ;) (At least GL3 doesn't have to update the lightmap textures, because the dynamic lights are applied via shader, and I _did_ implement batched drawcalls for models, where it's one drawcall per model instead of dozens or hundreds or whatever it was. Implementing that did really improve performance.)

Yeah, I did the same here: 1 model is 1 draw call. I was amazed to find there were many to draw one single enemy.
You're only missing the world and brushes drawing. If you find the motivation, I hope this work can help you.

DanielGibson:
> You're only missing the world and brushes drawing.

IIRC in GL3 2D/UI elements including texts are also still one drawcall per element (for text: per char) instead of one per texture - optimizing that should also really help

protocultor:
Aw yeah, that was my first commit (third in this PR). I remember my Rpi 3 chugging because the console was full with text :)
Take a look at my [Buffer2DQuad() function](https://github.com/yquake2/yquake2/pull/1124/commits/50aebd2de45b8dbcf18e459b38796864534d0fc5#diff-6f2d1cf90a9816be109e0847e8f63dee25ccdb22ceee6ebb20708420dd93ca1eR126).

Yamagi:
> @Yamagi, I was embarrassed to make a pull request only for this, so I would appreciate it if you can apply this yourself.

Applied as 528ffe6a6d5eff14e40f2d8261bd2c0125647a86. Here is another Windows testbuild: [quake2-20240721-528ffe6.zip](https://github.com/user-attachments/files/16323437/quake2-20240721-528ffe6.zip)

This time I checked by hand if the files were copied or if Windows Defender denied it again without printing an error.

0lvin:
> All right, this menu in particular encompasses text + player avatar 2D image + the special border made with characters + the animated player model, and the whole mix doesn't exist anywhere else in Q2's code.

Thank you, I have merged updated code to remaster repository. I have not found any issues for now. It looks as time to make something like it in Vulcan render:-).

@Yamagi What do you think about port some changes from remaster branch back after release?

I have so far,:
* Support player models inside pak files,
* Binary search of the file in pak files,
* Support non numeric names of track files CD_TRACK,
* Full support of all kinds of pcx files version 5 based on gimp and ffmpeg codes.

That looks good features for the base repository.

Pickle:
I ran before and after the glbuffer commits on my atom with ion2 on linux
mint.
Intel i915 Pineview G 18.4 to 20.6
NV GT210 with Nouveau 11.5 to 11.8

Both are pretty bad performers for opengl, but its nice to see a increase.
Id really be interested to see end up in the GLES3 to try on a powervr
system (trimui smart pro) Last build gave at 1280x720 is 52.9 fps
So im expecting this fix would put it over 60 fps.

Yamagi:
> @Yamagi What do you think about port some changes from remaster branch back after release?
> 
> I have so far,:
> 
>     * Support player models inside pak files,
> 
>     * Binary search of the file in pak files,
> 
>     * Support non numeric names of track files CD_TRACK,
> 
>     * Full support of all kinds of pcx files version 5 based on gimp and ffmpeg codes.
> 
> 
> That looks good features for the base repository.

Sound good. Can you provide a pull request?

PR #1128, opened 2024-07-28
OpenGL ES 1.0 renderer

protocultor:
Nicknamed **_GLES1_**, it's completely dependent on GL1; it consists of 15 `#ifdef` and 5 `#ifndef` (`YQ2_GL1_GLES`) over GL1 in total.

**Limitations**

- Doesn't support visual stereo modes, at all.
- No texture modes supported, for alpha or solid, other than `GL_RGBA`.
- Just like with GLES3, `gl_drawbuffer` is not supported.
- `vid_gamma` doesn't work. This is the exact same issue GL1 has with SDL3 (probably my next project...)

Note that it does compile on Windows with MSYS2, but you'll need a decent implementation of ANGLE to run it; I'd say is not worth it. I haven't tested in MacOS, since I don't have a Mac anymore. Since this is meant for SoCs like Raspberry Pi, Linux is the proper OS to use this. Compilation is done using `make with_gles1`; this will generate all 5 renderers available now in the main project (_gl3_, _gles3_, _gl1_, _gles1_, _soft_).
_CMake_ was only tested on Linux, and it does work there, since the basic headers for GLES1 are normally included together with the desktop GL ones. Since I couldn't test it with any other OS, I opted to deliver this commented out at the end of `CMakeLists.txt` (actually enclosed with an `if(FALSE)`, but works the same).

**New cvars**

- `gl1_discardfb`: available only in GLES1, sends a hint to the GPU to discard the contents of depth and stencil buffers at the end of each frame. Didn't make any difference in my Raspberry Pi 3 B+, but it was an advice I took after seeing it repeteadly over the net, so I assume there's hardware where this is needed.
- `gl1_lightmapcopies`: now this is useful. Available in both GL1 and GLES1, maintains 3 copies of each lightmap in video memory (meaning: they are "different textures"), and rotate them, using a different one each frame. _Tile-Based Rendering_ GPUs, like those in mobile and embedded devices, are downright killed when there's an attempt of `glTexSubImage2D()` on one of the textures that has been just used in the last few frames, so with this workaround we don't have to use the same lightmap texture we just used in the last frame, but a "different" one, a copy. Note that this backfires on _Immediate-Mode Rendering_ GPUs (like traditional desktop cards), so use the option that makes the most sense for your hardware. For this reason, by default this is enabled in GLES1 and disabled in GL1.

If you're going to test this with a Raspberry Pi 3 or older, please use the _Legacy_ driver, not one of the GL ones. The GL ones are to support traditional, desktop OpenGL, but they are **slow**, no matter what you do for optimization. Legacy driver doesn't support desktop OpenGL, but does support OpenGL ES (1 & 2), and does it very fast; this renderer was mostly meant for use on this. And of course, make sure `gl1_lightmapcopies` is `1`.

Credits to @thenesis-org for the lightmap copies' thinking.


Yamagi:
Do you have a link or something what I need to do to active the legacy driver on my RPi3? RaspiOS Bookworm doesn't seem to ship it anymore, Bullseye ships at least the kernel part and there's a option in `raspi-config` to activate the driver. But no matter what I do, ref_gles1 ends up with llvmpipe instead of hardware rendering. I haven't tried Buster so far.

protocultor:
Buuh, I just use Buster because that's what the [RetroPie](https://github.com/RetroPie/RetroPie-Setup) project uses, and did not realize how outdated it is. It seems like the Raspberry Pi Foundation (and Broadcom) are just forgetting about legacy hardware; shame about the lack of support.
I know it's out of LTS, but I'd still recommend Buster as long as you don't connect it to a critical server or something.
I've just uploaded a video of this on YT; give it a few hours and you'll see how it works in Buster. Warning, I don't have capture equipment and I'm an horrible cameraman, but should be a decent enough show on how this runs:
https://youtu.be/eeijJNQFGKo

Yamagi:
I got it working. The documentation is even worse than their support for old hardware:

1. X11 must not be running. There's not GLX module for the legacy driver. X11 always loads Mesa GLS and that makes the Broadcom EGL modul unavailable.
2. On Buster it's enough to select the Legacy driver and reboot. Newer (or all?) Bullseye images don't ship the userland parts of the Legacy driver. They can be downloaded from https://github.com/raspberrypi/firmware/tree/oldstable and must be placed into /opt/vc. Other directories don't work. Bookworm is missing the kernel part, no chance there.

Performance is impressive, I get 96 FPS on my RPi 3B+ at 1280x1024. If with 4x MSAA it's still 85 FPS. I'll do more thoroughly testing next week or so.

What I noticed so far:

* Linking `-lGL` in line 662 of the Makefile looks wrong. Maybe `-lGLESv2`, but do you need to link the library at all? Since you are using glad it might not be needed.
* The already overcomplicated renderer fallback logic selects GL3 if GLES1 failed. A system requiring GLES1 won't support GL3 in most or even all cases. Better fall back to GL1 or directly to the software renderer?
* `demomap demo2.dm2` (Pumping Station) breaks things badly at the very first frame. The world model is missing, the console and menu are broken afterwards. Maybe some kind of messed up GL state.
* `map waste1` ends up in some kind of endless loop. The only way out is to kill the game.

Pickle:
its working on my intel pineview at 1366x768 at about 40 fps in the demo. Pretty sad its being beat by a RPi 3B+.
but i switched to linux mint debian based and its giving much better numbers than the previous linux mint ubuntu based was giving.

protocultor:
>     1. X11 must not be running. There's not GLX module for the legacy driver. X11 always loads Mesa GLS and that makes the Broadcom EGL modul unavailable.
> 
>     2. On Buster it's enough to select the Legacy driver and reboot. Newer (or all?) Bullseye images don't ship the userland parts of the Legacy driver. They can be downloaded from https://github.com/raspberrypi/firmware/tree/oldstable and must be placed into /opt/vc. Other directories don't work. Bookworm is missing the kernel part, no chance there.

You just have written the best documentation this hardware has :)
I completely forgot about X11, shame it doesn't support the legacy driver.

>     * Linking `-lGL` in line 662 of the Makefile looks wrong. Maybe `-lGLESv2`, but do you need to link the library at all? Since you are using glad it might not be needed.
> 
>     * The already overcomplicated renderer fallback logic selects GL3 if GLES1 failed. A system requiring GLES1 won't support GL3 in most or even all cases. Better fall back to GL1 or directly to the software renderer?
 
Thanks for the heads up, both have been fixed in the latest commit.
Renderer fallback logic is now: _custom -> gl3 -> gles3 -> gl1 -> gles1 -> soft_.
In Buster, outside of X11, it crashes when it arrives at _gl1_ though, but that has always been like that, surely is a driver limitation. In X11, as you say, uses Mesa for _gl1_, but at least it does get to _soft_ then attempting _gles1_.

>     * `demomap demo2.dm2` (Pumping Station) breaks things badly at the very first frame. The world model is missing, the console and menu are broken afterwards. Maybe some kind of messed up GL state.
> 
>     * `map waste1` ends up in some kind of endless loop. The only way out is to kill the game.

Couldn't reproduce neither of these on my end. All demos (official, at least) always worked fine in Buster, and the `waste1` situation sounded like the issue fixed by #1133; I tested it but didn't hang for me. I still merged master back just in case (should have done a rebase though...)
I'll install Bullseye next week (maybe even Batocera) and try to reproduce these to see if there's something that can be done.

protocultor:
> its working on my intel pineview at 1366x768 at about 40 fps in the demo. Pretty sad its being beat by a RPi 3B+. but i switched to linux mint debian based and its giving much better numbers than the previous linux mint ubuntu based was giving.

Forgot to mention something; I don't know what _Pineview_ is, but a suggestion I can give is: try with `gl1_lightmapcopies 0` and `1`, in either GL1 or GLES1, and see what performs better for you. Don't forget to `vid_restart` when changing it.

====================================================

About the RPi 3B+, I didn't know about [PINN](https://sourceforge.net/projects/pinn/) and I'm glad I found it, because this allowed me to quickly test GLES1 (and GL1) in 4 different OSs. Actually installed 5 OSs, but one is "uncooperative". This is my experience with them (no _X11_ here).

### RetroPie 4.8 (Buster)
My "_comfort zone_", I just installed it again to test if what I had with GLES1 can be easily replicated, and it can be. Runs over _Buster_, and as shown in the youtube video, only supports 1920x1080 in fullscreen, but thats all it needs because runs over 60 fps most of the time. Only in the first demo of Xatrix (`xdemo1.dm2`) drops to 50 in a couple of busy moments, but that's very uncommon. Particles appear as circles, handles MSAA really well. IMHO where this renderer really shines. On the other hand, the GL1 renderer crashes here.
```
-------- vid initialization --------
SDL version is: 2.0.10
SDL video driver is "RPI".
SDL display modes:
 - Mode  0: 1920x1080@60
...
GL_VENDOR: Broadcom
GL_RENDERER: VideoCore IV HW
GL_VERSION: OpenGL ES-CM 1.1
```
### Bullseye
Enabled by Yamagi's trick of using `/opt/vc` from raspi's firmware repo. Runs "not as well", but performance is still very good. Stutters at 1080p, and cannot disable vsync, but at 1280x720 never, **never** drops below 60 fps, not even in the mentioned `xdemo1.dm2`. Also, the mentioned resolution runs in fullscreen, unlike in Buster. Sometimes when playing with the video options loses the fullscreen, but putting it in 1080p and then dropping the resolution again solves it. Particles appear as squares, and performance dies when attempting MSAA. Still a pretty good show for GLES1.
Something funny: GL1 does work here, and features/extensions don't seem to differ from GLES1. Seems like it goes at exactly the same speed as GLES1, so choosing one over the other doesn't matter. What does matter is the usage of `gl1_lightmapcopies`, because disabling it make the fps drops noticeable in either renderer; it loses the "locked to 60 fps" behavior in 720p.
```
-------- vid initialization --------
SDL version is: 2.0.14
SDL video driver is "KMSDRM".
SDL display modes:
 - Mode  0: 1920x1080@60
...
 - Mode 28: 640x480@60
...
GL_VENDOR: Broadcom
GL_RENDERER: VC4 V3D 2.1
GL_VERSION: OpenGL ES-CM 1.1 Mesa 20.3.5	(<- for GLES1)
GL_VERSION: 2.1 Mesa 20.3.5			(<- for GL1)
```
### Buster BETA
installed this by accident, thinking it was the release version (I misunderstood PINN's description, which is not the clearest). And _beta_ it is, because it doesn't work; only videos and general 2D elements are shown, and even these elements glitch. Seems like GL state is nonexistent in this. And just like Buster release, GL1 crashes. The worst, don't bother.
```
-------- vid initialization --------
SDL version is: 2.0.9
SDL video driver is "RPI".
SDL display modes:
 - Mode  0: 1920x1080@60
...
GL_VENDOR: Broadcom
GL_RENDERER: VideoCore IV HW
GL_VERSION: OpenGL ES-CM 1.1
```
### Bookworm
What Yamagi said, doesn't even start. However, GL1 does function here again, and seems to work at the same speed as Bullseye's renderers, if not slightly faster, with the addition that in here you can disable vsync; maybe that's the reason it feels faster. Having no vsync helps verify how practical it is to have `gl1_lightmapcopies` available in GL1, since in `demo1.dm2` we get **94** fps when `gl1_lightmapcopies` is `0`, and **114** fps when it is `1`; all of this is, again, at 1280x720 fullscreen (drops to 54 fps at 1080p). So, I still see it as a win for the existence of GLES1 :) . Again, as Bullseye's renderers, particles are square and MSAA is slow.
```
-------- vid initialization --------
SDL version is: 2.26.5
SDL video driver is "KMSDRM".
SDL display modes:
 - Mode  0: 1920x1080@60
...
 - Mode 28: 640x480@60
...
GL_VENDOR: Broadcom
GL_RENDERER: VC4 V3D 2.1
GL_VERSION: 2.1 Mesa 23.2.1-1~bpo12+rpt3
```
The game in _Bookworm_ does drop a scary message sometimes in the shell (not in game's console): `Draw call returned  Invalid argument.  Expect corruption.` ...but that corruption never came. Speaking of that: none (except _Buster Beta_) gave me a bad GL state at any moment, so I still cannot replicate @Yamagi 's issues. Any info you have on this, please share.

Have to mention that I copied `/opt/vc` from Buster / RetroPie to all other OSs, and it didn't make any difference; certainly the kernel plays a bigger part on this than I though. Also have to mention that, in all of them and very occasionally, the game will hang for a couple of milliseconds and then continue with high fps as if nothing happened; this may be my power supply's fault...

The last OS was **_Batocera 39_**, but couldn't do anything with it since I don't have the resources to set up [its complicated cross-compiling development environment](https://wiki.batocera.org/compile_batocera.linux) so, job for another person.

In the end, even if the newer OSs support GL1 and would make GLES1 redundant, none of them are as fast (and clean looking) as GLES1 running on Buster, so IMHO is still worth keeping. Had [feedback from a RetroPie user that says the game "works smooth at 1920x1080" on a Pi Zero 2 W](https://retropie.org.uk/forum/post/299006); this is, again, with Buster.

I'm grateful of any feedback I receive of this, so please tell me if there's anything I'm missing.

Yamagi:
@protocultor, are you okay with merging this?

protocultor:
> @protocultor, are you okay with merging this?

If you don't mind, I want you to wait for a day, just to test how well the recently reduced **glad** works in the different Pi OSs.

protocultor:
Well, it took less time than expected. No evident issues came up with the latest change, in any OS.
I could even test **[Arch](https://archlinuxarm.org/platforms/armv8/broadcom/raspberry-pi-3)**, and the results are exactly the same as Bookworm, with a single difference: GLES1 "starts" but it only comes as a black screen; only the game audio can be heard. GL1 works, at the same speed as Bookworm, and also benefits from `gl1_lightmapcopies`.
I am now okay with the merge. Thanks.

Yamagi:
> I could even test [Arch](https://archlinuxarm.org/platforms/armv8/broadcom/raspberry-pi-3), and the results are exactly the same as Bookworm, with a single difference: GLES1 "starts" but it only comes as a black screen; only the game audio can be heard.

The normal x86 Arch is somewhat special under the Linux distro in not shipping the GLES1 Mesa library and it's headers. That should interfere with Broadcoms video core blob, though. As an educated guess I would say that the blob is build against RaspOS and cannot really work on other distros, at least not without using the RaspOS kernel and fiddling with libraries.

protocultor:
Just putting this here for posterity: using `r_clear 1` on the RPi provides an 8 to 10% framerate improvement; this applies to any OS, in either `gl1` or `gles1`. Maybe this was the real effect that `gl1_discardfb` was after.
Only side effect so far: cinematics start with a pink texture.

protocultor:
Turns out, the `glClear` trick was a known one. #1152 includes a way that doesn't clash with traditional `r_clear`. Using a black color for the glClear makes the thing very rarely noticeable.

> Also have to mention that, in all of them and very occasionally, the game will hang for a couple of milliseconds and then continue with high fps as if nothing happened; this may be my power supply's fault...

It wasn't, **OpenAL** is the culprit. Using the SDL sound system (`s_openal 0`) makes the game run like butter.
I tested some settings on `~/.alsoftrc` but had no luck with anything I tried; it either doesn't solve the problem, or adds a noticeable delay to every sound. If somebody knows some safe settings, please share.
No big loss in any case.

Yamagi:
I somehow lost track of this bug...

> Using a black color for the glClear makes the thing very rarely noticeable.

...I wanted to say, that the original map compiler used a rather simple flooding algorithm prone to miss small surfaces to calculate the visibility sets. I guess that clearing to black should be okay, because the game is mostly blackish and the problematic surfaces aren't that easy to spot anyways.

> It wasn't, OpenAL is the culprit. Using the SDL sound system (s_openal 0) makes the game run like butter.

Maybe something in the SoC - the CPU, memory controller, etc. - is too slow to handle Quake II itself and OpenAL? Quake II itself is single threaded, OpenAL runs inside the main thread and is somewhat CPU and memory bandwidth demanding... After your report and the discussion in #1135 I wonder if should replace the *Sound Quality* setting in the options menu with a sound backend selector? These days only masochists and telephone phreaks are using 11 kHz sounds, they could still set it manually through the console.

protocultor:
> ...I wanted to say, that the original map compiler used a rather simple flooding algorithm prone to miss small surfaces to calculate the visibility sets. I guess that clearing to black should be okay, because the game is mostly blackish and the problematic surfaces aren't that easy to spot anyways.

Well, using black is at least less ridiculous than the pink that's normally used :) . That makes every missing pixel to stand out easily.
I can only see it on the edges of the skybox cube, at the borders of liquid surfaces, and very rarely, when pulling the main menu in demos; it flickers for a couple of seconds, then it stops and gets back to normal.

> After your report and the discussion in #1135 I wonder if should replace the _Sound Quality_ setting in the options menu with a sound backend selector?

That's a good idea. Another way would be to include the sound backend selector together with the quality setting, and to not clutter the main menu, delete the "_pause on minimized_" option added in #1017. As you said yourself in that PR, that's an option very few users would worry about changing, and can be done through the console.

PR #1130, opened 2024-07-29
Backport changes from remaster repository 

0lvin:
* Use hash from string entities as part of .ent filename. 
* Support player models inside pak files,
* Binary search of the file in pak files,
* Support non numeric names of track files CD_TRACK,
* Support of pcx files version 5 based on gimp and ffmpeg codes.


0lvin:
Classic yquake2 code could load old maps from remaster release, it has issues with lights, but player could pass all old mods in theory. In such case player should remove all old .ent files as updated maps have different list of entities.

BjossiAlfreds:
What is the hash function and what is being hashed? The entire content of the ent file? And what is the point of this? Is the remastered version looking for a specific hash value and rejects the load if the hash doesn't match?

In any case this should be optional for YQ2. If there is no hash in the name, load it and leave the responsibility to the user.

0lvin:
Content of LUMP_ENTITIES lump of map is hashed and used for search related override.ent file, as result ent file has `<map name>@<hash of content>.ent` name template. 

It resolves case when map has same name but different lump content, like case when ent files for classic quake 2 put to directory with remaster pak.

It's replication of [QuakeSpasm](https://github.com/yquake2/yquake2remaster/issues/4#issuecomment-1741814349) logic proposed by @SirYodaJedi

BjossiAlfreds:
Thanks for the clarification! I definitely see the benefit of this approach now.

I have an idea, this may be your plan already but just in case it's not - when the CRC is calculated, write its value to a read-only cvar called for example "maphash". This would make it quick and easy to get the map hash to write into new ent filenames.

0lvin:
> I have an idea, this may be your plan already but just in case it's not - when the CRC is calculated, write its value to a read-only cvar called for example "maphash". This would make it quick and easy to get the map hash to write into new ent filenames.

Code writes expected file name with hash to logs, so by warning message could be checked what code expects as replacement and what hash should be.

SirYodaJedi:
> support 24bit pcx images 

Neat feature, but could confuse new modders. From @Paril on Map-Center discord:
image
image

Perhaps have a warning when loading a true-color PCX if `developer 1` is enabled?

0lvin:
> > support 24bit pcx images
> 
> Neat feature, but could confuse new modders. From @Paril on Map-Center discord: image image
> 
> Perhaps have a warning when loading a true-color PCX if `developer 1` is enabled?

Sure I will add warning if pcx file with uncommon flags.

src/client/menu/menu.c
```diff
@@ -76,13 +76,13 @@ qboolean m_entersound; /* play after drawing a frame, so caching won't disrupt t
 
 typedef struct
 {
-    void (*draw)(void);
```
Yamagi, line 79: I know that the usage whitespaces instead of tabs in menu.c is annoying. Even more annoying is that I messed this up ~15 years when I reformated the code to a consistent style with `uncrustify`. But I think it's too late to change it to tabs, because it touches every line and breaks `git blame` and the like. With is very useful with the  error prone menu code and it's pointer magic. Therefor I would like to keep this file whitespace intended.

src/game/savegame/savegame.c
```diff
@@ -831,6 +831,7 @@ ReadGame(const char *filename)
 		{"YQ2-3", 3},
 		{"YQ2-4", 4},
 		{"YQ2-5", 5},
+		{"YQ2-6", 6},
```
Yamagi, line 834: Is adding the `music` field to `spawn_temp_t` backward compatible or does this mess with the overall state when loading a v5 savegame into a v6 game library?

src/client/menu/menu.c
0lvin: I have removed code style changes commit, little code style change left only for functions where I have logic changes.

src/game/savegame/savegame.c
0lvin: I have not found any issues with load 'music' in code without such commit and with it. It was mostly safety check, I have removed this change for now. I suppose we could create such check in future if someone have found some issue.

BjossiAlfreds:
Changing spawntemp is safe. It is only used during `SpawnEntities` to catch fields that are not mapped to `edict_t`. In my own dll I got rid of it entirely, replaced it with a dynamic dictionary data structure.

Yamagi:
Thank you for the changes and the confirmation that a new field in `spawn_temp_t` is harmless. I'll merge it, let's see who complains. :)

src/client/menu/menu.c
turol: It's possible to configure git to ignore certain commits
https://stackoverflow.com/questions/34957237/can-i-configure-git-blame-to-always-ignore-certain-commits-want-to-fix-git-blam

Issue #1131, opened 2024-07-30
Clipboard support in console


OpenRift412:
Just a minor QoL feature. Weirdly enough, this is a feature in vanilla 3.20, but somehow isn't here? Is there a reason for that?

protocultor:
"Extension" of #964 ?
A possible reason it works in 3.20 is that the game probably uses native / OS input for keyboard, while YQ2 uses SDL for cross-platform compatibility.

BjossiAlfreds:
It shouldn't be difficult to set up a wrapper function that hides platform dependent implementations using ifdefs. The Wibdows API is fairly simple. The other part to this is detecting Ctrl+V input. I haven't looked if this code is already in place or we need to write it ourselves.

BjossiAlfreds:
I spent some time studying the keyboard/console code and experiments today and got CTRL actions in console to work again and also made it possible to move the cursor inside the edit line without erasing using the left/right arrow keys.

It seems that when CTRL is held down, pressing other keys counts as a key up/down event rather than text input. This causes `Key_Event` to be called with `special == false` which prevents the event from reaching `Key_Console`. I don't understand this code well enough to address this properly so I did a small hack that flips `special` to `true` if CTRL is being held while in the console and key is an alphabet letter.

I then wrote quick little code that uses `SDL_GetClipboardText()` and attached it to CTRL+v.

Finally, since I had seen requests for this before, I separated backspace from left/right arrow so you can move the cursor inside the edit line.

I got this in my con-ext branch: https://github.com/BjossiAlfreds/yquake2/tree/con-ext

If you are interested @Yamagi I can PR. Needs thorough testing.

Yamagi:
A long time ago I planed to replace the home grown line editor in the console with linenoise: https://github.com/antirez/linenoise For one reason or the other it never happened.

> If you are interested @Yamagi I can PR. Needs thorough testing.

Yes please :) We should merge this early in the current release because, because as you said. It needs testing, the console is fragile...

jobukkit:
Would be nice for pasting server IP's into the address book too.

BjossiAlfreds:
> Would be nice for pasting server IP's into the address book too.

I added this feature couple months ago (along with CTRL+c/x and in-line editing), don't know if it has made it to a release yet though.

BjossiAlfreds:
[...]

PR #1134, opened 2024-08-07
Console clipboard support and input nav

BjossiAlfreds:
This PR addresses https://github.com/yquake2/yquake2/issues/1131 and other similar feature requests in the past.

* 2 new string functions in shared that might come in handy elsewhere:
- `Q_strdel` for deleting from a string.
- `Q_strins` for inserting into a string.

* Functions for clipboard access (using SDL):
- `IN_GetClipboardText`
- `IN_SetClipboardText`

* Left/right arrow navigation. Previously, left arrow did the same thing as backspace, but now you can navigate left and right through the input without erasing input.

* Restored CTRL actions in console. CTRL+L clears the console, and CTRL+V pastes text from clipboard into the current input.

* Added CTRL+c that copies the currently typed input to the clipboard.

* Added CTRL+x that does like CTRL+c and then clears the input.

* Added delete key functionality. Behaves like in text editors (delete currently highlighted character).

`Con_DrawInput` was the main reason for the limitations with the console input line. It was actually modifying the input buffer. Drawing code modifying its input is a big red flashing no-ono in my book so I rewrote it to be read-only and that made left/right arrow nav possible.

@Yamagi, I noticed the input line indicator `"]"` is currently part of the text buffer as the first character. This makes the code more fragile and hacky. I'd like to fix this but it would mean the ] is no longer part of the lines in console_history.txt. If that is not an issue then if you want I could move ahead with this change. I did a test implementation and it went smoothly.


Yamagi:
> I noticed the input line indicator "]" is currently part of the text buffer as the first character. This makes the code more fragile and hacky. I'd like to fix this but it would mean the ] is no longer part of the lines in console_history.txt.

It's okay for me. Do we need some backward compatibility logic in `Key_ReadConsoleHistory()` or doesn't the "]" do any harm when read back from the file into the buffer, if "]" is no longer part of the buffer?

BjossiAlfreds:
The code always read the line buffer from index 1 rather than index 0, to account for the `"]"`. Moving the indicator to `Con_DrawInput` and reading the buffer from index 0 does mean the `"]"` becomes a problem. But a simple if-check with a `memmove` call is all we need for backwards compatibility.

I pushed a commit that carries out this change. In my testing everything seems to work good. Tested:

* Auto-complete commands and map names
* Executing typed-in commands with the enter key
* How the input line drawing reacts when the string length exceeds console liewidth
* Reading both old and new console_history.txt
* Making sure left arrow and backspace can reach index 0 (and not underflow it)

I use grep to exhaustively investigate all references to `key_lines` arrays and followed any local aliases as well, so I hope I didn't miss anything or introduce any side-effects.

Yamagi:
Thank you for implementing this. :) Let's merge it and see if something breaks.

PR #1138, opened 2024-08-11
[WIP] CI FreeBSD.

devnexen:
inspired from ubuntu build.
note: could be done via Cirrus instead but is to keep within github actions for consistency.

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -1,0 +1,65 @@
+name: Testbuild for FreeBSD
+run-name: testbuild_freebsd
+on:
+  push:
+    branches:
+      - 'master'
+  pull_request:
+    types:
+      - opened
+      - edited
+concurrency:
+  # Cancel concurrent workflows for the same PR or commit hash.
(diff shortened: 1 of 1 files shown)


Yamagi:
This cannot work, because Github doesn't offer FreeBSD workers. I pondered if I should add FreeBSD and maybe OpenBSD through Cirrus but decided against it. In my experience FreeBSD and OpenBSD build when Linux does and anyone who uses these systems is likely skilled enough to build by hand.

devnexen:
> This cannot work, because Github doesn't offer FreeBSD workers. 
it is possible but you need [this](https://github.com/yquake2/yquake2/pull/1138/files#diff-6f2adbfa3a8863a041c5953f37db9c434866c90dd11b5b866c23c1c5a87f3e5cR25).
I pondered if I should add FreeBSD and maybe OpenBSD through Cirrus but decided against it. In my experience FreeBSD and OpenBSD build when Linux does and anyone who uses these systems is likely skilled enough to build by hand.

Oh I see, fair point.

Issue #1139, opened 2024-08-12
[Original bug] Spawntemp data usable after level load


BjossiAlfreds:
If the very last entity in the entity string contains spawntemp fields, like "item", it will stay for the rest of the level and all entities spawned mid-game can read the field value.

Example steps to reproduce:
1. Add "item" "ammo_cells" to the last entity in map fact2's entity string.
2. Start the map and go to where the misc_insane guys are falling into the slime.
3. Notice that they all drop cells when they die.

I will do a PR that addresses this issue.

Yamagi:
Fixed with the merge of #1140

PR #1140, opened 2024-08-12
Fixed spawntemp data leaking into mid-level spawned entities

BjossiAlfreds:
This PR addresses https://github.com/yquake2/yquake2/issues/1139.

This happens because `st` is reset 'before' parsing each entity in the string, so after the very last entity has been parsed, `st` won't be reset and its data remains for the rest of the level, accessible to entities spawed by for example a target_spawner, or the makron.

The fix was to simply add a `memset` after the spawn loop in `SpawnEntities`.

This issue can also apply to child entities, like for example door triggers. But those don't use any spawntemp fields as far as I know, but still worth noting for future reference.

### DIFF
--- a/src/game/g_spawn.c
+++ b/src/game/g_spawn.c
@@ -700,6 +700,9 @@ SpawnEntities(const char *mapname, char *entities, const char *spawnpoint)
 		ED_CallSpawn(ent);
 	}
 
+	/* in case the last entity in the entstring has spawntemp fields */
+	memset(&st, 0, sizeof(st));
+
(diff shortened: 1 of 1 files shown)


Yamagi:
I'm always amazed what obscure bugs your are still finding after all these years :)

Do we need the same for the addons?

Thank you!

BjossiAlfreds:
Yeah this is relevant to both rogue and xatrix and most likely zaero as well. I'll do PRs for those soon.

PR #1141, opened 2024-08-17
GL4: Enable sliders for overbrights, colours and intensity

atsb:
This will match the vk and gl3 menu sliders for their respective video modes.


Yamagi:
Thank you :)

PR #1142, opened 2024-08-17
AI, HuntTarget fix runtime warning.

devnexen:
`Conditional jump or move depends on unitialised value(s)...HuntTarget (g_ai.c:423)...`


Yamagi:
Thank you :) Do we need the same for the two official addons?

devnexen:
only xatrix it seems

PR #1143, opened 2024-08-19
sound: True randomness

apartfromtime:
True randomness allows producing a number twice.


Yamagi:
One can argue if the same track should be played several times in a row... I'm gonna merge it, let's see if anyone complains.

Yamagi:
I have moved true random track selection to a new mode `4`. Mode `3` is again normal random track selection which never plays a track twice in a row.

PR #1136, opened 2024-08-10
Update some SDL3 calls

ZNixian:
SDL3's API is still changing, and there's been a few breaking changes made since someone last worked on the SDL3 code here.

This series makes a bunch of small changes to get it to build and run on SDL3 commit ca2c9f680be4.

PR #1146, opened 2024-08-20
Partial fixes for SDL3

erysdren:
I got it building and running with the current revision of SDL3. This PR is a draft because I haven't figured out how to fix `SDL_StartTextInput()` in `/src/client/input/sdl3.c` - it now requires an SDL_Window pointer, which can't gracefully be pulled from any of the renderers as far as I can see. I'm hoping for some guidance on that point.

Thanks,
erysdren


protocultor:
I know it's just a draft, but as SDL3 is currently a work in progress, it would be really cool if you could mention the specific SDL commit your work targets. The "current revision" changes by the minute :)
Check #1136.
Your work feels worthy of being added to [this branch](https://github.com/yquake2/yquake2/tree/sdl3-20240811).

erysdren:
Looks like the last time I rebuilt SDL3 on my machine was [a7bed81](https://github.com/libsdl-org/SDL/commit/a7bed810b3499d157655ae6dcc7cd5e8a34fa549)

Yamagi:
> I got it building and running with the current revision of SDL3. This PR is a draft because I haven't figured out how to fix `SDL_StartTextInput()` in `/src/client/input/sdl3.c` - it now requires an SDL_Window pointer, which can't gracefully be pulled from any of the renderers as far as I can see. I'm hoping for some guidance on that point.

The SDL_Window is a global variable defined in the client in glimp_sdl3.c: https://github.com/yquake2/yquake2/blob/master/src/client/vid/glimp_sdl3.c#L55 Accessing it from the input backend isn't a problem. But what happens if we recreate the window? If text input must be reenabled every time it might be better to move the `SDL_StartTextInput()` call into `CreateSDLWindow()` and call unconditionally as soon as have a windows.

erysdren:
@Yamagi any further changes needed? thanks!

PR #1147, opened 2024-08-27
Flick Stick fixed duration to 0.1 seconds

protocultor:
#883 introduced an implementation of _Flick Stick_ for gyro gamepads that only targeted 60 Hz, vsync displays. This is because it used a fixed "_6 frames_" factors for angle change (6 * 16.6 ms = 100 ms), so if those frames lasted something other than the intended 16.6 ms each, the flick duration will be way too fast. The present PR solves this by calculating the angle change, using the time difference from the current frame vs. the last one. This allows for the flick to last the intended 100 ms, independent of the display refresh rate. Credits and thanks to the [JoyShockMapper](https://github.com/Electronicks/JoyShockMapper) team.

Extras:
- #1130 added two `PCX_Decode()` functions, which spam the following warning in at least one mod: `(texture.pcx) file has possible size issues`. It does it with the [Full MD2](https://www.nexusmods.com/quake2/mods/45) mod, which works OK under both Yamagi and the remaster, so I've "demoted" the warning to `developer 1` output only. Hope you don't mind @0lvin; I believe if there's a real problem with a texture mod, we will not need a console message to find out.
- Changed the default button for the `Esc` key in gamepads, from `Select` / `Back` to `Start`. It's what the remaster uses, and let's face it, what most people expect.
- At least for SDL 3, changed some text referring to a "_game controller_", to "_gamepad_". After all, this is the nomenclature SDL 3 is trying to push, so changed some `Printfs` to reflect it. In backend they are still referred as "_controller_" though.

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -570,12 +570,11 @@ it's `+set busywait 0` (setting the `busywait` cvar) and `-portable`
   `1`, which enables gamepad usage; `0` disables its detection at
   startup. Can only be set from command line.
 
-* **in_sdlbackbutton**: Defines which button is used in the gamepad or
-  joystick as the `Esc` key, to access the main menu and 'cancel' /
-  'go back' on its options. Default is `0`, which corresponds to the
-  Back/Select/Minus button. Set to `1` to use Start/Menu/Plus, and to
-  `2` to use the Guide/Home/PS button. Requires a game restart
-  (or controller replug) when changed.
+* **in_sdlbackbutton**: Defines which button is used in the gamepad as
+  the `Esc` key, to pull the main menu and 'cancel' / 'go back' on its
+  options. Valid values are `0` = Back / Select / Minus, `1` = Start /
+  Menu / Plus (default), or `2` = Guide / Home / PS. Requires a game
+  restart, or gamepad replug, when changed.
 
 * **joy_layout**: Allows to select the stick layout of the gamepad.
   - `0`: *Default*, left stick moves, right aims
--- a/src/client/cl_image.c
+++ b/src/client/cl_image.c
@@ -368,7 +368,7 @@ PCX_Decode(const char *name, const byte *raw, int len, byte **pic, byte **palett
 
 	if (image_issues)
 	{
-		Com_Printf("%s: %s file has possible size issues.\n", __func__, name);
+		Com_DPrintf("%s: %s file has possible size issues.\n", __func__, name);
 	}
 }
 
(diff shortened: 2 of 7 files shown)


0lvin:
I think it's just case when pic width != scan line size. Looks like some editors round up width to next even number. Gimp and ffmpeg have ignored such cases, old quake 2 code didn't like such case.

protocultor:
New _extra_: added a new way to get the renderer used, that allows the usage of `switch` statements in videomenu code.
The reason: #1141 lacked a simple `else` before a newly added `if`, that in turn made the final `else` of the chained `ifs` to be run when it shouldn't. That made appear an empty space of two lines (or so) in the middle of the video menu when using GL3. The proposed way should help ordering the different options of each renderer, avoiding these kind of conflicts.

apartfromtime:
> I think it's just case when pic width != scan line size. Looks like some editors round up line scan to the next even number. Gimp and ffmpeg have ignored such cases, old quake 2 code didn't like such case. If new mods use such optimized scan line, as for me it's fine to ignore such warning and show only for some one who works on creation of new mods.

The bytes per line of a .pcx must be an even number. Old q2 used the width (which can be odd), since all q2 .pcx files where colour index and power of two. A robust decoder should keep source and destination bytes independent.

0lvin:
I have checked code, we don't have check that width == scan line, so its something different.

0lvin:
Images in mod has no palette marker that should be before palette content. As described in wiki https://en.m.wikipedia.org/wiki/PCX:
```
If a PCX file has a 256-color palette, it is found 768 bytes from the end of the file. In this case the value in the byte preceding the palette should be 12 (0x0C).
```

Code in https://github.com/yquake2/yquake2/pull/1148/commits/a2536cc735574435f953c2e61b2145a57974018c adds code for ignore such case.

protocultor:
> @protocultor could you please check that it fixes issues?

Tested your `backport` branch, and it does work; `(texture.pcx) has no palette marker` message appears only when `developer 1`.
Do you want me to undo https://github.com/yquake2/yquake2/pull/1147/commits/4756c44155202cdb60e2d4eb31994614c569f366 ?
Thanks.

0lvin:
> > @protocultor could you please check that it fixes issues?
> 
> Tested your `backport` branch, and it does work; `(texture.pcx) has no palette marker` message appears only when `developer 1`. Do you want me to undo [4756c44](https://github.com/yquake2/yquake2/commit/4756c44155202cdb60e2d4eb31994614c569f366) ? Thanks.

Thank you! Yes or cherry-pick this change, if you like. Just to catch any other possible regression with pcx image decode.

protocultor:
Nah, I trust your PR will get merged eventually.

Yamagi:
Thank you both for your continued efforts regarding gamepad / game controller support and ironing out corner case problems :)

PR #1148, opened 2024-08-27
Fix memory leaks

0lvin:
* Fix memory leak with merge skin files names in menu in #1130 
* Fix file header leak for empty file
* Small cleanup for light map format in gl* renders.
* Fix crash received message with entity id bigger than static buffer of entities.


Yamagi:
Thank you :)

PR #1149, opened 2024-08-30
Bump to SDL3 gitrev f6fc5e2

erysdren:
More API churn. SDL functions now return 0 for error and 1 for success, instead of -1 for error and 0 for success. Great.


Yamagi:
Thank you for doing this :)

PR #1150, opened 2024-09-07
Fix Windows key bind name on non-Mac

amalchuk:
Just like it was done in ioquake/ioq3@9fbbf42.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-.vscode/
 /build/
 /release/
 *.mk
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.vscode/
 /build/
 /release/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
+.vscode/
```
Yamagi, line 1: Hi :) This doesn't belong here. We don't add editor specific excludes to out .gitignore, because we cannot support all of the hundreds editors and other tools out there. Such excludes should go to the users global ~/.gitignore.

Can you remove it? Otherwise it's fine and I'm going to merge it afterwards.

amalchuk: Okay, got it.

Yamagi:
Thank you :)

PR #1151, opened 2024-09-08
Fix broken links in the documentation

amalchuk:


doc/020_installation.md
```diff
@@ -334,26 +334,26 @@ is copied to the corresponding directory in the Quake II installation.
 
 ### Compiling on Windows with Visual Studio (2015 and newer)
 
-To compile with Visual Studio, CMake is required.
```
Yamagi, line 337: The double whitespace at the end of a line forces a line break and should not be removed. This is one of the more questionable syntax elements of Markdown...

amalchuk: Most editors are trim whitespace in the documents. If a line break is so necessary, it's possible to replace it with:
* `<br>`
* `\`
* double new line?

Yamagi: Than the editor config is broken. Double whitespaces are ugly but they are the only way to get a line break in standard Markdown:

* `<br>` requires inline HTML to be enabled and is hard to read in the play text file. We ship the documentation as plain text for the Windows build and some Linux distros do the same.
* `\` isn't standard Markdown and compatibility isn't guaranteed, again it's hard to read in the plain text form.
* A double new line starts a new paragraph.

amalchuk: Okay, got it :octocat:

amalchuk: @Yamagi Let's go to merge?

Yamagi: Yes :)

Yamagi:
Thank you for your work and for incooperating my change requests!

PR #1152, opened 2024-09-20
Improvements for GL1/GLES1

protocultor:
### 1. Fixed stencil shadow in stereo modes

When shadows with stencil enabled (`r_shadows` & `gl1_stencilshadow`) were used with `gl1_stereo` between `3` and `5`, the renderer was doing a "_wallhack_", showing entities behind walls. Fix is just disabling stencil for shadow in these stereo modes, since they use the stencil buffer for their own operation.

### 2. Stereo 3D modes (gl1_stereo) allowed in GLES1

Availability of each mode still depend on the hardware, e.g. RPi3 doesn't do modes `3` to `5`, not even in _RaspiOS Bookworm_ using GL1. At least it's no longer locked by software, and now RPi3 does `2`, `6` and `7` in both GL1 and GLES1.
This marks the end of the usage of classic `glBegin()`/`glEnd()` pairing in this renderer.

### 3. gl1_discardfb functionality expanded

Now it can be used in GL1. Taking a hint from [ARM Mali performance tips](https://developer.arm.com/community/arm-community-blogs/b/mobile-graphics-and-gaming-blog/posts/mali-performance-2-how-to-correctly-handle-framebuffers), we call `glClear()` at the start of each frame, and at the end, we call `glDiscardFramebufferEXT()` with the same parameters. The last step is only available with GLES1, but it seems the first is the most important one, since in the RPi3  it helps in _Bookworm_ and _Arch_ using GL1.
When `gl1_discardfb` is `1`, we clear stencil and depth, and when `2`, we additionally clear color, which at least in a RPi3 is even more helpful with performance, getting almost a 10% improvement over the current master branch.
Note that the "_clear color_" is black, unless `r_clear` is set.

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -508,7 +508,8 @@ it's `+set busywait 0` (setting the `busywait` cvar) and `-portable`
   squares.
 
 * **gl1_stencilshadow**: If `gl_shadows` is set to `1`, this makes them
-  look a bit better (no flickering) by using the stencil buffer.
+  look a bit better (no flickering) by using the stencil buffer. Does
+  not work when `gl1_stereo` is `3`, `4` or `5`.
 
 * **gl1_lightmapcopies**: When enabled (`1`), keep 3 copies of the same
   lightmap rotating, shifting to another one when drawing a new frame.
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -517,9 +517,10 @@ it's `+set busywait 0` (setting the `busywait` cvar) and `-portable`
   (dynamic lighting) causes slowdown. By default in GL1 is disabled,
   while in GLES1 is enabled. Needs `gl1_multitexture 1` & `vid_restart`.
 
-* **gl1_discardfb**: Only available in ES1. If set to `1` (default),
-  send a hint to discard framebuffers after finishing a frame. Useful
-  for GPUs that attempt to reuse them, something Quake 2 doesn't do.
+* **gl1_discardfb**: If `1`, clear color, depth and stencil buffers at
+  the start of a frame, and discard them at the end if possible. If
+  `2`, do only depth and stencil, no color. Increases performance in
+  mobile / embedded. Default in GL1 is `0`, while in GLES1 is `1`.
 
 
 ## Graphics (OpenGL 3.2 and OpenGL ES3 only)
--- a/src/client/refresh/gl1/gl1_main.c
+++ b/src/client/refresh/gl1/gl1_main.c
@@ -150,6 +150,7 @@ void LM_FreeLightmapBuffers(void);
 void Scrap_Free(void);
 void Scrap_Init(void);
 
+extern void R_SetDefaultState(void);
 extern void R_ResetGLBuffer(void);
 
 void
@@ -630,6 +631,19 @@ R_PolyBlend(void)
 	glColor4f(1, 1, 1, 1);
 }
 
+static void
+R_ResetClearColor(void)
+{
+	if (gl1_discardfb->value == 1 && !r_clear->value)
+	{
+		glClearColor(0, 0, 0, 0.5);
+	}
+	else
+	{
+		glClearColor(1, 0, 0.5, 0.5);
+	}
+}
+
 void
 R_SetupFrame(void)
 {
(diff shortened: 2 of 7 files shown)


Yamagi:
I need to find my old red/blue 3D glasses... I'll have a look in the next few days.

protocultor:
Even with the issue where the main menu flickers sometimes, the more I use it, the more I'm convinced the default value for GLES1 should be "_clear the color_", so I swapped the values of `gl1_discardfb` for `1` and `2`: `1` clears everything, `2` is the "special" value that doesn't do color. I even though of getting rid of `2`, but supposedly there's hardware where clearing the color is counterproductive (nothing I own, at least).
@Pickle, can you check what does each value do in Pineview? Thanks.

Yamagi:
Now I'm feeling nauseous 🤢. Which means: Good work, it's working fine, just a little bit slow and thus triggering motion sickness. Which is, of course, caused by the slow SoC.

Yamagi:
One small nit: The old trick forcing a renderer restart by setting the `vid_fullscreen` cvar to `modified` doesn't work anymore. This needs to be changed to an explicit call of the `vid_restart` command.

Something like this should do do the trick:

```
--- src/client/refresh/gl1/gl1_main.c
+++ src/client/refresh/gl1/gl1_main.c
@@ -1811,9 +1811,7 @@ RI_BeginFrame(float camera_separation)
                else
                {
                        R_Printf(PRINT_ALL, "stereo supermode changed, restarting video!\n");
-                       cvar_t  *ref;
-                       ref = ri.Cvar_Get("vid_fullscreen", "0", CVAR_ARCHIVE);
-                       ref->modified = true;
+                       ri.Cmd_ExecuteText(EXEC_APPEND, "vid_restart\n");
                }
        }
```

With this the renderer will restart after `gl1_stereo` was changed and not only print the message every frame.

protocultor:
> Now I'm feeling nauseous 🤢. Which means: Good work, it's working fine, just a little bit slow and thus triggering motion sickness. Which is, of course, caused by the slow SoC.

Haha, sorry but I only said "stereo modes are supported", not that they were going to be fast. I was mostly worried about getting rid of `glBegin()`/`glEnd()`, since they are not supported in GLES1.
I believe we are close to the limits of the capabilities of a Pi 3, this time for real.
Something funny to try if you still have _Bullseye_ installed: unlike all the other OSs, "changing resolution" here doesn't change the actual, real display mode, but generates a "full screen texture" with the requested mode (or a renderbuffer object?), which allows the user to go crazy with the custom resolutions, without worrying about display support. So, if you try the following on a _16:9_ display:
```
r_customwidth 832
r_customheight 468
r_mode -1
```
...it will drop to 45 fps, but it'll be playable with `gl1_stereo 2`.
Something cool to try with `gl1_stereo 0`:
```
r_customwidth 1400
r_customheight 810
```
It just will get stuck at 60 fps.
If you want to know where did I get these numbers, the former are the proportions of 1280x720 multiplied by 0.65, and the latter, 1920x1080... by 0.75. Of course, none of these are real resolutions, but somehow Bullseye allows them, in both GL1 and GLES1.
(Note that using GLES1 in Bullseye is slightly faster, something that can be tested with a `timedemo 1` at 1920x1080. GL1 gets stuck at 30 fps, while GLES1 leans to 60 at times, using this PR's `gl1_discardfb 1`).
On the other hand, the highest resolution at constant 60 fps on _Bookworm_ and _Arch_ in my display is 1366x768... because it changes to the real resolution to display it. It also does 1400x900 at 60, but my monitor squashes the image, since it is not a 16:9 resolution.

> One small nit: The old trick forcing a renderer restart by setting the `vid_fullscreen` cvar to `modified` doesn't work anymore. This needs to be changed to an explicit call of the `vid_restart` command.
> 
> Something like this should do do the trick:
> ...
> With this the renderer will restart after `gl1_stereo` was changed and not only print the message every frame.

Crashes the game when changing `gl1_stereo` again and again in the middle of a level, but I prefer this to the console spam message. Thanks for the snippet.
I remember this to be a legacy/vanilla issue on how `vid_restart` works with a level loaded. It's not a problem if the user changes mode while "disconnected" in the console.

Yamagi:
> Something funny to try if you still have Bullseye installed: unlike all the other OSs, "changing resolution" here doesn't change the actual, real display mode, but generates a "full screen texture" with the requested mode (or a renderbuffer object?), which allows the user to go crazy with the custom resolutions, without worrying about display support.

I gave it a try. It's interesting, some kind of driver based up- / downscaling?

The gaming running makes it a bit better, but it's more me getting oldthan the 3D implementation. Even with 140 FPS on my desktops it's only a matter of minutes to trigger motion sickness.

> Crashes the game when changing gl1_stereo again and again in the middle of a level, but I prefer this to the console spam message. Thanks for the snippet.
> I remember this to be a legacy/vanilla issue on how vid_restart works with a level loaded. It's not a problem if the user changes mode while "disconnected" in the console.

Hmm... It's relative easy to trigger the crash on the RPI but not in my desktop with Mesa. Maybe some driver depended problem or driver <-> renderer interaction. Debugging is properly not worth the effort.

Merging this now. Keep up the good work :)

PR #1154, opened 2024-10-06
soft: fix texture lock

0lvin:
Regression has been  introduced in 39d2c7dbb45d53678e82baa1e727d2647d5ee755 as sdl3 has changed logic of SDL_LockTexture function call.


Yamagi:
Thank you :)

PR #1157, opened 2024-10-16
Clipboard and in-line editing for menu input fields

BjossiAlfreds:
This PR adds clipboard support and line editing extensions to menu fields (like the server address book).

CTRL actions:
* CTRL+V - Pastes clipboard into the current input field
* CTRL+C - Copies the input field content to the clipboard
* CTRL+X - Same as CTRL+C but clears the input field afterwards
* CTRL+L - Clears the current input field

Line navigation:
* Left/right arrows move the cursor left/right through the line, allowing for in-line editing
* Delete key deletes the character highlighted by the cursor
* Page End key jumps the cursor between the start and end of the current input string

And I added a new shared function, `Q_strisnum`, to test if a string is only made up of digits (for pasting input into number-only fields).

I also spotted an unsafe use of `strcpy`. When the value of the adr cvars are copied into the address book, there is no bounds checking, so cvar values longer than 79 bytes will corrupt static memory.

I don't know yet if this PR is fully ready. In my testing everything seems to work, both intended behavior and safeguards.

EDIT: Added code to reset the cursor pos to the end of the input when it loses focus (esc, up/down arrows).

### DIFF
--- a/src/client/cl_keyboard.c
+++ b/src/client/cl_keyboard.c
@@ -1420,7 +1420,8 @@ Key_Event(int key, qboolean down, qboolean special)
 	/* FIXME: Better way to do CTRL+<key> actions in the console?
 	   special should be set to true in this case.
 	*/
-	if (keydown[K_CTRL] && IsInConsole() &&
+	if (keydown[K_CTRL] &&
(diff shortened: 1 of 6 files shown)


Yamagi:
Thank you :)

PR #1158, opened 2024-10-26
Small fix for SDL3 API change for SetClipboardText

BjossiAlfreds:
Discovered that `SDL_SetClipboardText` API changed in SDL3 after the version I had been using, and it made the result of `IN_SetClipboardText` inconsistent between SDL2 and SDL3. Also added a comment to clearly define the return value of the wrapper.


Yamagi:
That SDL changed the return code for a lot of functions is really annoying and error prone... Thank you for fixing these! :)

PR #1163, opened 2024-11-16
Simplify compiling for targets that lack GL drivers

john-tornblom:
I also improved the logging of a couple of uncommon errors and warnings

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,6 +65,11 @@ option(CURL_SUPPORT "cURL support" ON)
 option(OPENAL_SUPPORT "OpenAL support" ON)
 option(SYSTEMWIDE_SUPPORT "Enable systemwide installation of game assets" OFF)
 option(SDL3_SUPPORT "Build against SDL 3 instead of SDL2" OFF)
+option(GL1_RENDERER "Build the GL3 renderer" ON)
+option(GL3_RENDERER "Build the GL3 renderer" ON)
+option(GLES1_RENDERER "Build the GLES1 renderer" OFF)
+option(GLES3_RENDERER "Build the GLES3 renderer" ON)
+option(SOFT_RENDERER "Build the software renderer" ON)
 
 set(SYSTEMDIR "" CACHE STRING "Override the system default directory")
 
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -189,11 +194,13 @@ else()
 	list(APPEND yquake2SDLLinkerFlags ${SDL2_LIBRARY})
 endif()
 
-# We need an OpenGL implementation.
-set(OpenGL_GL_PREFERENCE GLVND)
-find_package(OpenGL REQUIRED)
-list(APPEND yquake2IncludeDirectories ${OPENGL_INCLUDE_DIR})
-list(APPEND yquake2OpenGLLinkerFlags ${OPENGL_LIBRARIES})
+if(GL1_RENDERER OR GL3_RENDERER OR GLES1_RENDERER OR GLES2_RENDERER)
```
protocultor, line 197: `GLES2_RENDERER` does not exist, so I assume it should be actually `GLES3_RENDERER`. But still have to mention that the only renderer that requires the OS's OpenGL implementation is GL1; the other three use loaders made with _glad_. Take a look to the `glad*` directories inside `refresh`, they include everything needed to load each OpenGL "version".
So, I believe this could be replaced with `if(GL1_RENDERER)`.

john-tornblom:
Ah, my apologies, the GLES2_RENDERER was a typo. And you are correct, all of the default renderers can be compiled without linking to the OpenGL lib provided by the OS. I hope I addressed this in [9700eb6](https://github.com/yquake2/yquake2/pull/1163/commits/9700eb6baf84d688ad919d45e41812164834e564)

src/backends/generic/misc.c
```diff
@@ -156,8 +156,9 @@ const char *Sys_GetBinaryDir(void)
 	SetExecutablePath(exeDir);
 
 	if (exeDir[0] == '\0') {
-		Com_Printf("Couldn't determine executable path. Using ./ instead.\n");
-		Q_strlcpy(exeDir, "./", sizeof(exeDir));
+		getcwd(exeDir, sizeof(exeDir));
+		strcat(exeDir, "/");
+		Com_Printf("Couldn't determine executable path. Using %s instead.\n", exeDir);
```
Yamagi, line 161: This breaks the build under MacOS and Windows. I think that unistd.h is available under MacOS, so it's properly enough to change the `#if defined` in line 32. For Windows it's more tricky. We can't use the normal `_getcwd()` because it doesn't support wide chars in the returned path, breaking when there are non-ascii characters in it. German umlauts, cyrillic, japanese, chinese, etc. So the way to go is to use `_wgetcwd()` followed by `WideCharToMultiByte(). There is an example in line 67 showing how to use `WideCharToMultiByte()`.

john-tornblom:
I believe the testbuild workflows for windows and macos will pass now

Yamagi:
Thank you for the fast fix :) (After I have taken over a week...)

john-tornblom:
No problem at, thank you for the merge

Issue #1135, opened 2024-08-08
Several problems with `s_sndrestart` and the SDL sound backend


Yamagi:
#1132 brought some problems with the sound system to light:

* It's easy to crash the game, to end up without sound or even endless loops during shutdown when doing a `snd_restart`. Especially when changing backends or after having disconnected from a (local) server.
* For the `wsaapi` SDL sound driver YQ2 must not insist on AUDIO_S16SYS or AUDIO_U8, the backend only guarantees AUDIO_F32LSB. Instead have SDL recode the audio by setting `obtained` passed to `SDL_OpenAudio()` (or the SDL3 equivalent) to NULL: https://wiki.libsdl.org/SDL2/SDL_OpenAudio

Yamagi:
Before I committed the transparent format conversions, ASAN had a chance of about 25% to trigger somewhere in SDL when doing a `disconnect ; s_openal 1 ; snd_restart`. It happened only when switching from SDL to OpenAL and only if some amount of sound sampled were loaded.

After I committed the change it doesn't trigger anymore. I find this strange, because the change didn't touch the SDL sound shutdown logic and the selected driver is *pulseaudio* in both cases.

Is anybody still able to crash the game with `snd_restart`? If yes, what is the exact sequence to trigger it? Up to date Windows binaries can be found here: https://github.com/yquake2/yquake2/actions/runs/10758446004/artifacts/1905710869 (You need to be logged in to download them).

Yamagi:
> yquake2.exe +disconnect

That crashes the game also for me. Looks like the game tries to recover the state of the OGG backend (title and playback position) without having it saved first. I'll come up with a patch.

> when did the default of s_openal changed to 0?

It didn't, it's still 1: https://github.com/yquake2/yquake2/blob/master/src/client/sound/sound.c#L1635

Yamagi:
I have pushed a fix. `OGG_SaveState()` and `OGG_RecoverState()` now abort early if `ogg_enable` is not set to `1`, fixing this bug and maybe some variants of it.

Yamagi:
You could try HRTF: https://github.com/kcat/openal-soft/blob/master/alsoftrc.sample#L103

You need to install a HRTF profile, though. The standard profile is under https://github.com/kcat/openal-soft/blob/master/hrtf/Default%20HRTF.mhr and can be referenced with the `hrtf-paths` config option.

[...]

PR #1170, opened 2024-12-04
Sound backend selector + weapon preview for cycleweap

protocultor:
As discussed in https://github.com/yquake2/yquake2/pull/1128#issuecomment-2365045562, I took the liberty to replace the Sound Quality selector with a Sound Backend selector.

Another good reason to make this is that the current _Quality_ option... doesn't work. Try to lower the quality to _normal_, then get out of the menu, then get back to it. The reason it "gets back" to _high_ is because [the validation for quality compares s_loadas8bit](https://github.com/yquake2/yquake2/blob/master/src/client/menu/menu.c#L2257), which [never changes from 0, even if s_khz does](https://github.com/yquake2/yquake2/blob/master/src/client/menu/menu.c#L2346).

In my tests I haven't been able to break the backend. As discussed in #1135, I assume that in those systems that do, a game reboot solves it. Funnily, I'm more worried about the text used in the menu; any feedback is appreciated.

In addition to this, and after seeing how well it worked in #1076, the "pickup style" preview for the upcoming weapon of `weapprev`/`weapnext` was added to `cycleweap`. Including PRs for the mission packs.
Note that the preview will only appear when `cycleweap` is called with 3 or more parameters, to maintain its original intent of being a custom "Xatrix-style" weapon selector (#472).

And finally, changed the weapon binds for gamepads/controllers in the menu to include the mission packs' weapons.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -286,6 +286,10 @@ else
 SDLCFLAGS := $(shell sdl2-config --cflags)
 endif
 
+ifdef NO_SDL_GYRO
+SDLCFLAGS += -DNO_SDL_GYRO
+endif
+
 # ----------
 
 # Base include path.
--- a/doc/030_configuration.md
+++ b/doc/030_configuration.md
@@ -39,18 +39,17 @@ Yamagi Quake II ships with 4 renderers:
 
 ## Choosing a Sound System
 
-Yamagi Quake II ships with 2 sound system:
+Yamagi Quake II ships with 2 sound systems:
 
 * The **OpenAL** sound system: This is the default and highly
-  recommended. It provides full surround sound support and even HRTF for
+  recommended. It provides full surround sound support and HRTF for
   headphones. But also the plain stereo playback is much better than in
   the original sound system. The setup is done mostly through OpenAL,
   have a look at the documentation of your OpenAL library.
 * The **SDL** sound system: This is the classic sound system, providing
-  an experience like the original client. Set `s_openal` to `0` and
-  execute an `snd_restart` to activate it. The classic sound system may
-  be somewhat problematic on modern systems like Windows 10 or Linux
-  with Pulseaudio.
+  an experience like the original client. It's less CPU demanding than
+  OpenAL. Choose it in the options menu, or set `s_openal` to `0` and
+  execute an `snd_restart`, to activate it.
 
 
 ## Tuning for Precise Timings
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -511,6 +511,10 @@ it's `+set busywait 0` (setting the `busywait` cvar) and `-portable`
   look a bit better (no flickering) by using the stencil buffer. Does
   not work when `gl1_stereo` is `3`, `4` or `5`.
 
+* **gl1_waterwarp**: Intensity of the "squeeze/stretch" effect on the
+  FOV when diving underwater. Can be any floating point number, `0`
+  disables it (Vanilla Quake II look). Default `1.0`.
+
 * **gl1_lightmapcopies**: When enabled (`1`), keep 3 copies of the same
   lightmap rotating, shifting to another one when drawing a new frame.
   Meant for mobile/embedded devices, where changing textures just shown
--- a/src/client/input/sdl2.c
+++ b/src/client/input/sdl2.c
@@ -173,19 +173,20 @@ static cvar_t *gyro_calibration_x;
 static cvar_t *gyro_calibration_y;
 static cvar_t *gyro_calibration_z;
 
-#if SDL_VERSION_ATLEAST(2, 0, 14)	// support for controller sensors (gyro, accelerometer)
+// If the used SDL version doesn't support gamepad sensors...
+#if !SDL_VERSION_ATLEAST(2, 0, 14)
+// ...disable support for reading them.
+#define NO_SDL_GYRO
+#endif
 
+#ifndef NO_SDL_GYRO	// use SDL_CONTROLLERSENSORUPDATE to read gyro
 static unsigned int num_samples;
-#define NATIVE_SDL_GYRO	// uses SDL_CONTROLLERSENSORUPDATE to read gyro
-
-#else	// for SDL < 2.0.14, gyro can be read as a "secondary joystick" exposed by dkms-hid-nintendo
-
+#else	// gyro can be read as a "secondary joystick" exposed by dkms-hid-nintendo
 static unsigned int num_samples[3];
 static SDL_Joystick *imu_joystick = NULL;	// gyro "joystick"
 #define IMU_JOY_AXIS_GYRO_ROLL 3
 #define IMU_JOY_AXIS_GYRO_PITCH 4
 #define IMU_JOY_AXIS_GYRO_YAW 5
-
 #endif
 
 // To ignore SDL_JOYDEVICEADDED at game init. Allows for hot plugging of game controller afterwards.
@@ -789,7 +790,7 @@ IN_Update(void)
 				break;
 			}
 
-#ifdef NATIVE_SDL_GYRO	// controller sensors' reading supported (gyro, accelerometer)
+#ifndef NO_SDL_GYRO	// gamepad sensors' reading is supported (gyro, accelerometer)
 			case SDL_CONTROLLERSENSORUPDATE:
 				if (event.csensor.sensor != SDL_SENSOR_GYRO)
 				{
@@ -804,7 +805,7 @@ IN_Update(void)
 					break;
 				}
 
-#else	// gyro read as "secondary joystick"
+#else	// gyro read from a "secondary joystick" (usually with name ending in "IMU")
 			case SDL_JOYAXISMOTION:
 				if ( !imu_joystick || event.cdevice.which != SDL_JoystickInstanceID(imu_joystick) )
 				{
@@ -831,12 +832,12 @@ IN_Update(void)
 					break;
 				}
 
-#endif	// NATIVE_SDL_GYRO
+#endif	// !NO_SDL_GYRO
 
 				if (gyro_active && gyro_mode->value &&
 					!cl_paused->value && cls.key_dest == key_game)
 				{
-#ifdef NATIVE_SDL_GYRO
+#ifndef NO_SDL_GYRO
 					if (!gyro_turning_axis->value)
 					{
 						gyro_yaw = event.csensor.data[1] - gyro_calibration_y->value;		// yaw
@@ -864,7 +865,7 @@ IN_Update(void)
 								gyro_yaw = axis_value - gyro_calibration_z->value;
 							}
 					}
-#endif	// NATIVE_SDL_GYRO
+#endif	// !NO_SDL_GYRO
 				}
 				else
 				{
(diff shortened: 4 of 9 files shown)


protocultor:
Just because it's simple, just added an extra: "_water warp_" effect for GL1. Actually a "squeeze/stretch" to the FOV when underwater, it's _very_ fast and looks decent enough for hardware without GL3 support. "Borrowed" from Fitzquake, so credits to [John Fitzgibbons](https://www.celephais.net/fitzquake/).

A simple improvement to the original is that the new cvar for it, `gl1_waterwarp`, accepts floating point values, to control the intensity of the effect. Low values like `0.35` look more natural to me.

protocultor:
Since no one has stopped me, new extras:

- Proper detection of gyro gamepads with SDL3. Of course, a funny function return value was the reason it wasn't working (it was actually funny in SDL2, but consistency is something SDL is just trying to fix :) ).

- A new Makefile parameter, `NO_SDL_GYRO`, to disable Gyro support with SDL. An extension of #990, this allows to force the "Switch controller gyro" for SDL versions prior to 2.0.14... in versions later than that.
The reason? Raspberry Pi is not supporting sensor reading for gamepads in any of the newer SDL versions, under any of the available OSs. I assume they aren't building SDL with HIDAPI support for the Pis, but the real reason I don't actually know. In any case, and at least in _Arch Linux ARM_, Nintendo gamepads show a second "IMU joystick", just like when using `dkms-hid-nintendo` in _Buster_ and _Bullseye_, proving that the IMU works (which makes more baffling that SDL is not reading it). With this commit, at least we force to read the "IMU joystick", even under a potential SDL 3 version. I'd rather have a fully working SDL library, but until then, this is better than not having gyro...

Yamagi:
Two weeks on the return and such a nice pull request as christmas present at my return home :)

> As discussed in https://github.com/yquake2/yquake2/pull/1128#issuecomment-2365045562, I took the liberty to replace the Sound Quality selector with a Sound Backend selector.

Thank you for doing it. It was an my list by Quake II isn't really high priority right now.


> Another good reason to make this is that the current Quality option... doesn't work. Try to lower the quality to normal, then get out of the menu, then get back to it. The reason it "gets back" to high is because [the validation for quality compares s_loadas8bit](https://github.com/yquake2/yquake2/blob/master/src/client/menu/menu.c#L2257), which [never changes from 0, even if s_khz does](https://github.com/yquake2/yquake2/blob/master/src/client/menu/menu.c#L2346).

It was broken nearly 15 years ago by me in [66ac4643cbfec4697d829bb9ba8c4cf97e6d4232](https://github.com/yquake2/yquake2/commit/66ac4643cbfec4697d829bb9ba8c4cf97e6d4232#diff-dde31035935b758ab8ba49a687e7b60998b95732bd6a74a0f98c7595d040fb5aL1147) And no one ever noticed or reported it. I doubt that there was ever really a need for lowering the sound quality, maybe besides very low end machines in the late 1990th. The sound mixing and spartializing isn't that CPU demanding, even on my Pentium 166 - which was well below the recommended Pentium 200 - the game ran fast enough with high sound quality and soft rendering to 680x480.

> In my tests I haven't been able to break the backend. As discussed in https://github.com/yquake2/yquake2/issues/1135, I assume that in those systems that do, a game reboot solves it. Funnily, I'm more worried about the text used in the menu; any feedback is appreciated.

The opening bracing look wrong but that's the broken monospace font. The brace should be right aligned, it's left aligned instead. Nothing we can do about it. The question is if 'quality' and 'performance' are the right descriptions. There are players who argue that OpenAL sounds wrong. An example are the sirens in city2(?). I don't have a better idea, though. And quality vs. performance isn't wrong, the SDL is faster.

> weapprev/weapnext was added to cycleweap. Including PRs for the mission packs.
Note that the preview will only appear when cycleweap is called with 3 or more parameters, to maintain its original intent of being a custom "Xatrix-style" weapon selector (https://github.com/yquake2/yquake2/pull/472).

> And finally, changed the weapon binds for gamepads/controllers in the menu to include the mission packs' weapons.

Cool, thank you :)

> Just because it's simple, just added an extra: "water warp" effect for GL1. Actually a "squeeze/stretch" to the FOV when underwater, it's very fast and looks decent enough for hardware without GL3 support. "Borrowed" from Fitzquake, so credits to [John Fitzgibbons](https://www.celephais.net/fitzquake/).

Looks good, way better than no water effect. Not to say really good for something which is implemented without shaders. It has no measureable performance impact on my laptop. And if it has it can be switched off.

> Proper detection of gyro gamepads with SDL3. Of course, a funny function return value was the reason it wasn't working (it was actually funny in SDL2, but consistency is something SDL is just trying to fix :) ).

And it's annoying. @DanielGibson and me already fixed several changed return codes. I hope we finally found all of them...

> The reason? Raspberry Pi is not supporting sensor reading for gamepads in any of the newer SDL versions, under any of the available OSs. I assume they aren't building SDL with HIDAPI support for the Pis, but the real reason I don't actually know.

That's properly the reason. There should any difference between the platforms in SDL, of course. That's still one of the bigger problems with all these ARM tinker boards out there. The distro support is still flaky and suffers from strange decisions. With RaspiOS being one of the better, at least there aren't tied to one specific kernel anymore and they ship all the required tools to build an own version of SDL and most other libraries.


Can I merge this or do you plan another round of additions?

protocultor:
> Two weeks on the return and such a nice pull request as christmas present at my return home :)

Ho ho ho, Merry Christmas :)

> I doubt that there was ever really a need for lowering the sound quality, maybe besides very low end machines in the late 1990th. The sound mixing and spartializing isn't that CPU demanding, even on my Pentium 166 - which was well below the recommended Pentium 200 - the game ran fast enough with high sound quality and soft rendering to 680x480.

My experience is the same, with the same processor (with MMX!). I did use the low quality sound for a while, but since it reduces the quality so much (_11 khz_), and didn't make any visible improvement in performance, I ended up reverting it to "high" (_22 khz_). I ran the game at 512x384 :)
The backend selector may be used only once in an installation, but at least provides a more useful option, something with a more notorious effect.

> The opening bracing look wrong but that's the broken monospace font. The brace should be right aligned, it's left aligned instead. Nothing we can do about it. The question is if 'quality' and 'performance' are the right descriptions. There are players who argue that OpenAL sounds wrong. An example are the sirens in city2(?). I don't have a better idea, though. And quality vs. performance isn't wrong, the SDL is faster.

Yeah, this is what I was afraid of. Do we reduce options to just "**openal**" and "**sdl**", no further description?

> > Of course, a funny function return value was the reason it wasn't working
> 
> And it's annoying.

Agreed.

> > I assume they aren't building SDL with HIDAPI support for the Pis, but the real reason I don't actually know.
> 
> That's properly the reason. There should any difference between the platforms in SDL, of course. That's still one of the bigger problems with all these ARM tinker boards out there. The distro support is still flaky and suffers from strange decisions.

Note that I just learned to build SDL2 on Arch, and in the "_configure_" step, both SDL_HIDAPI and SDL_HIDAPI_JOYSTICK appear as `(Wanted: ON): ON`, so at least doesn't look like it's being disabled willingly by the source or the people in charge of the packages. SDL_HIDAPI_LIBUSB is OFF, but supposedly it shouldn't affect. I'll check the SDL code in the next few weeks and see if there's something that can be done.

> Can I merge this or do you plan another round of additions?

The "extras" were supposed to be future PRs, but the xmas miracle happened and they ended up being small and clear enough to be put here.
I don't believe the next round of additions will be like that. My only doubt is what do we do with the sound backend selector text. Apart from that, this is ready to be merged.

Yamagi:
> My only doubt is what do we do with the sound backend selector text.

What about leaving it at 'openal' and 'sdl' without further hints, like most (all?) other options and update the description in the [configuration guide](https://github.com/yquake2/yquake2/blob/master/doc/030_configuration.md#choosing-a-sound-system) a little bit.

Something like:

```diff
diff --git a/doc/030_configuration.md b/doc/030_configuration.md
index 0e49bf1c..124df7c0 100644
--- a/doc/030_configuration.md
+++ b/doc/030_configuration.md
@@ -42,15 +42,14 @@ Yamagi Quake II ships with 4 renderers:
 Yamagi Quake II ships with 2 sound system:

 * The **OpenAL** sound system: This is the default and highly
-  recommended. It provides full surround sound support and even HRTF for
+  recommended. It provides full surround sound support and HRTF for
   headphones. But also the plain stereo playback is much better than in
   the original sound system. The setup is done mostly through OpenAL,
   have a look at the documentation of your OpenAL library.
 * The **SDL** sound system: This is the classic sound system, providing
-  an experience like the original client. Set `s_openal` to `0` and
-  execute an `snd_restart` to activate it. The classic sound system may
-  be somewhat problematic on modern systems like Windows 10 or Linux
-  with Pulseaudio.
+  an experience like the original client. It's less CPU demanding than
+  OpenAL. Choose it in the options menu or set `s_openal` to `0` and
+  execute an `snd_restart` to activate it.


 ## Tuning for Precise Timings
```

protocultor:
People don't RTFM but it does reflect better the current state of affairs.
Thanks for the feedback. I believe the PR is ready, it's your call now.

Yamagi:
> People don't RTFM but it does reflect better the current state of affairs.

Docs don't exists to be read. They exists that we can tell people that they should have read the docs ;)

Many thanks for your continued efforts!

protocultor:
> > That's properly the reason. There should any difference between the platforms in SDL, of course. That's still one of the bigger problems with all these ARM tinker boards out there. The distro support is still flaky and suffers from strange decisions.
> 
> Note that I just learned to build SDL2 on Arch, and in the "_configure_" step, both SDL_HIDAPI and SDL_HIDAPI_JOYSTICK appear as `(Wanted: ON): ON`, so at least doesn't look like it's being disabled willingly by the source or the people in charge of the packages. ... I'll check the SDL code in the next few weeks and see if there's something that can be done.

Ok, now I know why it didn't work. To detect joysticks/gamepads, SDL does the following:
1. First search and enumerate HIDAPI devices.
2. If no devices are found, search using the native method of the OS.

What happens in the RPi is that the detected joystick is found in the second step, by `SDL_LINUX_JoystickDriver`, which is more limited in features than the first driver, `SDL_HIDAPI_DeviceDriver`, even if the joystick in question was found through `udev`.
To solve this, the OS must provide _Hidraw_ read/write access to the joystick. In desktop Arch, there's this package:
https://aur.archlinux.org/packages/game-devices-udev
...but even if ARM doesn't have it, you can manually install the original set of rules the package above is based on:
https://codeberg.org/fabiscafe/game-devices-udev
Just follow the instructions in "_Others_" (OSs).

After having installed them, `SDL_HIDAPI_DeviceDriver` does find the joystick and all the available sensors it has. Tested, and fully working, in _Arch Linux ARM_ and _Raspberry Pi OS Bookworm_. Didn't work on _Bullseye_, but that's probably because SDL 2.0.14 doesn't have decent sensor support (it was the first SDL version with it).
I will still defend my work on `NO_SDL_GYRO` here; it still works for people who doesn't want to set too may "rules" in their systems.

Yamagi:
> I will still defend my work on NO_SDL_GYRO here; it still works for people who don't want to set too many "rules" in their systems.

And it will run out of the box. That the udev rules are missing is so obscure that even most experienced users won't know it. And won't find it through Google either.

protocultor:
> That the udev rules are missing is so obscure that even most experienced users won't know it. And won't find it through Google either.

You're right. Just threw a couple of bones to search engines, I hope they catch them:
https://forums.raspberrypi.com/viewtopic.php?t=382036
https://www.reddit.com/r/GyroGaming/comments/1hufckv/if_youre_using_linux_and_gyro_doesnt_seem_to_work/

PR #1174, opened 2024-12-21
disable certificate validation for HTTPS

SiemensSchuckert:
Disable SSL certificate check - to allow download from servers with self-signed cert, or when some certs are missing from system certificate store that CURL uses.

the assumption is: transport security is not necessary for game resources download, so Quake better be compatible with more server configurations


0lvin:
Could it be cvar settings?

SiemensSchuckert:
now with new cvar `cl_http_verifypeer`

Yamagi:
Thank you for adding the cvar and the documentation for it. I would have also requested it. While I agree that the certificates validity doesn't really mater given the general state of Quake II and security, some users still may want to verify it.

PR #1175, opened 2024-12-21
Make more function local and use const with names

0lvin:
* applied recommendation from msvc for use size_t with stolen. 
* make local function static
* Image name in functions const


Yamagi:
Thank you :) Doing the same for the addons would be nice if you have the time.

PR #1176, opened 2024-12-28
GL3_LoadPic fix UB.

devnexen:
Mainly because gl3image_s and image_s are not the same size, also fixing its signature while at it.

### DIFF
--- a/src/client/refresh/gl3/gl3_image.c
+++ b/src/client/refresh/gl3/gl3_image.c
@@ -374,8 +374,8 @@ FloodFillSkin(byte *skin, int skinwidth, int skinheight)
 /*
  * This is also used as an entry point for the generated r_notexture
  */
-gl3image_t *
-GL3_LoadPic(char *name, byte *pic, int width, int realwidth,
+struct image_s *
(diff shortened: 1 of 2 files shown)


0lvin:
Same size is not required, client logic has never use/know what kind of pointer is returned from render library. And all renders has different sizes of image_t structure. Client must never try to check or interpret pointer to image in any way. Have you found places where client tried to with image_s internals?

devnexen:
> Same size is not required

Oh ok, did not know ; then it s pointless to bother.

> Have you found places where client tried to work with image_s internals?

Nope.

0lvin:
Currently in most of cases, `struct image_s` is alias to something like `const void *`.

If we speak about some more clear solution as refactoring, we could replace it to some integer value, in such case we could reallocate image list when we need such, instead keeping statically allocated 1024 or so images array. In same time, I am not sure that such refactoring worth, as most of maps are created with such limitation, and such change will only make code come sophisticated with dynamic allocation and will help with caching images between maps reload. At least it is true for classic yquake2 code, it could provide some benefits for yquake2remaster with extended count of models and size of maps.

PR #1178, opened 2025-01-22
Fix compialtion for SDL 3.2.0

erysdren:
Seems to be the only thing required on a cursory glance.


protocultor:
Interestingly, all these lines can be deleted.
The original reason for setting these hints come from #857, where they were needed for gyro while using Bluetooth in PS4 and PS5 controllers. In SDL3, both hints became one, and it is now enabled by default. So, there's no longer a need to set it.
Still, this change works, and maybe it's better to have it enabled in a clear, readable way.

Yamagi:
Thank you both. I have removed the lines.

Issue #1180, opened 2025-02-01
stb: Several memory access violations in stb_image and stb_vorbis


mnhauke:
A vulnerable copy of the stb code in seems to be present in
https://github.com/yquake2/yquake2/blob/master/src/client/refresh/files/stb_image.h
https://github.com/yquake2/yquake2/blob/master/src/client/sound/header/stb_vorbis.h

GHSL-2023-145_GHSL-2023-151/GHSL-2023-165_GHSL-2023-172: Several memory access violations in stb_image and stb_vorbis

GHSL-2023-145 - CVE-2023-45661
GHSL-2023-146 - CVE-2023-45662
GHSL-2023-147 - CVE-2023-45663
GHSL-2023-148 - CVE-2023-45664
GHSL-2023-150 - CVE-2023-45666
GHSL-2023-151 - CVE-2023-45667
GHSL-2023-165 - CVE-2023-45675
GHSL-2023-166 - CVE-2023-45676
GHSL-2023-167 - CVE-2023-45677
GHSL-2023-168 - CVE-2023-45678
GHSL-2023-169 - CVE-2023-45679
GHSL-2023-170 - CVE-2023-45680
GHSL-2023-171 - CVE-2023-45681
GHSL-2023-172 - CVE-2023-45682

PRs (not merged yet):
https://github.com/nothings/stb/pull/1539
https://github.com/nothings/stb/pull/1541
https://github.com/nothings/stb/pull/1543
https://github.com/nothings/stb/pull/1545
https://github.com/nothings/stb/pull/1547
https://github.com/nothings/stb/pull/1549
https://github.com/nothings/stb/pull/1551
https://github.com/nothings/stb/pull/1553
https://github.com/nothings/stb/pull/1554
https://github.com/nothings/stb/pull/1555
https://github.com/nothings/stb/pull/1556
https://github.com/nothings/stb/pull/1557
https://github.com/nothings/stb/pull/1558
https://github.com/nothings/stb/pull/1559
https://github.com/nothings/stb/pull/1560

References:
https://securitylab.github.com/advisories/GHSL-2023-145_GHSL-2023-151_stb_image_h/

openSUSE bugtracker:
https://bugzilla.suse.com/show_bug.cgi?id=1236654

[...]

PR #1181, opened 2025-02-02
update stb_* codes to latest master

0lvin:
Partial fix of #1180


Yamagi:
You're fast :) Let's wait with merging this until it's clear if should switch the upstream repo to https://github.com/NBickford-NV/stb

DanielGibson:
I wouldn't rely on the stb master having all required fixes, it only rarely gets updates and the repo has >130 open PRs.
Using Nia Bickford's branch is probably a better idea.

0lvin:
What do you think about apply each CVE related pull request as separate patch, and when nothing's/STB will be updated try to rebase patches?

DanielGibson:
> What do you think about apply each CVE related pull request as separate patch, and when nothing's/STB will be updated try to rebase patches?

Oh god no.
That would be a lot of work (also remember that merging multiple pull requests can create new issues when they don't interact well).
Just use the file as it is from a trusted upstream, either stb directly (I guess his branch should at least work, even if it might have vulnerabilities that aren't very relevant for us) or Nia Bickford

DanielGibson:
To elaborate a bit more, it looks like Nia did and does a lot of work on this, for example by fuzzing the libraries and adding lots of tests. I'm sure she also adds new upstream changes from Sean Barrett, if any.

If you really want to spend more time on this than every once in a while copying Nia's latest state, I'd suggest collaborating with her (write more tests, create PRs with code from PRs against stb's repo if she hasn't merged them already, etc) instead of duplicating her work. Collaboration is one of the things Open Source is for :)

PR #1183, opened 2025-02-02
Consistent binding between multiple gamepad types + labels by style

protocultor:
### Long but needed explanation of the problem...

One of the "features" of old SDL2's `SDL_GameController` is that the face buttons were treated as **A/B/X/Y**, and their position didn't matter for the most part, since Nintendo gamepads exchange positions between A and B, and they also do it with X and Y.
```
     (Y)              (X)
  (X)   (B)        (Y)   (A)
     (A)              (B)

  Standard         Nintendo
```
This was kinda cool on a usability perspective, since e. g. you could always use 'A' to confirm and 'B' to cancel in menus, and you didn't have to worry about their positions. The problem: button assignments were not actually "cross-gamepad" since this inconsistency between positions will get actions really meant for the "bottom" button in the "right" button, and so on.

As many things, this didn't age well, and current SDL3's `SDL_Gamepad` threw away the button labels and replaced them with enforced-position buttons, with the button names being referenced as cardinal directions.
```
      (North)
  (West)   (East)
      (South)

 SDL3 nomenclature
```
At least in terms of functionality, users of "Standard" button labels were mostly unaffected by this change. Labeling of buttons are now [responsability of the developer](https://github.com/libsdl-org/SDL/blob/main/docs/README-migration.md#sdl_gamecontrollerh), but even if you don't do it, and keep the SDL2 button labeling / indexing, "standard" users won't miss much (as seen in YQ2 itself).
Nintendo users are another thing; they really have their buttons swapped for good, both label and functionality.

### So, what is this for?

Since SDL3 enforces position of buttons, this PR makes use of it and names them NORTH, SOUTH, EAST and WEST; thanks to this, we can truly have bindings that will work across multiple gamepads, no matter the vendor. Note that by manually disabling `SDL_HINT_GAMECONTROLLER_USE_BUTTON_LABELS` we have the same behavior in SDL2: force positions, not labels, so we end up with both SDL2 and SDL3 working the same.
In addition to that:
- Button labels from _Xbox_, _Playstation_ and _Switch_ gamepads can be used in the "customize buttons" menu, just for the cool factor. Style is autodetected, or can be manually selected.
- Users can select which button confirms and which cancels in menus. You can use [Playstation japanese style](https://www.denofgeek.com/games/playstation-5-x-button-confirm-cancel-japan/) if you want, with status bar instructions in "customize buttons" reflecting this. By default is also in "autodetect", applying japanese style if using a Nintendo gamepad.
- The left stick can be used to navigate menus.
- Now GLES1 compiles without warnings when using SDL 3.2.x.

Now that button mapping is truly cross-platform, cross-hardware, cross-anything... wouldn't be useful to have default binds set in `yq2.cfg`?

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -581,16 +581,16 @@ it's `+set busywait 0` (setting the `busywait` cvar) and `-portable`
 * **sw_colorlight**: enable experimental color lighting.
 
 
-## Game Controller
+## Gamepad
 
 * **in_initjoy**: Toggles initialization of game controller. Default is
   `1`, which enables gamepad usage; `0` disables its detection at
   startup. Can only be set from command line.
 
-* **in_sdlbackbutton**: Defines which button is used in the gamepad as
+* **joy_escbutton**: Defines which button is used in the gamepad as
   the `Esc` key, to pull the main menu and 'cancel' / 'go back' on its
-  options. Valid values are `0` = Back / Select / Minus, `1` = Start /
-  Menu / Plus (default), or `2` = Guide / Home / PS. Requires a game
+  options. Valid values are `0` = Start / Menu / Plus (default), `1` =
+  Back / Select / Minus, or `2` = Guide / Home / PS. Requires a game
   restart, or gamepad replug, when changed.
 
 * **joy_layout**: Allows to select the stick layout of the gamepad.
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -602,6 +602,12 @@ it's `+set busywait 0` (setting the `busywait` cvar) and `-portable`
   - `2`: *Playstation*, 4 & 5 format
   - `3`: *Switch*, traditional Nintendo button format
 
+* **joy_confirm**: Style of *confirm* and *cancel* buttons in menus. As
+  with the previous one, SDL 2.0.12 is required for `-1` to work.
+  - `-1`: *Autodetect*, sets to `1` if Nintendo, `0` otherwise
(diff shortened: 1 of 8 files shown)


DanielGibson:
In dhewm3 I also use the positional stuff internally (and only use the name of the button for displaying it).

But TBH that's not perfect either.

Because people using a Nintendo gamepad are used to A being confirm, even though it's in the EAST position on that gamepad.

The whole situation just sucks.

I suggest building a timemachine and beating beating the responsible people at Sega (that's where Microsoft's layout originates) with a club until they agree to use Nintendos button layout.

protocultor:
> In dhewm3 I also use the positional stuff internally (and only use the name of the button for displaying it).
> But TBH that's not perfect either.
> Because people using a Nintendo gamepad are used to A being confirm, even though it's in the EAST position on that gamepad.

Personal opinion: that's the reason why I preferred the SDL2 "labeling" system.
Binding was not cross-gamepad, but that's why you have a button assign menu. On the other hand, you could always leave specific operations like menu confirmation to the "A" button, and no matter your gamepad, you will always get it right since, as you said, "A" is SOUTH in Xbox and PS, and EAST in Nintendo.
Welp, priorities are now different I guess.

About dhewm3, it's cool that you were ready for the SDL3 changes months ago. Maybe the 4th commit of this PR can inspire you to find the solution for the button swap in Nintendo gamepads.

protocultor:
New extra: fixed wrong behavior on the "gyro mode" selector. Now it updates correctly when changed, and "always on" mode works properly.
Thanks @AL2009man for the report.

Yamagi:
Coo, good work. Thank you :)

PR #1184, opened 2025-02-08
sdl3: Fix memory leak in IN_Controller_Init

0lvin:
Detected by ASAN=1
Build Win64-SDL3 as part of github actions


0lvin:
In release notes of sdl2-2.32 is recommended to use sdl3 instead, looks like it's time to start test builds of sdl3 also.

Broken Win64 should be not so critical, so sdl3 is tested with it.

Yamagi:
The first distros like Arch and it's derivatives already removed SDL2 in favor of sdl2_compat. Using Win64 as test target for the time being is a good idea. Thank you for implementing this.

Issue #1185, opened 2025-02-23
Defensive Measures (defens) - lift-door buttons not working?


inf78:
I was just in a mood for Defensive Measures again after like 20 years or so, but sadly it doesn't seem to work properly in Yamagi.
In the first map (defens), once player gets inside the building and deals with some baddies, there are only two ways in which progression is possible - left or right along the green lights. At both ends, there is an elevator behind a closed door, larger door on the left and smaller door on the right.
Both of these should normally open after pressing the lift button, but not in Yamagi. Simply nothing happens and player is stuck. 
The door should open not immediatelly, but after a delay of about a 1,5 secs or so, so maybe that's the problem? Anyway, if I noclip inside the elevator and press the button to open the lift door, that works. Once outside the elevator and pressing the button, it doesn't work again.
Tested in 8.41 and latest source.

I tried the map (ver 1.02) in vanilla v3.20, v3.24, q2pro and q2rtx - all these work just fine.
I am not a mapper so playing with the ent file would likely not help much.

The map set is inside this archive: https://www.gamefront.com/games/quake-2/file/super-pack-sp-1
There is also an older 1.0 version and some 1.1 updated on moddb (which I don't even think is really 1.1). Nonetheless, I tried these in q2pro and no problem with these either.

BjossiAlfreds:
I just tested and can confirm this. I don't have time to do thorough investigation for now, but for those who want to take a crack at it, you can do "teleport 72 1780 -800" for example. And the func_door there has model "*121", for reference into the entity string of the map.

BjossiAlfreds:
I found out the problem and made a PR for the fix candidate. All the technical details are in the PR comment.

If you have the tools you can compile and test it: https://github.com/yquake2/yquake2/pull/1186

protocultor:
Just for the record, the new CI testbuilds generate packages ready to use. This is the Win32 one:
https://github.com/yquake2/yquake2/actions/runs/13500769151/artifacts/2641795994
(Appears in "_Testbuild for Win 32_" -> "_Upload testbuild package_" -> "_Artifact download URL_").

BjossiAlfreds:
> Just for the record, the new CI testbuilds generate packages ready to use. This is the Win32 one: https://github.com/yquake2/yquake2/actions/runs/13500769151/artifacts/2641795994 (Appears in "_All checks have passed_" -> "_Testbuild for Win 32_" -> "_Upload testbuild package_" -> "_Artifact download URL_").

[...]

PR #1186, opened 2025-02-24
Fixed doors not being usable with NULL activator

BjossiAlfreds:
This PR addresses issue reported in https://github.com/yquake2/yquake2/issues/1185.

When the button is pressed, it targets a `trigger_elevator` which works as a controller entity for `func_train` which is the elevator platform. Each button has a `pathtarget` that tells the `trigger_elevator` which `path_corner` to go to. Each `path_corner` has a pathtarget that targets the door of the respective floor. The failure here is that there is a NULL check for `activator` at the top of `door_use`. Since the elevator platform is managed by a controller entity, it has no activator set, which causes the NULL check to return early, preventing the door from opening.

This is another side-effect of all the NULL checks at the top of functions. `door_use` was needlessly checking for a NULL activator. This fix may fix related issues in other custom maps, and possibly subtle map bugs in official maps (though I have no examples of this).

I also spotted possible trouble if the `func_train` controlled by a `trigger_elevator` gets killtargeted or disappears for another reason, so I added safety checks for this.

### DIFF
--- a/src/game/g_func.c
+++ b/src/game/g_func.c
@@ -1429,12 +1429,12 @@ door_go_up(edict_t *self, edict_t *activator)
 void
 door_use(edict_t *self, edict_t *other /* unused */, edict_t *activator)
 {
-	if (!self || !activator)
+	edict_t *ent;
+
+	if (!self)
 	{
 		return;
 	}
-
-	edict_t *ent;
 
 	if (self->flags & FL_TEAMSLAVE)
 	{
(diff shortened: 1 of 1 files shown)


inf78:
I confirm it works as it should, great!
And as you mentioned, it was the main reason I reported this, since this could break any other map.

Thanks for such a quick fix!

Yamagi:
I'm always amazed that you guys fix my bugs faster than I can introduce them. Thank for doing this. :) Hopefully it was the last fallout of the sanity checks... About 15 years later, with a lot of more experience and better tooling I wouldn't add them again, instead focus on fixing the underlying bugs.

BjossiAlfreds:
> I'm always amazed that you guys fix my bugs faster than I can introduce them. Thank for doing this. :) Hopefully it was the last fallout of the sanity checks... About 15 years later, with a lot of more experience and better tooling I wouldn't add them again, instead focus on fixing the underlying bugs.

When I started thinking about the bug I actually had a hunch it might be another NULL activator bug. If that had not been the case, debugging would have taken more time I imagine.

I tend to avoid checking NULLs unless the code is meant to support and handle it, so for example I never test this/self pointers in callback functions. It might be an interesting strategy to include extra sanity checks in release builds and keep them to a minimum for test builds so testers crash the game and discover more bugs, and makes it easier to debug and fix them.

PR #1187, opened 2025-03-01
Update the CHANGELOG for the upcoming 8.50 release

Yamagi:
This is a draft for the CHANGELOG update for the upcoming 8.50 release. Feedback is welcome.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,33 @@
+Quake II 8.41 to 8.50:
+- Consistent bindings between multiple gamepad types. (by protocultor)
+- Improved menu navigation with gamepads. (by protocultor)
+- The SDL 3 code now supports the stable SDL 3 API. Several bugs with
+  SDL 3 were fixed. (by 0lvin, BjossiAlfreds, erysdren and protocultor)
+- Water warp effect for the OpenGL 1.4 renderer. (by protocultor)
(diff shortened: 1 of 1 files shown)


BjossiAlfreds:
"temporary spawnflags" should be "temporary spawnfields". And maybe you can include a line for fixing doors not opening in rare cases, from the recent PR merge. Other than that it looks good to me.

Yamagi:
I have pushed an update.

runlevel5:
Great thanks for yet another awesome release.

Just wondering if vulkan renderer will be bundled along in the future release?

0lvin:
Vulkan render will not be included to official release, and it will be released as separate [package](https://github.com/yquake2/ref_vk/releases) if be required by ABI compatibility reasons. 

In same time you could use [yquake2remaster](https://github.com/yquake2/yquake2remaster/releases) if you like to have all renders in single package.

protocultor:
About line 13, I wish the [OpenGL ES 1.0 renderer](https://github.com/yquake2/yquake2/pull/1128) was added in the text, even if its not recognized as "official". Maybe:
```
- Performance optimizations for the OpenGL 1.4 renderer, especially on
  embedded and mobile GPUs. Includes an alternative GL ES 1.0 renderer
  (by protocultor)
```

All the rest looks fine.

Yamagi:
I've incorporated the change.

PR #1190, opened 2025-03-09
Fix missing limits header in SDL2 module

efliks:
Hi there,
I was not able to build unless adding the limits.h include. Was it missing? If there is a better solution or I did something wrong, please disregard this tiny PR.
Greetings,
MF

PS. My build config:

Build configuration
YQ2_ARCH = x86_64 COMPILER = gcc
WITH_CURL = no
WITH_OPENAL = yes
WITH_RPATH = yes
WITH_SDL3 = no
WITH_SYSTEMWIDE = no
WITH_SYSTEMDIR =

### DIFF
--- a/src/client/input/sdl2.c
+++ b/src/client/input/sdl2.c
@@ -35,6 +35,7 @@
  */
 
 #include <SDL2/SDL.h>
+#include <limits.h>
 
 #include "header/input.h"
 #include "../header/keyboard.h"
--- a/src/client/input/sdl3.c
+++ b/src/client/input/sdl3.c
@@ -35,6 +35,7 @@
  */
 
 #include <SDL3/SDL.h>
+#include <limits.h>
 
 #include "SDL3/SDL_gamepad.h"
 #include "SDL3/SDL_properties.h"


protocultor:
No, not at all, I just find weird that it doesn't happen in the Ubuntu versions I've tested.
Does Debian have SDL3 packages? Can you test compilation with that?

efliks:
> No, not at all, I just find weird that it doesn't happen in the Ubuntu versions I've tested. Does Debian have SDL3 packages? Can you test compilation with that?

Currently, I only have libsdl2-dev installed, and libsdl3-dev does not seem to exist in Debian stable. It only exists in Debian unstable: https://packages.debian.org/source/sid/libsdl3

runlevel5:
FYI I do not run into any compilation issue on Fedora (rawhide with SDL2 2.32.50) so IMHO it is very likely an issue of Debian.

DanielGibson:
I don't get what you all are discussing about.

The code uses `SHRT_MAX`, which is defined in `<limits.h>`, so it should be included.
Seems like so far it was included **by pure chance** via some other header (or probably a chain of several headers), and on @efliks system that's not the case for some reason.

That's no defect in debian, just add the `#include` :-p

runlevel5:
@DanielGibson I agree with you after seeing your comment

runlevel5 (approved):

Yamagi:
Thank you :)

PR #1193, opened 2025-03-17
mod names list static allocation free at exit approach

devnexen:


0lvin:
Instead atexit could be called in Qcommon_Shutdown, as used in all other subsystems. What do you think?

devnexen:
You probably mean calling Mods_NamesFinish there instead right ? no point of calling atexit at this stage :)

0lvin:
CL_Shutdown looks as right place to deallocate list of modules.

Yamagi:
Thank you :)

PR #1195, opened 2025-03-18
Small fix for Flick Stick

protocultor:
#1147 was made to avoid the obligatory 60Hz requirement to use Flick Stick. Unfortunately, I made a bad judgement in terms of the way the time difference calculation was made.
What was used to obtain "the current time" at execution was `cls.realtime`, which comes from `curtime` in `src/common/frame.c`, which is close but not exact enough for what is needed. This PR fixes this by using `sys_frame_time`, which is also a variation of `Sys_Milliseconds()`, but at least its calculated locally in the `sdlX.c` file, being updated more timely.
Thanks to the people at the Gyro Gaming discord for bringing this to my attention.

### DIFF
--- a/src/client/input/sdl2.c
+++ b/src/client/input/sdl2.c
@@ -1312,7 +1312,7 @@ IN_FlickStick(thumbstick_t stick, float axial_deadzone)
 			// Flicking begins now, with a new target
 			is_flicking = true;
 			flick_progress = 0.0f;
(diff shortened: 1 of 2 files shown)


Yamagi:
I vaguely remember that I once stumbled  into exactly that problem while debugging some strange timings vs. input problems. Thank you for fixing this.

PR #1198, opened 2025-03-29
gl3 issues fixes.

devnexen:
segfault on video mode changes from no fullscreen to native/fullscreen or vice-versa. memory leak on SDL init as we dlload the gl3 lib dynamically for mode changes.

### DIFF
--- a/src/client/refresh/gl3/gl3_main.c
+++ b/src/client/refresh/gl3/gl3_main.c
@@ -379,10 +379,14 @@ SetMode_impl(int *pwidth, int *pheight, int mode, int fullscreen)
 
 	R_Printf(PRINT_ALL, " %dx%d (vid_fullscreen %i)\n", *pwidth, *pheight, fullscreen);
 
-
 	if (!ri.GLimp_InitGraphics(fullscreen, pwidth, pheight))
 	{
 		return rserr_invalid_mode;
+	}
+
+	if (mode == -2 || fullscreen)
+	{
+		GL3_BindVBO(0);
 	}
 
 	/* This is totaly obscure: For some strange reasons the renderer


src/client/refresh/gl3/gl3_main.c
```diff
@@ -684,6 +688,7 @@ GL3_Shutdown(void)
 		GL3_SurfShutdown();
 		GL3_Draw_ShutdownLocal();
 		GL3_ShutdownShaders();
+		GL3_ShutdownLibrary();
```
DanielGibson, line 691: Shouldn't this be done after GL3_ShutdownContext()?
I'm not sure, but it could be that SDL_GL_DestroyContext() uses libgl?

devnexen: Yes good and no relation with gl buffers neither.

DanielGibson:
By the way, shouldn't the same problem exist in GL1?

devnexen:
I ve tried, but did not reproduce with gl1.

Yamagi:
Can you describe how the problem is reproduced? I tried it under Linux by switching from windowed mode to native fullscreen and didn't saw a crash.

devnexen:
This is what I get when I do what you mention

image

devnexen:
image

Yamagi:
Is the game build with ASAN and / or UBSAN when it happens?

I tried it this way:

1. Build without sanitizers
2. Start the game, enter base1
3. Switch between windowed and native fullscreen mode by alt-enter.

No matter how often I hit alt-enter the game doesn't crash.

devnexen:
it s build with ASAN with clang

Yamagi:
It's a problem with ASAN. ASAN / UBSAN and renderer or game dll reload doesn't work. Which isn't documented :/

ASAN / UBSAN need library objects to exist from the moment they are loaded until the program is quit. To achieve that we set the objects created by `dlopen()` to sticky when the game is build with sanatizers. The code is:

* https://github.com/yquake2/yquake2/blob/master/src/backends/unix/system.c#L450 for the game
* https://github.com/yquake2/yquake2/blob/master/src/backends/unix/system.c#L654 for the renderer lib

Setting the library objects to sticky breaks the assumption that unloading and reloading the renderer libraries reset the renderer state. You will end up with at least parts of the state before the reload remaining and the renderer crashes.

Setting the VBO is properly right and fixes some real corner case problems. Unloading the renderer library is likely wrong and can lead to crashed further down the callstack when destroying the window or quitting SDL. If it helps agains crashes with ASAN / UBSAN enabled it's pure luck.

The luck depends on the GPU driver. With my Radeon card it crashes somewhere in libGL and not even in Quake II.

Yamagi:
Thank you :)

Issue #1201, opened 2025-04-16
Investigating Quake 2's slow loading times


BjossiAlfreds:
I wanted to take a closer look at Quake 2's slow loading times.

The first thing I looked at was `CL_PrepRefresh` where all the loaded models are printed to the console at the end of the loading process. I felt it was strange why this was taking so long. My first experiment was to simply comment out the call to `SCR_UpdateScreen` inside the loop. This dramatically shortened the load time for me, as much as 4 seconds on a big level like city3 or rware2. This speed-up could be system dependent maybe.

However this only seems to impact loading a map from the console and loading autosaves, level transitions are still really slow. I don't really know why that is, I'm still studying the client-server back and forth involved. My hunch is that the slow loading is caused by a mix of networking and synchronization delays, and I wonder if there are clever ways to code around that.

If you want to try my change to PrepRefresh you can check out my silent-precache branch on my YQ2 fork. Basically I removed all the console printing from the loop and instead just print "models". I could do a PR if you like this change.

0lvin:
> My first experiment was to simply comment out the call to SCR_UpdateScreen inside the loop.

In theory, code could use dprintf for print models and skip screen refresh with developer flag equals to false 

I will try it in remaster repository. Remaster repository has much more aggressive caching, like left three last maps in memory, so level change sometimes feels faster.

BjossiAlfreds:
Even if dprintf works, I don't really view that information as useful for showing the user. I'm content with just seeing "models", personally.

As for caching, that may not actually help with this client. A modern PC's brute force should be able to load all the assets in a fraction of a second, maybe I'm overestimating the speed of modern PCs? I guess the client can cache some of the data as long as the map doesn't change, though.

DanielGibson:
You should do proper profiling of what takes so long (maybe integrate [Tracy](https://github.com/wolfpld/tracy) or [Remotery](https://github.com/Celtoys/Remotery) for that) to figure out what really uses up the time, and why (e.g. is there unnecessary blocking with mutexes or whatever involved, or even waiting for vsync? `SCR_UpdateScreen()` could be caused by the latter, maybe it could just be done less often?).

[...]

PR #1203, opened 2025-04-22
Attempts to speed up level load times

BjossiAlfreds:
This PR is my attempts to address points discussed in https://github.com/yquake2/yquake2/issues/1201 to speed up level load times, or rather, client connection times more technically speaking.

1. Model names and clients printed to console

The loops that loop through the model configstrings print out the string value to console, including empty lines that represent inline models. Each loop iteration calls `SCR_UpdateScreen` which is vey expensive.

I simply removed printing from the loops and instead just print "models" or "clients".

This change is part of the PR although @0lvin has made a commit that addresses the same issue. We can use that solution instead if preferred.

2. Message utilization

After the server receives the "new" message from a client, it will go through a process of sending configstrings and entity deltas in bulk messages. In the original code the message is only filled until `MAX_MSGLEN / 2` is exceeded, or 700 bytes. This will leave over 600 bytes unused.

I changed the code by creating `MSG_ConfigString_Size` and `MSG_DeltaEntity_Size` for predicting how much space the data will take, and introduced 2 defines: `CMD_MARGIN` which reserves 40 bytes for the command appended at the end of the message, and `SAFE_MARGIN` which reserves space for overhead (like packet header data) and misc additions. I'm currently using a value of 24 which seems to work well.

A `SAFE_MARGIN` value of around 500-600 would be fairly close to the original behavior.

I'm not aware of the server adding any other data to these configstrings/baselines messages while getting ready to spawn in. If unexpected overflows happen we can increase the value of `SAFE_MARGIN`.

3. Server slow to send client messages

The server ticks at 10 FPS. This is correct for active clients, but I feel this is fairly slow for clients who are still getting ready.

Most of the slowdown is with the configstrings/baselines messages. After receiving a "new" message, the server sends the first bulk of configstrings to the client along with a stuffed command to request the next bulk. The client then parses it and sends the stuffed command back to the server. For large maps this can take dozens of back-and-forths, which is several seconds of delay.

My first "toy attempt" to solve this is by splitting `SV_SendClientMessages` into 2 functions, the other being `SV_SendPrepClientMessages`. The latter function is called every server frame and only sends messages to clients who are not yet spawned in, while the original function still takes care of sending messages every 10 server frames (~0.1 second).

4. Reconnection delay

Reconnection happens on level transitions or when a map loads after a video (like the intro or end-of-unit videos). This one still eludes me. At the end of `SV_Map` the server first broadcasts a "changing" command, sends it, and then broadcasts a "reconnect". For reasons I still can't figure out, it takes the reconnect command up to a second to be processed client-side. Is the server slow to send it or the client slow to read it? Not sure at this time.

In any case, I simply removed the call to `SV_SendClientMessages` after broadcasting the "changing" command and that removed the delay. The client is fast and responsive to reconnect. 

NOTE: Removing the call to `SV_SendClientMessages` was made possible by the change I described in 3. I guess the reason for it in the original code was to make sure the clients receive "changing" quickly, instead of having to wait for the next full server frame.

I was thinking this delay may be intentional but I don't really see a reason for it, after `SV_SpawnServer` the server should be ready to accept clients as far as I can tell.

5. Misc stuff

I spotted that the index parameter to the "baselines" and "configstrings" server commands is not checked, so buggy/malicious clients could send negative values leading to reading out of bounds.

I also made a wrapper `MSG_WriteConfigString` function, not necessary but figured it is better than sending the components individually in every use case.

---

With all of these optimizations I am getting very large reductions in time spent on the loading screen. That includes starting a new game through the menu, map command, and in-game level transitions

These changes are experimental so this is a draft/test PR. In particular I don't feel confident in my solutions to 3 and 4.

I have only tested this locally, where server and client are both running on the same system/in the same process. I feel it is important to test this in a real online setting, where server and client are on different networks.

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -13,6 +13,7 @@ have been renamed. The prefixes are:
 * `ogg_`: Ogg/Vorbis music playback.
 * `r_`: Common to all renderers.
 * `s_`: Sound system.
+* `sv_`: Server 
 * `sw_`: Software renderer.
 * `vid_`: Video backend.
 
@@ -56,6 +57,40 @@ it's `+set busywait 0` (setting the `busywait` cvar) and `-portable`
   sleep and tells the operating system to send a wakeup signal when it's
   time for the next frame. The latter is more CPU friendly but can be
   rather inaccurate, especially on Windows. Use with care.
+
+* **sv_optimize_sp_loadtime**<br />
+  **sv_optimize_mp_loadtime**: These cvars enable/disable optimizations
+  that speed up level load times (or more accurately, client connection).
+
+  sp stands for singleplayer and mp for multiplayer, respectively.
+  The sp version is enabled by default (value 7) while multiplayer is 0.
+
+  The cvar value is a bitmask for 3 optimization features:
+  * **1: Message utilization**<br />
+    When the server sends the client configstrings and other data
+    during the connection process, the message buffer is only used
+    around 50-60%. When this flag is enabled, the message is used
+    much better, dramatically reducing the amount of messages needed
+    to deliver all the data to the client.
+
+  * **2: Server send rate**<br />
+    By default, the server sends messages to clients once every
+    0.1 seconds, roughly. This slows down sending data to clients,
+    especially in singleplayer. This is normal for active clients,
+    but for connecting/inactive clients, this delay is unnecessary.
+    When this flag is set, the server will send messages to
+    inactive clients ~8x more frequently.
+
+  * **4: Reconnection**<br />
+    When the server changes maps, like on level transitions,
+    the server first sends a "changing" command to all clients,
+    and then a "reconnect" command. The delay between these commands
+    can be quite long, ~1 second. This flag will avoid this delay when
+    set, by sending the two commands within the same message.
+
+  Simply add these flag values together to get the cvar value you want.
+  For example, sendrate + reconnect = 2 + 4 = 6.
+  Set to 7 for all optimizations, or 0 to disable them entirely.
 
 * **cl_maxfps**: The approximate framerate for client/server ("packet")
   frames if *cl_async* is `1`. If set to `-1` (the default), the engine
--- a/src/client/cl_main.c
+++ b/src/client/cl_main.c
@@ -222,9 +222,7 @@ CL_Record_f(void)
 			}
 
 			MSG_WriteByte(&buf, svc_configstring);
-
-			MSG_WriteShort(&buf, i);
-			MSG_WriteString(&buf, cl.configstrings[i]);
+			MSG_WriteConfigString(&buf, i, cl.configstrings[i]);
 		}
 	}
 
--- a/src/client/cl_network.c
+++ b/src/client/cl_network.c
@@ -444,6 +444,13 @@ CL_Changing_f(void)
 
 	SCR_BeginLoadingPlaque();
 	cls.state = ca_connected; /* not active anymore, but not disconnected */
+
+	/* reset this to 0 just in case it didn't get a chance to settle normally
+		this became a problem with the faster client connection changes
+		but is a good idea to do this regardless
+	*/
+	anykeydown = 0;
+
 	Com_Printf("\nChanging map...\n");
 
 #ifdef USE_CURL
--- a/src/client/cl_view.c
+++ b/src/client/cl_view.c
@@ -240,7 +240,7 @@ CL_PrepRefresh(void)
 {
 	char mapname[MAX_QPATH];
 	int i;
-	char name[MAX_QPATH];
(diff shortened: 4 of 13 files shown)


BjossiAlfreds:
Discovered the first little bug. When walking into a level transition while holding a key like W for moving forward, `anykeydown` would leak into the next map and permanently add `BUTTON_ANY` to every usercmd.

It seems the any key mechanism relies on the client to settle it back to 0 by running extra `Key_Event` calls. I'm actually surprised this thing has never gotten stuck for me before, seems like a really sensitive balance.

I added a hard reset to 0 into the changing command. That seems to fix it. Probably a good idea to do that regardless, just in case.

Yamagi:
So... I finally found some time to have a deeper look. These are some good findings and on my for todays standards rather slow laptop the results are immediately visible.

Going through the changes:

* *Model names and clients printed to console*: We should (and will) merge this. It's without risk, makes things a lot faster and no one needs this lines anyways.
* *Message utilization*: This should be rather low risk, too. It might be a good idea to have a look at q2pro or r1q2. Chances are good that they already implemented some kind of optimization for this problem. Regarding `SAFE_MARGIN` we must take into account that  Quake II has a max message size of 1400 with is ... brave ... for a UDP based protocol. To be on the safe side a message should never be bigger than 1280 bytes. Above that one can never be sure that package won't get fragmented or dropped.
* *Server slow to send client messages*: It's hard to guess the risk. This might be totally harmless or there might be other clients or even things like mods which somehow work by luck as long as the message flow isn't too fast... I don't really have an opinion here without more testing.
* *Reconnection delay*: More or less the same as the last point.
* *Misc changes*: We should merge these.

Aaaand... the `anykeydown` might be the last missing bit to finally solve the problem with aborting cinematics. Over the years I have added several fixes to different parts to the client to ensure that no input data leaks over a map change skipping cinematic. It has gotten better, up to the point that it only happens in maybe 1 out of 50 cinematics. Too few to debug it further.

BjossiAlfreds:
Regarding message size, as far as I can tell from the original `SV_Multicast` and `SV_SendClientMessages`, the server sends only reliable messages (`netchan.message`) to non-spawned clients, and my code doesn't change that. So the big configstrings/baselines packets would only be sent that way. So I don't think the 1200 limit is anything to worry about.


Regarding server send rate, I forgot to mention this earlier but there is actually a case where the server rapid-fires messages to the client: with the `map` command in single-player. I was surprised why that seemed to load faster than level transitions or a new game. One reason was the reconnection delay but the other was that the server was not doing any sleep frames for the first several frames because `svs.realtime >= sv.time`.

Unless this is YQ2-specific behavior, I think we can infer that all other Q2 clients that implement the original protocol can handle the server sending messages very rapidly.


Regarding reconnection delay, it causes the `changing` and `reconnect` commands to be sent to the client in the same packet. I don't think that should cause any problems. There would just be a split second delay where the client is still running the old map until the server has finished `SV_SpawnServer` and sends the commands to change and reconnect.


I did get an idea, I could make 2 cvars: `sv_optimize_sp_loadtime` and `sv_optimize_mp_loadtime`. A value of 0 is off, 1 = optimize send rate, 2 = optimize reconnect and 3 = optimize both.

I imagine single-player should be fairly safe. The connection is through localhost and both server and client are YQ2. I have also been testing and playingh and all is working fine by the looks of it. So I'd set the cvar default to 3 (opt-out).

Multiplayer is more uncertain hence why I'd set the defailt value to 0 (opt-in). Connection is (usually) across networks or computers, and server and client can mix and nmatch. Also I have not tested it at all in multiplayer.

And if no problems are reported or encountered and confidence grows, we can in the future remove the cvars and make the feature fully built-in.

BjossiAlfreds:
I went ahead and made a commit with these cvars I suggested above.

They are bitmask values.
1 = optimize message utilization
2 = optimize server sendrate
4 = optimize reconnection on map changes

Default for SP is 7, 0 for MP.

These cvars are not latched so you can change them quickly and test the difference as well as for debugging if you are running into problems and suspect the optimizations are at fault.

Yamagi:
Thank you for the cvars. I like it this way. I'll do a 8.51 hotfix release in the next few days and merge this afterwards.

Yamagi:
Is it ready or have you planed more changes?

BjossiAlfreds:
I feel it is ready now. The other day I just cleaned up some previous commits and improved the client dropping on message overflow a bit since that code was a little wonky with the separated `CL_SendClientMessages` and `SV_SendPrepClientMessages`. I've been playing in-game and all seems to work for me.

BjossiAlfreds:
I went ahead and rebased the branch on current master. You can merge it now.

Yamagi:
Thank you :) Let's see if anything comes up.

PR #1210, opened 2025-05-05
Add const, static and fix function definition

0lvin:
Found in https://github.com/yquake2/yquake2remaster/issues/38


0lvin:
Before applying #1208, function parameters and keydown were defined as boolean, but used as int in different place.

Yamagi:
Thank you :)

Issue #1213, opened 2025-05-22
[Original bug] 2 silent doors in jail4 (Torture Chamber)


BjossiAlfreds:
In jail4 (Torture Chambers) there is this one room that has doors that make no sound at all. `teleport 200 300 -600` to teleport there. I had always noticed this but never looked into it until now. These doors have `"sounds" "1"` set in spawnflags which is easy enough to fix.

Coincidentally I also found 2 ambient sound `target_speaker` entities with spawnflag 2048 set for seemingly no reason, as all the other ambient sound speakers don't have it.

I can do a PR for this if you like.

Yamagi:
I'm always amazed that you find this little nits no one noticed over all the years. If course I wold like to have a PR for this.

BjossiAlfreds:
> I'm always amazed that you find this little nits no one noticed over all the years. If course I wold like to have a PR for this.

I had actually noticed this since the 90s when I was first playing the game, I was just coincidentally reminded of it while playing the other day and figured I'd look into it.

The PR is ready now.

PR #1214, opened 2025-05-24
Mapfix for jail4: 2 silent doors

BjossiAlfreds:
This PR addresses https://github.com/yquake2/yquake2/issues/1213

* Fix for 2 silent doors
* Add missing 2048 spawnflag to some non-DM entities like `trigger_key` and `path_corner` used by monsters
* Removed 2048 spawnflag for 2 ambient `target_speaker`
* Added missing targetnames to `info_player_coop`. This is also done as a codehack
* Removed unused targets from some `light` entities

As always, all the changes are documented in comments inside the ent file.
And would be nice if others could do a quick run of the level to test, in case I missed or broke something.


Yamagi:
> And would be nice if others could do a quick run of the level to test, in case I missed or broke something.

I just did a quick run and everything is okay.

PR #1217, opened 2025-05-28
Mapfix for hangar1: Fixed target_speakers and trigger_push overspeed

BjossiAlfreds:
This PR adds more mapfixes to the hangar1.ent file.

* Added missing spawnflag 1 to a few ambient `target_speaker`. These are water pumping sounds playing by the big water pipe by the water
* Fixed `trigger_push` overspeed. The default (1000) will overflow the 6-byte interger vector used by clients. 1000 * 10 * 8 = 14464 (65536 overflow bit thrown away). So I gave it a speed of 180 (14464 / 80 = 180.8), which is the actual speed it has in-game
* Added spawnflag 2048 to a bunch of non-DM entities

On a side note, I wanted to make the commander's head button permanently pressed after using the key, but there is no way to do this without code extensions as far as I know. Unless maybe using a `func_wall` to replace the button entity when using the key. I left it unchanged.

### DIFF
--- a/stuff/mapfixes/baseq2/hangar1@09a0.ent
+++ b/stuff/mapfixes/baseq2/hangar1@09a0.ent
@@ -1,3 +1,14 @@
+// FIXED ENTITY STRING (by BjossiAlfreds and ConHuevosGuey)
+//
+// 1. Fixed missing spawnflag 1 on ambient target_speaker (b#1)
+//
+// 2. Fixed 2x stuck monster_infantry (b#2)
+//
+// 3. Flagged some non-DM entities with 2048 (b#3)
+//
+// 4. Fixed trigger_push overspeed (b#4)
+//
+// 5. Optimized trigger chains (b#5)
 {
 "nextmap" "hangar2"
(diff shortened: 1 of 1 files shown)


Yamagi:
> On a side note, I wanted to make the commander's head button permanently pressed after using the key, but there is no way to do this without code extensions as far as I know. Unless maybe using a func_wall to replace the button entity when using the key. I left it unchanged.

I wouldn't mind a code change as long as is optional. The Question is if it's worth the effort.

BjossiAlfreds:
In my own mod I added a spawnflag 1 to `func_button` that makes the button pause-able in its pressed/unpressed state when triggered. So the `trigger_key` targets the button and makes it stay down. New spawnflags are very risky though, because the maps don't always get them right. Too risky for a YQ2 release.

You are right though, probably not worth the effort. We'd have to do extensions with code hacks, so that the ent files remain compatible with original spawning code. Would be annoying to see "yq2_spawnflags is not a field" warnings when using the ent file with custom DLLs...

PR #1218, opened 2025-05-28
Fixed SP_CreateUnnamedSpawn spawning entitiy unnecessarily

BjossiAlfreds:
I noticed the `G_Spawn` call in `SP_CreateUnnamedSpawn` was done unconditionally, so for every info_player_start in every level you load, it was spawning equally as many `noclass` entities that would never be freed.

This PR fixed that issue and I also found the code pretty copy-pasty so I wrapped it into a helper function.


Yamagi:
It must have been nearly 15 years since I wrote that hack... Thank you for cleaning it up.

Issue #1223, opened 2025-06-08
[Original bug] Monsters some times act confused while following combat points


BjossiAlfreds:
Steps to reproduce:
1. Load map `jail3`
2. Set `notarget` and go up the stairs ahead
3. Shoot the medic with the blaster
4. Even though his `point_combat` is straight ahead, he turns 180 degrees, runs to the wall, turns around again and finally goes up the stairs

You also see a similar thing happen in map `lab` with the medic near the blue key door - turns in a weird way before going for his combat marker. You can probably find many other examples of this with a keen eye.

This happens because monsters don't set their `ideal_yaw` in `FoundTarget`. Same problem in `point_combat_touch` when they target the next marker in a longer path. In the case of the medic in jail3, his ideal yaw is 0 degrees, so in `SV_StepDirection` he gradually turns to it, then runs at the wall, `SV_NewChaseDir` then finally starts guiding him towards his goal entity.

I will do a PR to address this.

BjossiAlfreds:
Yeah the expansions have the same problem. I just did PRs for those as well.

PR #1224, opened 2025-06-08
Fixed monsters acting confused while following combat points

BjossiAlfreds:
This PR addresses https://github.com/yquake2/yquake2/issues/1223.

Simply setting `ideal_yaw` correctly in `FoundTarget` and `point_combat_touch` fixes the issue.


Yamagi:
Thank you for the three PRs. :)

Issue #1230, opened 2025-06-11
q2pro doesnt have railgun graphical glitch like yamagi


xboxownzuall:
you might want to ask if you can borrow some of skullers code so you can fix it as you can see it will render back of the gun hell even in yamagis software rendering mode it doesnt do it so its all a freaking cop out lol not to fix it man period.

![Image](https://github.com/user-attachments/assets/e003feab-4136-4ca3-bc27-64263be5b557)

xboxownzuall:
You guys keep dodging the question though man. q2pro doesnt do this online muchless offline and to have normal arm maximum visable arm distance in an online server use cl_adjustfov 1 it does not clip into your players view clip into it.

its something off with yamagis player height or gun placement by default causing it to clip into the screen period.

xboxownzuall:
You guys keep dodging the question though man if this was all the case. q2pro  would do it also but it doesnt do it in online servers muchless offline and to have normal arm maximum visable arm distance in an online server use cl_adjustfov 1 it does not clip into your players view like it does yamagi.

its something off with yamagis player height or gun placement by default values causing it to clip into the screen period.

xboxownzuall:
That bug is from modified 3.20 for modern resolutions or mark shans 3.24 patch. obviously q2pro skuller tweaked it so the guns dont clip into the players screen no matter what fov you are using especially in an online server where you cant fine tune every adjustment like offline. skuller did something so it doesnt clip into your screen period inspect his code and see why his doesnt.

maraakate:
Shut the fuck up

Yamagi:
Again, we had this before: Vanilla Quake II was build under assumptions. In this case that the display has an aspect of 4/3. The world moved on an sometimes we need to break these assumptions to stay compatible with modern hardware. There are several ways how to do that, often all of them have side effects. There is no right or wrong, q2pro just did it somehow different than YQ2.

If you don't like YQ2s approach you have several options:

* Use the soft renderer. It doesn't support projection matrices and renders the gun distorted but doesn't have clipping issues.
* Adjust the gun fov. That also distorts the gut but solves the clipping.
* Use the mod @protocultor pointed out.
* Send a pull request which implements the q2pro approach as an **optional** way to render the guns.
* Just use q2pro and shut the fuck up.

Issue #1229, opened 2025-06-11
PLEASE FIX RAILGUN GRAPHICAL GLITCH HOLY HELL ITS BEEN THERE FAR TOO LONG


xboxownzuall:
FIX THE HOLE IN THE RAILGUN WHEN IT RECOILS MAN HOLY SH*T. YAMAGI IS THE ONLY SOURCE PORT WITH THIS ISSUE MAN.

ITS WITH THE OPEN GL RENDERING MODES/VULKAN THE SOFTWARE RENDERER DOESNT DO IT.

HERE IS COMPARISON WITH Q2PRO SOURCE PORT IT DOESNT SUFFER FROM THAT GLITCH USING SAME FOV 110 AND RESOLUTION.

FIRST IMAGE YAMAGI 8.51 SECOND IMAGE Q2PRO

![Image](https://github.com/user-attachments/assets/1a76534f-3443-4748-95ce-0bbd43426f65)

![Image](https://github.com/user-attachments/assets/ade5ce15-7adf-40b2-b25b-1198ab1ad332)

Yamagi:
There's a problem with your shift key. You should fix it.

The weapon models are too short to render them with correct gun fov on widescreen displays without getting cut off at the rear end. If you don't like it, change the gun fov. See the documentation: https://github.com/yquake2/yquake2/blob/master/doc/030_configuration.md#field-of-view

PR #1232, opened 2025-06-13
Revert GL1 big lightmaps (#1108) + fixes

### LINKED ISSUE #1220: Bug in mine2 red light
One of the lights on the Mine2 map does not light up correctly in either singleplayer or multiplayer.

edit:
i use opengl1 on linux

edit 2:
opengl3 and software looks fine.

Image
Image
Image
Image
Image
Image

### LINKED ISSUE #1231: even with md2 weapon models the railgun still clips through players screen as seen here.
sorry for coming across as douche btw just trying to get it fixed.

its the gun_x setting not adjusted right too too far left toward center of the screen clipping into your view and when you adjust it to like gun_x 3 or something like that it no longer does it.

see if you can have it default more to off to the right of the screen and so that it works even in an online in a server since servers dont allow fine tune adjustment like that offline, but skuller somehow fine tuned it so it works even in an online server though by default its adjusted right out of the box and works even online his fine tunings.

this is a problem with 3.24 patch and yamagi and directXQII/DirectQII all of you guys just left it stock settings after adding higher fov than 90 with guns being visable to 3.20 you left stock xyz settings from 3.20 patch didnt do any fine tuning.

https://www.nexusmods.com/quake2/mods/45

this screenshot im using the md2 weapon models and its still clipping doing it obviously it's default gun_x settings messed up too much towards center of screen your weapon is.

Image

protocultor:
Was mostly done in #1124, but this PR completes the cleaning job, restoring _classic Q2_ static handling of lightmaps and _scrap_ (atlas for HUD elements).
The work done by commit 4503333b should alleviate what big lightmaps used to do.
Note that the _scrap_ remains at its vanilla Q2 size (256x256), with an additional transparent border for each element inside it.

Additionally, _GL1 multitexturing_ now properly detects when a lightmap's cache reset is needed, fixes #1220 and every surface with more than 2 _static light styles_ (normal, flicker, strobe, pulse, etc.) applied to them.

Also, OpenGL's perspective projection (_frustum_) has been modified; `zNear` is down from `4` to `2` in all GL renderers, GL3 included. Fixes #1231, #1230, #1229, world hunger, and who knows what else.
It just might introduce other visual bugs; we will know soon enough. If it does, please only revert this individual commit: 56b7aa6.

### DIFF
--- a/.github/workflows/win32.yml
+++ b/.github/workflows/win32.yml
@@ -66,8 +66,8 @@ jobs:
         cp doc/080_contributing.md publish/quake2-win32-${{github.sha}}/misc/docs/080_contributing.txt
         cp doc/090_filelists.md publish/quake2-win32-${{github.sha}}/misc/docs/090_filelists.md
         # SDL2
-        wget -c https://github.com/libsdl-org/SDL/releases/download/release-2.32.0/SDL2-2.32.0-win32-x86.zip
-        unzip -o SDL2-2.32.0-win32-x86.zip
+        wget -c https://github.com/libsdl-org/SDL/releases/download/release-2.32.8/SDL2-2.32.8-win32-x86.zip
+        unzip -o SDL2-2.32.8-win32-x86.zip
         cp SDL2.dll publish/quake2-win32-${{github.sha}}/
         # openal-soft
         wget -c https://github.com/kcat/openal-soft/releases/download/1.23.1/openal-soft-1.23.1-bin.zip
--- a/.github/workflows/win64.yml
+++ b/.github/workflows/win64.yml
@@ -68,8 +68,8 @@ jobs:
         cp doc/080_contributing.md publish/quake2-win64-${{github.sha}}/misc/docs/080_contributing.txt
         cp doc/090_filelists.md publish/quake2-win64-${{github.sha}}/misc/docs/090_filelists.md
         # SDL3
-        wget -c https://github.com/libsdl-org/SDL/releases/download/release-3.2.4/SDL3-3.2.4-win32-x64.zip
-        unzip -o SDL3-3.2.4-win32-x64.zip
+        wget -c https://github.com/libsdl-org/SDL/releases/download/release-3.2.16/SDL3-3.2.16-win32-x64.zip
+        unzip -o SDL3-3.2.16-win32-x64.zip
         cp SDL3.dll publish/quake2-win64-${{github.sha}}/
         # openal-soft
         wget -c https://github.com/kcat/openal-soft/releases/download/1.23.1/openal-soft-1.23.1-bin.zip
--- a/src/client/refresh/gl1/gl1_buffer.c
+++ b/src/client/refresh/gl1/gl1_buffer.c
@@ -183,7 +183,7 @@ R_ApplyGLBuffer(void)
 			if (gl_config.lightmapcopies)
 			{
 				// Bind appropiate lightmap copy for this frame
-				lmtexture += gl_state.max_lightmaps * cur_lm_copy;
+				lmtexture += MAX_LIGHTMAPS * cur_lm_copy;
 			}
 			R_MBind(GL_TEXTURE1, lmtexture);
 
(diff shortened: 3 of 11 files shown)


Yamagi:
> Was mostly done in https://github.com/yquake2/yquake2/pull/1124, but this PR completes the cleaning job, restoring classic Q2 static handling of lightmaps and scrap (atlas for HUD elements).
> The work done by commit https://github.com/yquake2/yquake2/commit/4503333b3a5c170a1388066a40d9c82daea35053 should alleviate what big lightmaps used to do.
> Note that the scrap remains at its vanilla Q2 size (256x256), with an additional transparent border for each element inside it.
> 
> Additionally, GL1 multitexturing now properly detects when a lightmap's cache reset is needed, fixes https://github.com/yquake2/yquake2/issues/1220 and every surface with more than 2 static light styles (normal, flicker, strobe, pulse, etc.) applied to them.

Thank you for this round of fixes. I known how frustrating all these corner cases can be :/

> Also, OpenGL's perspective projection (frustum) has been modified; zNear is down from 4 to 2 in all GL renderers, GL3 included. Fixes https://github.com/yquake2/yquake2/issues/1231, https://github.com/yquake2/yquake2/issues/1230, https://github.com/yquake2/yquake2/issues/1229, world hunger, and who knows what else.
> It just might introduce other visual bugs; we will know soon enough. If it does, please only revert this individual commit: https://github.com/yquake2/yquake2/commit/56b7aa607e8c1003201db40582bde4eb423c4f59.

@DanielGibson and me both vaguely remember that there was a reason to keep zNear at 4. And of course we don't remember what it was... We've tested the obvious cases with the branch applied and didn't see any problems, even the always problematic plasma beam in rogue looks fine. Let's merge it and see who complains next time and what insults they throw at us.

Yamagi:
Not in baseq2 or the addons and likely not caused by `zNear`. Setting `r_farsee 1` can cause z-fighting, but that is expected. `gl_zfix` is another hack with possible side effects including z-fighting and again unrelated to `zNear`, I think. I really don't remember.

Yamagi:
Mkay, I found it while testing #1237. Setting `zNear` to something smaller than `4` breaks not-so-wide resolutions in some cases. For example GL1 at `r_mode 16` aka 1280x1024:

[Bildschirmaufnahme_20250630_191046.webm](https://github.com/user-attachments/assets/31909228-ba93-49b5-9727-86daabda64b0)

Should I just revert it and we must endure another century of world hunger or has someone a better idea?

maraakate:
I should also make clear that I am absolutely NOT slamming the author's of the PRs. Rather, I am saying from years of hacking Q2 engine that many changes like these that seem clever and fix things end up being a trade off of another bug or cause weird subtle bugs that dont get reported for months because the game doesn't get played hard enough through the entire campaign.

Yamagi:
> Im saying this as a friend. These new PRs sound cool, but then they end up with very subtle bugs that involve many play throughs at your expense. Speaking from wayyy too much experience with Daikatana 1.3. I think you might want to consider serious proof of testing with these PRs that do such things. You'll find months or a year later you end reverting stuff. Q2 engine too volatile to fix it "right" on a technical level.

You're totally right that many well-meant bugfixes have a potential for introducing new bugs. That's why I'm cautious regarding changes to the 'core logic' or whatever we want to call it. Especially everything that includes changes to the game code, movement, entity handling and timings often make things worse. These PRs are somewhere in between. They mostly touch parts which are not so critical and add real value. Without the recent work by @protocultor - just to name him as one example - YQ2 wouldn't run nearly half as good on older low end GPUs and single board computers, we wouldn't have good support for game controllers, etc. Therefor I'm keeping merging this PRs as long as I'm confident that the submitters are up to the task fixing the fallout. And most of them totally are, they are often way better software developers than ever was and will be.

0lvin:
Could zNear be made cvar, and let's player what is the best for and name it `feature`? What do you think?

protocultor:
> Could zNear be made cvar, and let's player what is the best for and name it `feature`? What do you think?

Agree with this. It's what Q2Pro does, although it doesn't even have it documented. I'll include it in #1237

Yamagi:
Good idea. Let the player decide which death they want to die ;)

Issue #1225, opened 2025-06-09
Error: S_AliasName: out of sfx_t


Franck6666:
Hi,

after latest git update this error  appaers rejualry all 2 minutes andgame crashes

Error: S_AliasName: out of sfx_t

I googled this error and A say<s this:

is a similar error "S_Findname: out of sfx_t" reported in forums discussing Half-Life and its mods, which suggests an issue with sound effect management in the game.

Regarding the "S_Findname: out of sfx_t" error in Half-Life or similar games, it typically indicates a problem with the game's sound system, possibly due to missing sound files or incorrect configuration.￼3
￼4
 Ensuring that all necessary sound files are correctly placed and referenced can resolve this issue.

Yamagi:
This might be caused by #1216 but I need some more informations to be able to reproduce this. Can you tell me in which level this happens and send the *config.cfg* and and the full terminal output of the game? Under Windows it's written to *My Documents\YamagiQ2\stdout.txt* and under Linux into the terminal the game was started from.

0lvin:
I will create pull request with possible fix.

0lvin:
Could you please try with #1226?

Franck6666:
Hiya,

sry for the delay. before sending  the onfig i tryed  disabling "Ogg music" and it solved the issue. But before  latest git, itwwas enaibled and no probblems appeared like these crashes with inbetween no other messages like i wrote u, appaeered in the linux terminal output.

Hopes this clears up something

Franck6666:
Update, the same Error appaered  again but later with oog music disabled

Yamagi:
> Could you please try with https://github.com/yquake2/yquake2/pull/1226?

I've played through power2.bsp with #1226 applied and didn't encounter the problem.


> On a side note, it seems overly harsh to me to crash the game when it runs out of sfx's. Wouldn't it be enough to just print a warning to the console/stdout and not play the sound?

It should be enough, crashing the game should be the last resort.

BjossiAlfreds:
I made a branch that replaces crashing with console warnings, if you are interested (didn't include the fix from https://github.com/yquake2/yquake2/pull/1226):

https://github.com/BjossiAlfreds/yquake2/tree/sounds-sfx-refactor

Yamagi:
It's now fixed in the `master` branch.

PR #1233, opened 2025-06-16
Sfx allocation prints warnings instead of crashing and fix vulnerability

BjossiAlfreds:
This PR is inspired by issue https://github.com/yquake2/yquake2/issues/1225.

It does 2 things:
* Failure to allocate sfx now prints warnings instead of crashing the game. So the show can go on if this happens
* Fixed a vulnerability where a faulty or malicious server could make clients read memory out of bounds by making a non-player entity play a sexed sound effect. `ent->number` was being used to index the `cl.configstrings` array unguarded.

There is also a bit of refactoring work (adding const, DRY principle, ...) and I added a function `CL_MaxClients` that others might find useful in the future, it returns the current session's max clients, just a wrapper around the `CS_MAXCLIENTS` configstring value.

### DIFF
--- a/src/client/cl_main.c
+++ b/src/client/cl_main.c
@@ -108,6 +108,25 @@ extern cvar_t *allow_download_sounds;
 extern cvar_t *allow_download_maps;
 
 /*
+ * Returns max clients in the current session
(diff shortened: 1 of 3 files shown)


Yamagi:
Thank you :)

PR #1234, opened 2025-06-17
Gyro improvements

ceski-1:
This PR improves and extends gyro functionality for gamepads. This is based on my work for [Woof!](https://github.com/fabiangreffrath/woof/wiki/Gamepad#gyro-options), which was based on the pioneering work by Jibb Smart and Nicolas Lessard for [JoyShockMapper](https://github.com/Electronicks/JoyShockMapper) and [GyroWiki](http://gyrowiki.jibbsmart.com/).

The gyro features follow the de facto standard set by Jibb Smart, popular competitive FPS games, and Steam Input. Briefly, this means gyro space options (local, player, world), tightening, tiered smoothing, threshold-based acceleration, and accurate, real world sensitivity (RWS).

I tested this PR under Windows 11, Debian 12, and RetroPie + Raspberry Pi 3 A+ (for dkms-hid-nintendo support), using DualSense and Switch Pro controllers.

Gyro-specific code is partitioned into `gyro.c` and `gyro.h` under the MIT license. All other changes fall under GPL2, as per usual.

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -701,17 +701,44 @@ it's `+set busywait 0` (setting the `busywait` cvar) and `-portable`
   `+gyroaction` bind to enable, `2` = on with `+gyroaction` to
   disable (default), `3` = always on.
 
-* **gyro_turning_axis**: Sets which gyro axis will be used for turning.
-  The default `0` is "yaw" (turn), for people who prefer to hold their
-  controller flat, like using a pointing device. `1` is "roll" (lean),
-  for people who hold the controller upright, or use a device with the
-  controller attached to the screen, e.g. Steam Deck.
+* **gyro_space**: How motion input is converted into yaw and pitch
+  rotations. `0` = local space, `1` = player space (default),
+  `2` = world space.
 
-* **gyro_tightening**: Threshold of rotation in degrees per second,
-  where gyro inputs below it will be dampened. Meant to counter a
-  noisy gyro and involuntary hand movements. Default `3.5`.
+* **gyro_local_roll**: How gyro roll contributes to turning when using
+  local space. `0` = off, `1` = on (default), `2` = inverted.
 
-* **gyro_calibration_(x/y/z)**: Offset values on each axis of the gyro
+* **gyro_yawsensitivity**: How quickly the view turns left or right. A
+  value of 1.0 means one full rotation of the controller rotates the
+  in-game view by 360 degrees. Default `2.5`.
+
+* **gyro_pitchsensitivity**: How quickly the view looks up or down.
+  Default `2.5`.
+
+* **gyro_tightening**: Threshold of rotation, in degrees per second,
+  below which gyro input is reduced to stabilize aiming. Default `3.0`.
+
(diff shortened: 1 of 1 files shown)


0lvin:
Could you please update Makefile also?

AL2009man:
found a bug with World Space. Tilting the roll axis while on a Yaw position confuses World Space

https://github.com/user-attachments/assets/43e58cd1-e97e-4be1-8818-1443761e0a87

protocultor:
> what do you think?

I think this is a great job! I didn't have the "_world_" problems exposed by AL2009man, but even if they exist, having _player space_ as default is a great choice. Makes sense, since Jibb Smart designed it to overcome the negative points of both _world_ and _local space_:
http://gyrowiki.jibbsmart.com/blog:player-space-gyro-and-alternatives-explained

Only "negative" point I can think of is the default **_smoothing_**, which is too high for my tastes. Ironwail used smoothing alone to overcome the "shaky aim" problem, and it's feeling of lag is not a good one; unfortunately I see it here too (the reason why I just bothered with _tightening_ alone in YQ2). Maybe reducing the default to 2.0 is enough? I kinda feel that, with both at 3.0 degrees per second, the smoothing is triggered before the tightening. In any case, I'm glad it's easily configurable on the new dynamic gyro menu :) .

Code looks pretty good too. Extra points for making it work with Buster on the RPi.

ceski-1:
> Only "negative" point I can think of is the default **_smoothing_**, which is too high for my tastes. Ironwail used smoothing alone to overcome the "shaky aim" problem, and it's feeling of lag is not a good one; unfortunately I see it here too (the reason why I just bothered with _tightening_ alone in YQ2). Maybe reducing the default to 2.0 is enough?

Thank you for the review, I updated the default smoothing [here](https://github.com/yquake2/yquake2/compare/726ebd4c74d901180a78e29d7da982b89de0993e..a71836b894483c551b2a4bb217cd9da56e92c761).

protocultor:
Thank you! Needless to say, I approve of this wholeheartedly.

ceski-1:
[This](https://github.com/yquake2/yquake2/compare/a71836b894483c551b2a4bb217cd9da56e92c761..e8af8fcd70d637808ab98b82a021f17a29480cbd) looks like the correct way to force a menu to refresh while retaining menu layer order. (The menu systems in these games always have some quirks. 🙂)

Yamagi:
> [This](https://github.com/yquake2/yquake2/compare/a71836b894483c551b2a4bb217cd9da56e92c761..e8af8fcd70d637808ab98b82a021f17a29480cbd) looks like the correct way to force a menu to refresh while retaining menu layer order. (The menu systems in these games always have some quirks. 🙂)

Q2s menu is ... interesting. We had so much fun with it over the years. Memory leaks, dangling pointers, unreproduceable crashes, etc. ;)

Is this PR ready or should wait with merging it?

ceski-1:
> Is this PR ready or should wait with merging it?

It's ready for merging now, thanks.

Yamagi:
Thank you! :)

Issue #1236, opened 2025-06-25
`vid_restart` crashes the game sooner or later in Windows


protocultor:
I was doing "something" for GL1 that might require constant `vid_restart`s from the user, and after 2, 3, maybe 5 attempts, the game eventually crashes. Tested my luck again using GL3, and yes, the game crashes all the same if you retry enough times.
In GL3 it falls here, in line 1163:
https://github.com/yquake2/yquake2/blob/master/src/client/refresh/gl3/gl3_main.c#L1163

The reason is that the entity recovered from `gl3_newrefdef.entities` looks OK:
![Image](https://github.com/user-attachments/assets/a6fdd89e-d808-43d3-b03d-0e5afaf77482)

...but the model it points to is not null, but filled with garbage, indeterminate values:
![Image](https://github.com/user-attachments/assets/437e6630-3ece-4a37-ace6-1b42398e34f2)

I don't even know if I'm seeing wrong and the entity was "lost" in the first place (doesn't look like it), but the issue is that some pointers get eventually lost if you `vid_restart` hard enough, and I'm failing to see the reason, since it looks stable the first few restarts. As this deals with entities present, when you are disconnected in the console you can restart all you want and it will never fail. The crashes happen when you "_apply_" in the video menu while a demo is playing.

Interesting facts:
- Unfortunately I no longer have the means to test other OSs, but it seems this ONLY happens in Windows. I say this because the RPi is a lot more stable (crashes because of my badly built SDL2 version instead).
- This happens since YQ2 version `8.00-RC1`. The previous version, `7.45a`, is rock solid when downloaded from `deponie.yamagi.org`, but when I compile it, the same problem happens. Note that you must change "vertical sync" in the video menu for "apply" to work in 7.45a and lower.

This makes me think that maybe some optimization in `gcc` or any other `MSYS2` component that appeared in the last 5 years may have something to do with it; compiling with `-O0` gives the same result though.

Just posting because I don't know what to do with this info, and I'm not carrying it to the grave.
I hope this is not seen as a disadvantage for what I am doing (spoiler: gamma correction for GL1 with SDL3).

Yamagi:
Nice. 😬 It's also fairly reproduceable on (my?) Linux. Changing the renderer, e.g. gl1 -> soft -> gl1 -> soft -> ..., seems to increase the chance of a crash on Linux. I'm able to crash the game when building with `gcc` or with `clang`. Maybe some kind of optimization or something implemented by both compilers.

[...]

PR #1237, opened 2025-06-27
Gamma correction for GL1 with SDL3

protocultor:
Resurrected the Vanilla Q2 gamma table. Thanks to the fact that is still applied to textures today (with `gamma value == 1.0`), getting it back was easy. The difference with the old implementation is that _alpha surfaces_ and `gl_polyblend` effects take gamma into account for their color, and most importantly, the gamma table use is no longer limited to loading textures. Now it is used for picking color intensity in lightmaps, meshes, particles, beams, and even cinematics.

It's difficult to tell apart the end result from the SDL2 gamma functions IMHO. What's noticeable is the lack of color banding at extreme values, and of course, that gamma no longer affects the whole screen, just the game window :)

The great disadvantage this has is that requires `vid_restart` to apply the new `vid_gamma` value, with the problem described in #1236.
EDIT: #1238 solves the crashes, with it the game can re-apply gamma with no drawbacks.

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -529,6 +529,11 @@ it's `+set busywait 0` (setting the `busywait` cvar) and `-portable`
   "flash" you see when getting injured or picking up an item. In GL1 is
   also used for looking underwater. Default is `1` (enabled).
 
+* **gl_znear**: Sets the distance to the *near depth clipping plane* of
+  the player view. Reducing it may allow some weapon animations to not
+  get "clipped" by the player view (e.g. railgun firing), at the risk
+  of heavy glitches with some hardware configurations. Default is `4`.
+
 * **gl_texturemode**: How textures are filtered.
 
   - `GL_NEAREST`: No filtering (using value of *nearest* source pixel),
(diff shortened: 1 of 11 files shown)


protocultor:
Recently included is the fix for `zNear`, produced by (and talked about in) #1108 . Default is the classic `4`, the user can change it to whatever they want.

> A nice and clean solution to the problem. I have played through the first unit and fiddled a little bit with gamma values. Looks good and as you said, it's really next to indistinguishable from the SDL2 gamma tables. May I ask how many hours it took to implement it?

Compared to the amount of time I spent "investigating the capabilities of fixed-function OpenGL" in this regard, it took next to nothing.
I should have started looking at the existing gamma table first. After I understood how it worked, between looking for effects unaffected by gamma, searching for them in code and changing them, maybe 10 hours at max.

Yamagi:
> Compared to the amount of time I spent "investigating the capabilities of fixed-function OpenGL" in this regard, it took next to nothing.
I should have started looking at the existing gamma table first. After I understood how it worked, between looking for effects unaffected by gamma, searching for them in code and changing them, maybe 10 hours at max.

Still a lot of work and time investment, when counting the investigation before even more. The (Yamagi) Quake II community owns you more than one beer for your work.

PR #1238, opened 2025-06-28
Temp ent refactoring and temp ent model state cleared on renderer restarts

BjossiAlfreds:
The following was done:
* Fixes issue https://github.com/yquake2/yquake2/issues/1236.
* Temp ent static state is now cleared on renderer restarts/changes
* Added `static` to a few private functions in the temp ent code file
* Created utility functions to simply the beams/sustains code (DRY principle)
* `CL_ClearTEnt` now also `NULL`-ifies the `cl_mod_` and `cl_sfx_` static vars
* Adds a `r < 0` check for indexing `splash_color` array since `MSG_ReadByte` can return `-1`
* Temp ent global variables now `static`
* `CL_RandomFootstepSfx` and `CL_PowerScreenModel` accessor  functions, since they were needed elsewhere (exposing these variables directly is a bad idea I think)
* Renamed `playerbeam` variables and functions to the more correct `heatbeam` and cleaned up `CL_ParseHeatBeam` and `CL_AddHeatBeams`
* Removed unused `cl_sfx_plasexpl` variable
* Removed unused `cl_mod_monster_heatbeam` variable. Was never actually being used, `CL_ParseHeatBeam` always set `model` to `cl_mod_heatbeam` if this model was passed to it
* Potential fix for heatbeam appearing in the wrong location for left/center weapon modes
* Replaced `return` with `continue` in `CL_AddBeams` loop `cl_mod_lightning` special case. I see no reason why it should stop adding beams if this happens

I don't know if this fixes the `vid_restart` crashes for others but it does for me.

Because of my refactoring, if others could test the temp entities (beams and steam namely) is a good idea.

### DIFF
--- a/src/client/cl_tempentities.c
+++ b/src/client/cl_tempentities.c
@@ -116,7 +116,6 @@ static struct model_s *cl_mod_plasmaexplo;
 
 static struct model_s *cl_mod_lightning;
 static struct model_s *cl_mod_heatbeam;
-static struct model_s *cl_mod_monster_heatbeam;
 static struct model_s *cl_mod_explo4_big;
 
 /*
@@ -292,7 +291,6 @@ CL_RegisterTEntModels(void)
 	cl_mod_explo4_big = R_RegisterModel("models/objects/r_explode2/tris.md2");
 	cl_mod_lightning = R_RegisterModel("models/proj/lightning/tris.md2");
 	cl_mod_heatbeam = R_RegisterModel("models/proj/beam/tris.md2");
-	cl_mod_monster_heatbeam = R_RegisterModel("models/proj/widowbeam/tris.md2");
 }
 
 /*
@@ -340,7 +338,6 @@ CL_ClearTEntModelVars(void)
 
 	cl_mod_lightning = NULL;
 	cl_mod_heatbeam = NULL;
-	cl_mod_monster_heatbeam = NULL;
 	cl_mod_explo4_big = NULL;
 }
 
@@ -458,10 +455,12 @@ CL_ParseBeam(struct model_s *model, qboolean with_offset)
  * adds to the cl_playerbeam array instead of the cl_beams array
  */
 static void
-CL_ParsePlayerBeam(struct model_s *model)
+CL_ParseHeatBeam(qboolean is_monster)
 {
+	static const vec3_t plofs = {2, 7, -3};
+	const vec_t *ofs;
 	int ent;
-	vec3_t start, end, offset;
+	vec3_t start, end;
 	beam_t *b;
 	int tm;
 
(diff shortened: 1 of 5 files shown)


protocultor:
_Xatrix's_ first and third demo (`xdemo1.dm2` and `xdemo3.dm2`) have beams, and they work fine while restarting / switching renderers. From the other demos, at least enemy fire also work fine (or at least, they don't manage to crash the game).

BjossiAlfreds:
> _Xatrix's_ first and third demo (`xdemo1.dm2` and `xdemo3.dm2`) have beams, and they work fine while restarting / switching renderers. From the other demos, at least enemy fire also work fine (or at least, they don't manage to crash the game).
> 
> This thing feels more stable than vanilla Q2. Great job.

Good to know! `CL_ClearTEntModels` memsets the `cl_beams` and `cl_playerbeams` arrays so there won't be any lingering model pointers in there when the renderer resets/changes.

src/client/cl_tempentities.c
```diff
@@ -389,95 +486,78 @@ CL_ParsePlayerBeam(struct model_s *model)
 		MSG_ReadPos(&net_message, offset);
 	}
 
+	if (!model)
```
protocultor, line 489: I feel this validation should be above line 474. Just to "keep order", nothing actually terrible.

BjossiAlfreds: I see what you mean but I put it below the last `MSG_Read` call so that the client message buffer doesn't go out of sync.

BjossiAlfreds: Hmm, this is actually worth looking at more closely, because if either of the static vars are NULL, it could desync the message buffer. I wonder if there is a way to make this deterministic.

BjossiAlfreds: I made some interesting discoveries:
* `CL_ParsePlayerBeam` was only used for `TE_HEATBEAM` and `TE_MONSTER_HEATBEAM`, so I renamed it to `CL_ParseHeatBeam`
* The `MSG_ReadByte` in the else clause was never actually reachable because `model` was always either `cl_mod_heatbeam` or `cl_mod_monster_heatbeam`. So I was able to simplify the code

See what you think of the changes in the new commit I just made.

BjossiAlfreds: The `model = cl_mod_heatbeam;` line to override `cl_mod_monster_heatbeam` looks intentional. Maybe the widow beam model is broken or code for it was never finished. `CL_AddPlayerBeams` only checks for `cl_mod_heatbeam` so even if we restored the widow beam it wouldn't do anything without quite a bit of effort.

Should I just remove the registration for the widow beam entirely?

BjossiAlfreds:
I just did a commit. I "tricked" the renderer into mirroring the beam to the left-side when left-handed mode is used, while still keeping the `RF_WEAPONMODEL`  flag. Hopefully it keeps the FOV fix.

I also made the flag not get applied for center weapon, as the beam was not being rendered at all that way.

src/client/cl_tempentities.c
```diff
@@ -1377,318 +1419,207 @@ CL_AddBeams(void)
...
-			hand_multiplier = -1;
-		}
-
-		else
-		{
-			hand_multiplier = 1;
-		}
-	}
-	else
+	switch (hand_pos)
```
protocultor, line 1516: I'm too old and dumb to understand this code, but since this is the only usage of `HandPos()`, just to set `hand_mul`, shouldn't `HandPos()` itselt return the correct value for `hand_mul` directly? The `switch()` feels unnecessary, and you can even return an `int` if the data type is important.

In other topic, with `hand 1` (left), the plasma beam shoots wherever, ahead of where you are pointing with the mouse; something like a Wii shooter. `0` and `2` work just fine.

protocultor: Yeah, it looks pretty good now, with no apparent "misbehavior". Thanks for your work!

Yamagi:
Step 4 is the one I was missing... Thank you for doing my work with finishing the investigation and implementing a fix :)

I have played through some maps fix focus on the heatbeam and found no obvious problems introduced by the refactorings and the heatbeam changes. I'll merge it, let's see if some corner cases come up. Good job!

PR #1241, opened 2025-07-12
OpenAL double free fix when unloading the library.

devnexen:
Since it is preferable not to touch legacy YQ2 api, creating a separate call just for OpenAL alone, making sense only on unixes though.

```c
Block was alloc'd at
==16821==    at 0x4844F93: operator new(unsigned long) (vg_replace_malloc.c:487)
==16821==    by 0x5FAAE0A: ??? (in /usr/lib/x86_64-linux-gnu/libopenal.so.1.24.2)
==16821==    by 0x6CDA78E: ??? (in /usr/lib/x86_64-linux-gnu/pipewire-0.3/libpipewire-module-protocol-native.so)
==16821==    by 0x6CCB670: ??? (in /usr/lib/x86_64-linux-gnu/pipewire-0.3/libpipewire-module-protocol-native.so)
==16821==    by 0x6CCBE07: ??? (in /usr/lib/x86_64-linux-gnu/pipewire-0.3/libpipewire-module-protocol-native.so)
==16821==    by 0x6491C45: ??? (in /usr/lib/x86_64-linux-gnu/spa-0.2/support/libspa-support.so)
==16821==    by 0x644AB54: ??? (in /usr/lib/x86_64-linux-gnu/libpipewire-0.3.so.0.1402.0)
==16821==    by 0x4BE5B7A: start_thread (pthread_create.c:448)
==16821==    by 0x4C635EF: clone (clone.S:100)
...
Address 0x5db1f90 is 0 bytes inside a block of size 384 free'd
==16821==    at 0x484886D: operator delete(void*, unsigned long) (vg_replace_malloc.c:1181)
==16821==    by 0x4B94D50: __cxa_finalize (cxa_finalize.c:97)
==16821==    by 0x5F3ADB6: ??? (in /usr/lib/x86_64-linux-gnu/libopenal.so.1.24.2)
==16821==    by 0x4001FC1: _dl_call_fini (dl-call_fini.c:43)
==16821==    by 0x4002425: _dl_catch_exception (dl-catch.c:215)
==16821==    by 0x4002B1C: _dl_close_worker (dl-close.c:264)
==16821==    by 0x4003405: _dl_close (dl-close.c:795)
==16821==    by 0x4002398: _dl_catch_exception (dl-catch.c:241)
==16821==    by 0x40024BE: _dl_catch_error (dl-catch.c:260)
==16821==    by 0x4BE1D66: _dlerror_run (dlerror.c:138)
==16821==    by 0x4BE1A95: dlclose@@GLIBC_2.34 (dlclose.c:31)
==16821==    by 0x1C87F2: Sys_FreeLibrary (system.c:645)
```

### DIFF
--- a/src/backends/unix/system.c
+++ b/src/backends/unix/system.c
@@ -690,6 +690,43 @@ Sys_LoadLibrary(const char *path, const char *sym, void **handle)
 	return entry;
 }
 
+void *
+Sys_LoadLibraryNoUnLoad(const char *path, const char *sym, void **handle)
+{
+	void *module, *entry;
+
+	*handle = NULL;
+	module = dlopen(path, RTLD_LAZY | RTLD_NODELETE);
+
+	if (!module)
+	{
+		Com_Printf("%s failed: %s\n", __func__, dlerror());
+		return NULL;
+	}
+
+	if (sym)
+	{
+		entry = dlsym(module, sym);
+
+		if (!entry)
+		{
+			Com_Printf("%s failed: %s\n", __func__, dlerror());
+			dlclose(module);
+			return NULL;
+		}
+	}
+	else
+	{
+		entry = NULL;
+	}
+
+	Com_DPrintf("%s succeeded: %s\n", __func__, path);
+
+	*handle = module;
+
+	return entry;
+}
+
 /* ================================================================ */
 
 void
--- a/src/backends/windows/system.c
+++ b/src/backends/windows/system.c
@@ -698,6 +698,12 @@ Sys_LoadLibrary(const char *path, const char *sym, void **handle)
 	return entry;
 }
 
+void *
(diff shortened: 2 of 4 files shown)


Yamagi:
This is not the right way to fix these leaks. Setting `RTLD_NODELETE` prevents linker objects (e.g. libraries) getting deleted from memory after being loaded once. That masks leaks in  libraries because they never get deleted from memory, it does not really solves them. And it introduces new bugs. Most libraries expect to get reloaded into clean memory which is not the case with `RTLD_NODELETE` set. And when the library isn't needed anymore it gets "leaked" by being kept in memory despite being unloaded.

If these are are really critical leaks and not just cosmetics they need to be fixed in the upstream project, e.g. openal-soft or even pipewire.

PR #1243, opened 2025-07-23
Fix for sofware rendering in SDL3

protocultor:
As discussed in  https://github.com/yquake2/yquake2/issues/1097#issuecomment-3092428582 , using `sw_partialrefresh` with SDL3 is counterproductive, with Windows being the most affected (with Linux in a lesser degree), so this cvar is disabled by default when building for SDL3 now.
Also included:
- New cvar `joy_trigger`, which specifies pressure on the gamepad triggers to register them as a button press. This threshold was hardcoded until now.
- For SDL3 gamepad input, made use of `SDL_GetGamepadButtonLabel()`, which is a more convenient way to extract labels for the currently connected gamepad. Allows for a more practical and future-proof way to detect styles of "labels" and "confirm" for gamepads, removing the need for a continuous update of the list of gamepad types (which SDL2 will still be using).
- GL1 error messages rewritten.

### DIFF
--- a/src/client/refresh/soft/sw_main.c
+++ b/src/client/refresh/soft/sw_main.c
@@ -392,7 +392,7 @@ R_RegisterVariables (void)
 	// On MacOS texture is cleaned up after render and code have to copy a whole
 	// screen to texture, other platforms save previous texture content and can be
 	// copied only changed parts
-#if defined(__APPLE__)
+#if defined(__APPLE__) || defined(USE_SDL3)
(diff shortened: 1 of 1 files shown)


0lvin:
Let's disable partial refresh for now, I will check  more correct way to apply partial refresh and create separate pull request.

0lvin:
When it was added it was +5% fps improvement as I remember.

PR #1244, opened 2025-07-25
Backport code from yquake2remaster 

0lvin:
Fix random issues:
* Transparent texture update in console(black color had previous frame artifacts)
* Implement shared printf functions in separate file reused in renders, and add implementation of client side printf for reuse render code in render code if need.
* Update soft partial refresh logic
* Apply coverity fixes to savegame code
* Add support 24bit color cinema code if `cl_showfps=0`


protocultor:
Good changes! Your last 2 commits clash with my PR though.

0lvin:
Could you please check?

I have more coverity fixes applied to yquake2remaster, but I think for current PR it's already big amount of changes.

Do someone like to have mp3 sound effects and mpeg cinema support? It also could be ported as separate PR.

Performance difference:
* sdl2 + partial refresh: 57.4 fps
* sdl2 + no partial refresh: 56.8 fps
* sdl3 + partial refresh: 49.9 fps
* sdl3 + no partial refresh: 48.9 fps

> Since you are consting the vector functions, maybe now would be a good time to turn `DotProduct`, `VectorAdd`, `VectorSubtract`, `VectorCopy` and `VectorClear` into functions?

Let it be separate PR. Do anyone have performance numbers difference between macros vs function for such case?

Yamagi:
Is this ready for merge?

0lvin:
Yes :-)

Yamagi:
Thank you!

PR #1247, opened 2025-09-01
workflows: update workflows

0lvin:
* Fix curl 32 bit download,
* Fix workflow run for forks with custom name
* Enable CodeQL and PREfast validation
* Use github-cli instead wget


protocultor:
Just FYI, SDL3 `3.2.22` just released. Annoying, I know :)

PR #1250, opened 2025-09-17
build: Don't hard-code a specific pkg-config executable

smcv:
When cross-compiling we'll need to use a pkgconf or pkg-config suitable for the host architecture, for example aarch64-linux-gnu-pkgconf. OS distributions typically expect to do this by setting the environment variable PKG_CONFIG, as used in Autotools.


Yamagi:
Thank you :)

PR #1251, opened 2025-09-20
Make SDL3 the default

Yamagi:
This pull request is mostly to test the CI workflows.

Done is:
* Switch the default in the Makefile and CMakeLists.txt.
* CI workflows were switched to SDL3. Ubuntu and Win MSVC cannot be switched right now.

Still missing is:
* Look through the documentation.
* Switch the optional renderers to SDL3.


Yamagi:
I'll see what we can do with Win32. Maybe something with the official SDL 3 developer download for Windows?

I haven't looked into CodeQL so far. A little bit of help would be nice :)

0lvin:
Looks as no simple solution to run codeql with SDL3, Ubuntu-latest is 24.04, so no SDL3 for now. I will create PR with disable SDL3 before validation, looks like it works

.github/workflows/win32.yml
```diff
@@ -68,14 +68,16 @@ jobs:
         cp doc/070_packaging.md publish/quake2-win32-${{github.sha}}/misc/docs/070_packaging.txt
         cp doc/080_contributing.md publish/quake2-win32-${{github.sha}}/misc/docs/080_contributing.txt
         cp doc/090_filelists.md publish/quake2-win32-${{github.sha}}/misc/docs/090_filelists.md
-        # SDL2
-        wget -c https://github.com/libsdl-org/SDL/releases/download/release-2.32.8/SDL2-2.32.8-win32-x86.zip
-        unzip -o SDL2-2.32.8-win32-x86.zip
-        cp SDL2.dll publish/quake2-win32-${{github.sha}}/
+        # SDL3
+        gh release download release-3.2.22 \
```
0lvin, line 72: Unfortunately GitHub has not provided 32 bit version of their tool, so only wget/curl solution possible.

0lvin:
I have created #1252 with force usage of SDL2 and cherrypicked several SDL2 changes from this PR.

.github/workflows/win32.yml
Yamagi: Great.

Yamagi:
Another round of changes:

* Made MacOS more robust, don't assume a specific SDL3 version.
* Merged Linux workflow changes from @protocultor.
* Removed .swp file.

The setup-sdl action hits the problem reported in https://github.com/libsdl-org/setup-sdl/issues/35 I have added our failure as another comment.

PR #1252, opened 2025-09-21
Force usage of SDL2 in codeql

0lvin:


0lvin:
I have splitted pull requests to two, this only only for codeql with SDL3 build, looks like codeql does not like manual run, so left auto build, with hack to replace SDL2 as default.

PR #1254, opened 2025-09-21
Fix prefast warings

0lvin:
Fix prefast fairings and remove maps auto complete limit to 1024.


Yamagi:
I knew, I said "tomorrow" and than life got into the way. It will take a few days more...

PR #1263, opened 2025-11-13
Progressive centerview

protocultor:
_Quake I_'s `centerview` did not operate like _Quake II_'s, which snaps you to the center immediately. Instead, _Q1_ slowly returns your view to the center in a progressive way. This PR was inspired by it, but the actual implementation is instead based on [how Flick Stick currently operates in YQ2](https://github.com/yquake2/yquake2/blob/QUAKE2_8_60/src/client/input/sdl3.c#L1506).
Time it takes to center is controlled by the new cvar `cl_centertime`; setting this to `0` restores the vanilla behaviour.

Also included:
- Heavy refactor on `cl_input.c`, `client.h` and GL1 header `local.h`.
- `gl1_flashblend` effects now use SDL3 gamma properly (not that anyone cared...).
- #1220 still appeared, in very rare occasions, when using `gl1_lightmapcopies 1`. Now it's fixed.
- Deleted forward and side sensitivities in gamepad menu; they weren't very useful. Replaced by documentation.
- Flick Stick: "snap-to-axis" is enabled only at the very start of a flick. Avoids "shaken up" player view at a flick's end.
- Low frequency gamepad rumble is dampened when using gyro. Also avoids to shake the player view because of vibration.

### DIFF
--- a/src/client/input/sdl2.c
+++ b/src/client/input/sdl2.c
@@ -217,7 +217,7 @@ static cvar_t *gyro_accel_multiplier;
 static cvar_t *gyro_accel_lower_thresh;
 static cvar_t *gyro_accel_upper_thresh;
 
-// Gyro is being used in this very moment
+// Gyro is ready to take player inputs
 static qboolean gyro_active = false;
 
 // Controller is connected and supports gyro, gyro is not disabled, game is not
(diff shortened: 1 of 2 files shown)


Yamagi:
> Quake I's centerview did not operate like Quake II's, which snaps you to the center immediately. Instead, Q1 slowly returns your view to the center in a progressive way. This PR was inspired by it, but the actual implementation is instead based on [how Flick Stick currently operates in YQ2](https://github.com/yquake2/yquake2/blob/QUAKE2_8_60/src/client/input/sdl3.c#L1506).
Time it takes to center is controlled by the new cvar cl_centertime; setting this to 0 restores the vanilla behaviour.

I've never played with `centerview`, but I'm no controller player either. I think players using controller will like it.

> Also included

More code cleanup is always good.

> gl1_flashblend effects now use SDL3 gamma properly (not that anyone cared...).

or no one noticed, didn't bother to report, etc...

Thank you :)

PR #1266, opened 2025-11-22
Makefile QUIET and WERR variables and separate folders for release and debug binaries

BjossiAlfreds:
This PR adds the following features to the build system:
* Some improvements to variable docs at the top of the file
* An optional `QUIET` variable that, when defined, silences the `===> CC ...` output lines.
* An optional `WERR` variable that, when defined, adds `-Werror` to `CFLAGS` to treat warnings as errors.
* Build and bin directories no longer hard-coded:
- `BINROOT`: The root dir for binaries (none by default)
- `BINDIR`: Root + build type. Example: `release` or `debug`
- `BUILDROOT`: Root dir for build objects (`build/` by default)
- `BUILDDIR`: Build root + build type. Example: `build/release` or `build/debug`
These variables can't be customized right now but we could maybe allow that in the future.
* Debug and release builds are now put in separate folders
- Release: `release/*` and `build/release/*`
- Debug: `debug/*` and `build/debug/*`

Let me know what you think of the debug/release dir stuff. If this is not wanted I can undo that commit and maybe do a separate PR for it with more customization.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /build/
+/debug/
 /release/
 *.mk
 *.user
--- a/Makefile
+++ b/Makefile
@@ -187,8 +187,13 @@ endif
 override BINROOT :=
 override BUILDROOT := build
 
-override BINDIR := $(BINROOT)release
-override BUILDDIR := $(BUILDROOT)
+ifdef DEBUG
+	override BINDIR := $(BINROOT)debug
+	override BUILDDIR := $(BUILDROOT)/debug
+else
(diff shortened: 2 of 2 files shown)


BjossiAlfreds:
I discovered a small bug where the `QUIET` variable is not applied from the config file. I accidentally put the `# Normalize QUIET value to either "x" or ""` block before the config include.

Should I do a hotfix PR or maybe you can make a quick fix commit where you move the config include above the normalize block?

PR #1268, opened 2025-12-03
build: Fixed QUIET var from config.mk being ignored

BjossiAlfreds:
A small hotfix PR to fix `QUIET` variable being ignored if defined in a config file. Accidentally put the variable check before the config file is read.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,10 @@
 #   i686-w64-mingw32   -> indicates i386
 # QUIET
 #   If defined, "===> CC ..." lines are silenced.
+# SOURCE_DATE_EPOCH
+#   For reproduceable builds, look here for details:
+#   https://reproducible-builds.org/specs/source-date-epoch/
+#   If set, adds a BUILD_DATE define to CFLAGS.
 # UBSAN
 #   Builds with undefined behavior sanitizer.includes DEBUG.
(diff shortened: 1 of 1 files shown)


BjossiAlfreds:
By the way, is there a reason why we aren't using the `-s` flag when linking? The binaries contain a lot of unused data that is only there to take up space.

Yamagi:
Thank you for the bug fix. I would have done it myself but you were faster :)

BjossiAlfreds:
> > By the way, is there a reason why we aren't using the -s flag when linking? The binaries contain a lot of unused data that is only there to take up space.
> 
> Because `execinfo()` needs the symbol table. However the game is nearly crashfree these days, most Linux distributions strip the binaries and `execinfo()` doesn't really work on Windows. Maybe it's time to change the crash handler to print the backtrace only when the game is build with `DEBUG`.

Oh right. I was under the impression `-s` only stripped unreferenced stuff so wouldn't impact debug symbols. What we could do is use a separate variable, like:

```
ifndef DEBUG
ifdef STRIP_SYMBOLS
override LDFLAGS += -s
endif
endif
```

Yamagi:
Sounds good and would help distributors who want to strip the symbols. I can implement int but it might take a few days.

PR #1277, opened 2025-12-24
Gamepad selection

protocultor:
Expanded the `in_initjoy` cvar, now it allows the user to select the used gamepad for the game.
E.g. the user has 2 gamepads connected; by running the game with `./quake2 +set in_initjoy 2` the game uses the second gamepad instead of the first detected one.
Note that this is still only a "preference", meaning the game will still skip to the next available joystick (or previous ones) if it couldn't init the one selected.

Also, fixed compilation with `curl` disabled (`WITH_CURL=no`) on Linux.

These are small things needed by [a certain other project](https://github.com/batocera-linux/batocera.linux/pull/14946). Happy holidays!


Yamagi:
Thank you :)

PR #1279, opened 2026-01-03
Infinite loop fix

protocultor:
Added in my previous PR, I wasn't careful with the special cases that required a `continue` (error handling, mainly) when changed a `for` loop to a `do`/`while`.
Additionally, refactored some haptic code to avoid repetition.


0lvin:
What do you think about split more IN_Controller_Init to several function, looks like it has 300+ lines (2540..2781) of code already?

PR #1289, opened 2026-02-14
Add cppcheck to static analysis checks

0lvin:
Add cppcheck validation and fix crashes with rerelease maps when command buffer is reinited inside execution command.

### DIFF
--- a/.github/workflows/linux_aarch64.yml
+++ b/.github/workflows/linux_aarch64.yml
@@ -19,6 +19,7 @@ jobs:
     permissions:
       actions: read
       contents: read
+      security-events: write
     strategy:
       fail-fast: false
       matrix:
@@ -29,7 +30,7 @@ jobs:
       run: |
         sudo apt update
         sudo apt install libgl1-mesa-dev libsdl2-dev libopenal-dev libcurl4-openssl-dev \
-            build-essential
+            build-essential snapd
     - name: Check out repository code
       uses: actions/checkout@v6
     - name: Build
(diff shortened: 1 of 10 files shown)


0lvin:
Currently mostly for new pull request checks, without fixes of detected issues.

Also port fix from yquake2remaster for rewrite command buffer in change level command.

0lvin:
I have created pull requests for all main modules. 

All have updates for workflows and shared code and interface, except zaero. What should we do with zaero? I have updated workflows, but shared code and interface we have under gpl licence in other mods, should we ask everyone contributed to shared and changed interfaces in game.h to provide accept to use this code and update zaero?

I am ok to use my contributions to other mods and update zaero code, but what about everyone else? What will be licence of zaero after that, should we put gpl header to files that are same with other mods?

Yamagi:
> I am ok to use my contributions to other mods and update zaero code, but what about everyone else? What will be licence of zaero after that, should we put gpl header to files that are same with other mods?

The old problem with the SDL license and the GPL... In general the SDK license and the GPL are completely incompatible. We can't merge GPL licensed code into the SDL licensed mods and the other way round. And, while it might be some kind of grey zone, we shouldn't link GPL licensed code with SDL licensed code and compile time.

I once considered a pinned issue here in the main repository were all contributors can give a global okay that they are okay with their code contributions getting relicensed to the SDK license. And of course revoke that permission for future contributions and any time. But then we could relicense xatrix and rogue to GPL and the problem vanished. Maybe such approach would still be an option?

0lvin:
Looks like we have workaround for porting changes legally, before re-license xatrix/rogue to GPL, it had same non-gpl license, as result all changes from that mods made before applying GPL license in theory could be applied to zaero. (https://github.com/yquake2/xatrix/issues/96)

What do you think? Have I missed something?

Yamagi:
That should be okay. xatrix before commit 0ee9b324a1d928b0cf819eba78ec657d053dfcb4 and rogue before commit db15317c900946d6e309a6d63b3383a85f908a9a were under SDK license and all changes until that point can be merged into other SDL licensed projects like zaero.

PR #1299, opened 2026-03-15
Savegame stability improvements

BjossiAlfreds:
This PR improves the stability of the savegame system.

* `sg_fread` and `sg_fwrite` wrapper functions to uniformly handle read/write errors
* Prevented indexing `g_edicts`, `game.clients` and/or `itemlist` out of bounds if the save file contains illegal index number
* Added missing `fclose` calls before a few `gi.error` calls
* Reduced funcStr buffer size from 2048 to 128 bytes, as 2048 is excessive
* Ensure string in funcStr buffer is null terminated
* Ensure strings in save header are null terminated
* A bit of cleanup in save compatibility check code
* Handle negative string length read from save
* Changed `if (*(edict_t **)p == NULL)` to `if (*(gitem_t **)p == NULL)` when checking if item pointer is NULL
* Use `ITEM_INDEX` macro to calculate item index
* Use `GetItemByIndex` when restoring item pointer from save
* Made `GetItemByIndex` return `NULL` for negative index
* Sanitize entity fields `inuse`, `s.number` and `classname` after reading the entity data from save file
* Sanitize `level` struct after reading from save file: inline strings, `pic_health` and `body_que` fields
* Sanitize `game` struct after reading from save file: inline strings
* Replaced `game.num_items` with a global variable `itemlist_len`. Storing num items in save files is unnecessary and invites security and stability risk. The `num_items` member is still in the struct for backwards compatibility
* Refactored item selection code for stability and security: Redesigned `SelectNextItem` `SelectPrevItem` `ValidateSelectedItem` and new `G_SetStats_SelectedItem` function
* Sanitize client structs after reading from save file: inline strings, `ammo_index` `flood_whenhead` and selected item
* Use `itemlist_len` instead of `MAX_ITEMS` in weapon selection cod
* The game can now save/load with missing mmoves/functions, printing a warning instead of erroring to console
* Use `ARREND` macro in field loops and removed blank last array entries
* Renamed `fields` array to `entfields`
* Field arrays and pointers now `const`
* `FindSpawnfield` function and `entfields` is now `static`
* Refactored `ED_ParseField` for simplicity and stability
* Clear `clients` pointer, `maxclients`, `maxentities` and `num_items` for saves as well as clearing them on load
* Copy `game` to a temp struct to avoid having to modify `game` before writing it to a save
* `WriteSaveHeader` and `WriteGameLocals` functions for clarity and optimizing stack allocation
* Ensure configstrings are null terminated after reading from save file

Also snuck in a small optimization in `Cvar_WriteVariables`. Noticed it was using an unnecessary temporary buffer, also removed a redundant call to `fflush` before `fclose`.


src/game/savegame/savegame.c
```diff
@@ -181,6 +181,26 @@ static field_t clientfields[] = {
 
 /* ========================================================= */
 
+static void
+sg_fread(void *dest, size_t n, FILE *f)
+{
+	if (!fread(dest, n, 1, f))
```
0lvin: I have rebased your changes over remaster code and coverity detected issue with change, here should be checked that function returns exactly n bytes read. Could you please check?
```
** CID 645076:       Error handling issues  (CHECKED_RETURN)
/src/game/savegame/savegame.c: 190           in sg_fread()


_____________________________________________________________________________________________
*** CID 645076:         Error handling issues  (CHECKED_RETURN)
/src/game/savegame/savegame.c: 190             in sg_fread()
184     
185     /* ========================================================= */
186     
187     static void
188     sg_fread(void *dest, size_t n, FILE *f)
189     {
>>>     CID 645076:         Error handling issues  (CHECKED_RETURN)
>>>     "fread(void * restrict, size_t, size_t, FILE * restrict)" returns the number of bytes read, but it is ignored.
190     	if (!fread(dest, n, 1, f))
191     	{
192     		fclose(f);
193     		gi.error("Error reading " YQ2_COM_PRIdS " bytes from save file", n);
194     	}
195     }
```

Issue #973, opened 2023-02-05
OpenAL HRTF sometimes sounds off


BielBdeLuna:
sometimes it sounds as if the virtual microphones lag behind the player, it can be heard with the EFX reverbs and without, so it's not dependant on them

BielBdeLuna:
it sounds sometimes like a low-pass filter, maybe it's the occlusion hurting the HRTF?

BielBdeLuna:
this can be heard in Quake 2 first demo, when the player throws the first grenade at the start: 

if you `toggle s_occlusion_strength` 1 and 0 at that same moment the player is running backwards as he throws the grenade you should hear the difference.

the doppler effect doesn't make any substantial change to the HRTF sound image.

Yamagi:
> this can be heard in Quake 2 first demo, when the player throws the first grenade at the start

In what demo? In *demo1.dm2* the player doesn't throw a grenade...

BielBdeLuna:
sorry I was using yquake with the quake 2 RTX where they changed the first demo, and a grenade is tossed there

Yamagi:
I can hear it.  I wonder if this really a code problem and not a unwanted side effect of Q2s rather basic sound system providing not enough data to OpenAL for 100% correct occulusion calculations. @0lvin, what do you think?

0lvin:
I will check, I suppose that issue related to incorrect combination of parameters for openal processing.

 Have you reproduced issue under Windows and Linux?

Yamagi:
Only under Linux

BielBdeLuna:
I think this is the combination of reverb and the occlusion, like if the occlusion where too great effect given the reverb and HRTF channel heavy nature, maybe is a question of switching the order the effects get to be applied?

pepelevamp:
i'm on windows and i found the doppler causes comb filtering in the environmental sounds.

the occlusion causes a hard pop when walking past bulk heads. i made a separate ticket for the occlusion one which i kinda regret after seeing this issue open here.

Yamagi:
Should be fixed by #1300

PR #1300, opened 2026-03-19
sound: use dedicated occlusion filter separate from underwater

devnexen:
The occlusion path was reusing the underwater lowpass filter which has an aggressive AL_LOWPASS_GAINHF (0.25) that strips the high frequency cues HRTF relies on for spatial positioning.

Introduce a separate occlusion filter with distance-scaled high frequency attenuation, preserving enough HF content for accurate 3D audio.

Ref yquake2/yquake2#973


Yamagi:
Thank you :)

PR #1301, opened 2026-03-23
defensive programming: harden various code paths

devnexen:
- game: add missing NULL owner check in Touch_DoorTrigger
- client: validate entity number from network in sound parsing
- server: use Q_strlcpy in SV_StatusString


Yamagi:
Thank you :)

PR #1302, opened 2026-04-06
server: OPTIMIZE_CSBASE flag for sv_optimize_sp/mp_loadtime cvars

BjossiAlfreds:
This PR adds new `OPTIMIZE_CSBASE` flag (`16`) for the `sv_optimize_sp/mp_loadtime` cvars and sets the sp default value to `31`.

The last configstrings packet is usually only partially filled with data. With this optimization enabled, the server will, if possible, use that leftover space for a first batch of entity baselines. In some levels this will shave off one roundtrip of client-server packets. Only a trivial optimization for single-player but may be more noticeable in multiplayer due to the added networking latenncy.

I was also thinking, should I increase the default value of `sv_optimize_mp_loadtime` to 7? It's been a while now and I feel this should be safe. Does anyone here have experience with playing LAN or internet multiplayer with loadtime optimizations enabled?

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -63,9 +63,9 @@ it's `+set busywait 0` (setting the `busywait` cvar) and `-portable`
   enable/disable optimizations that speed up level load times (or more
   accurately, client connection).  
   sp stands for singleplayer and mp for multiplayer, respectively.  
-  The sp version is enabled by default (value 15) while multiplayer
+  The sp version is enabled by default (value 31) while multiplayer
   is 0.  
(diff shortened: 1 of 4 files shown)


Yamagi:
> I was also thinking, should I increase the default value of `sv_optimize_mp_loadtime` to 7? It's been a while now and I feel this should be safe. Does anyone here have experience with playing LAN or internet multiplayer with loadtime optimizations enabled?

Only a little bit, mostly from some test runs. I would say that we should increase it and see what happens. It's the only way to get feedback. I'll do it.

Yamagi:
Thank you :)

PR #1306, opened 2026-04-13
Update the CHANGELOG for 8.70

Yamagi:
First cut of CHANGELOG updates for the upcoming 8.70. Spelling correction, grammar and a second look if all relevant changes are mentioned and properly credited is appreciated.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,55 @@
+Quake II 8.60 to 8.70:
+- The default SDL version is now 3.0. SDL 2.0 is still supported as an
+  optional build flag. There are no plans to remove SDL 2.0 support in
+  the foreseeable future, however new features may not be implemented
+  for it.
+- The map list maps.lst is now searched in all folders. (by 0lvin)
+- `centerview` was improved to operate more like under Quake I. Instead
+  of snapping back to center immediately, the view now returns to the
+  center in a progressive way. The old behavior can be restored by
+  setting `cl_centertime 0`. (by protocultor)
+- Several improvements to the build system, most notably separate dirs
+  for debug and release builds. (by BjossiAlfreds)
+- Optional support for XDG directories on unixoid platform. If no
+  ~/.yq2/ directory exists, gamedata, savegames, etc are written to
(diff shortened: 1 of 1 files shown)


BjossiAlfreds:
I think it would be good to mention that `maxentities` cvar value needs to be increased to make use of the raised entity limit in single player or when hosting a server.

Also, might be good to mention that the server can now handle model / sound / image configstring index overflow, printing console warnings instead of dropping the server.

protocultor:
I'll be the annoying one in this PR:
- In line 7, where it says "than" should say "like".
- In line 16, where it says "now" should say "no".
- Between lines 26-29, I'm the author :P (#1263 points 5 & 6, #1277)

turol:
Line 17 "send" -> "sent"
Line 19 "lessen" is wrong. You're increasing the limit.
Line 22 "where" -> "were"
Line 34 "more smooth" -> "smoother"
Line 36 "mapfixed" -> "mapfixes"
Line 44 "cleanup" -> "cleanups"

BjossiAlfreds:
Lessening and lowering aren't synonyms. Lessening a limit implies you are taking away from the limit - making it less restrictive. Lifting or increasing would be better words though I guess.

Yamagi:
> I think it would be good to mention that maxentities cvar value needs to be increased to make use of the raised entity limit in single player or when hosting a server.

> Also, might be good to mention that the server can now handle model / sound / image configstring index overflow, printing console warnings instead of dropping the server.

I have added both.

> In line 7, where it says "than" should say "like".
> In line 16, where it says "now" should say "no".

Fixed.

> Between lines 26-29, I'm the author :P

Hmpf. Yes. Of course. Fixed.

> Line 17 "send" -> "sent"
> Line 22 "where" -> "were"
> Line 34 "more smooth" -> "smoother"
> Line 36 "mapfixed" -> "mapfixes"
> Line 44 "cleanup" -> "cleanups"

Fixed.

> Line 19 "lessen" is wrong. You're increasing the limit.

> Lessening and lowering aren't synonyms. Lessening a limit implies you are taking away from the limit - making it less restrictive. Lifting or increasing would be better words though I guess.

I have changed the wording.

PR #1308, opened 2026-04-14
curl: harden HTTP download error paths and write callback

devnexen:
Fix write callback to return byte count instead of item count so curl validates partial writes correctly. Cast to byte in URL percent-encoding to avoid signed promotion writing 9 bytes when the pointer only advances 3. Clean up dangling queueEntry, close the temp file and decrement pendingCount when curl_multi_add_handle fails, so the download slot is recoverable.


Yamagi:
Thank you :)

PR #1311, opened 2026-05-03
server: cache per-client leaf data across SV_Multicast fanout

devnexen:
SV_Multicast iterates every connected client and previously recomputed CM_PointLeafnum + CM_LeafArea + CM_LeafCluster on the receiver's origin for each event. With many multicast events per server frame (sounds, temp entities, weapon/explosion effects), the same BSP descent and area lookup ran repeatedly per client even though the client origin only changes once per frame.

Cache the leaf/area/cluster trio on client_t, keyed by sv.framenum and origin. The first SV_Multicast of a frame populates the cache; subsequent events in the same frame reuse it, collapsing per-client lookup cost from O(events_per_frame) BSP walks to one. The origin compare keeps the cache correct if a client teleports or is moved mid-frame, and the underwater retry path stays dynamic since it probes a different point.

### DIFF
--- a/src/server/header/server.h
+++ b/src/server/header/server.h
@@ -147,6 +147,13 @@ typedef struct client_s
 	int challenge;                      /* challenge of this user, randomly generated */
 
 	netchan_t netchan;
+
+	/* per-frame caches for SV_Multicast fanout */
+	vec3_t cached_origin;
+	int cached_leafnum;
+	int cached_area;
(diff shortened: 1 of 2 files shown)


Yamagi:
Looks good. I'll merge it after the upcoming release.

Yamagi:
Thank you :)

PR #1312, opened 2026-05-18
Update Dockerfile to ubuntu 24:04 LTS

Opvolger:
Update the Dockerfile to supported Ubuntu and new Dockerfile specs


Yamagi:
Looks good. Thank you!

PR #1313, opened 2026-05-19
ref_soft: Fix particle clipping and improve translucent depth sorting

Shockblast:
Greetings, community.

This PR contains three small fixes to improve the Software Renderer (`ref_soft`), resolving particle disappearing issues and depth sorting bugs for translucent objects.

#### **Summary of Changes:**

* **Fix particle clipping at steep angles (`sw_part.c`):**
  * **Problem:** Particles of all types would suddenly disappear or flicker at certain camera angles (very noticeable at lower resolutions like 240p).
  * **Fix:** Corrected the depth-test calculation to ensure accurate particle culling regardless of the camera's angle.

* **Enable depth writing for translucent sprites (`sw_poly.c`):**
  * **Problem:** Translucent sprites were drawn without updating the Z-buffer, causing sorting issues when they overlapped.
  * **Fix:** Turned on depth writing for translucent sprites to ensure they overlap and occlusion works naturally.

* **Fix depth writing and screen tracking for 33% translucent models (`sw_polyset.c`):**
  * **Problem:** Translucent models had rendering artifacts and flickering.
  * **Fix:** Ensured 33% translucent models correctly write to the Z-buffer and trigger viewport damage tracking.

These fixes make the software renderer more stable and visually consistent. I can provide screenshots showing the before and after if needed!

I look forward to hearing your comments.

Before:
bfg-antes

After:
bgf-ahora

Before:
rail-antes

After:
rail-ahora

> **Note on AI Assistance:**
> *I prepared these fixes and code modifications with the assistance of an AI coding partner. All changes have been manually reviewed, compiled, and verified to ensure they meet Yamagi Quake II's standards.*

### DIFF
--- a/src/client/refresh/soft/sw_part.c
+++ b/src/client/refresh/soft/sw_part.c
@@ -101,7 +101,10 @@ R_DrawParticle(particle_t *pparticle, int level)
 	** render the appropriate pixels
 	*/
 	count = pix;
-	if ((pz[(vid_buffer_width * count / 2) + (count / 2)]) > izi)
+
+	int half_count = count / 2;
+
+	if (pz[(vid_buffer_width * half_count) + half_count] > izi)
 	{
 		// looks like under some object
 		return;
--- a/src/client/refresh/soft/sw_main.c
+++ b/src/client/refresh/soft/sw_main.c
@@ -820,9 +820,8 @@ R_DrawEntitiesOnList
 =============
 */
 static void
-R_DrawEntitiesOnList (void)
+R_DrawEntitiesOnList (qboolean translucent)
 {
-	qboolean translucent_entities = false;
 	int i;
 
 	if (!r_drawentities->value)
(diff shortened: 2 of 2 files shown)


src/client/refresh/soft/sw_part.c
```diff
@@ -101,10 +101,13 @@ R_DrawParticle(particle_t *pparticle, int level)
 	** render the appropriate pixels
 	*/
 	count = pix;
-	if ((pz[(vid_buffer_width * count / 2) + (count / 2)]) > izi)
 	{
```
Yamagi, line 108: Is the anonymous scope necessary?

Yamagi:
@0lvin, what do you think?

0lvin:
For particles looks as correct solution, except part that checked only middle line of particles, more correct way will be check each pixel, but need to check performance.

For translucent surfaces, I am not sure that we should change zbuffer on rendering, as result we could receive other artifacts when some of surfaces is partially sorted not by far to near direction. 

I will try to merge to remaster code and check.

Overall logic of changes looks good as for me.

src/client/refresh/soft/sw_part.c
Shockblast: nop, sorry

src/client/refresh/soft/sw_part.c
```diff
@@ -101,7 +101,10 @@ R_DrawParticle(particle_t *pparticle, int level)
 	** render the appropriate pixels
 	*/
 	count = pix;
-	if ((pz[(vid_buffer_width * count / 2) + (count / 2)]) > izi)
+
+	int half_count = count / 2;
+
+	if (pz[(vid_buffer_width * half_count) + half_count] > izi)
```
0lvin, line 107: Looks like add brackets should be enough, to resolve issues that width multiply to count and only after it divided.
```
if (pz[(vid_buffer_width * (count / 2) + (count / 2)] > izi)
```

Looks like it's not related to view angle, have I missed something?

0lvin:
Original code looks correct, code does not need to set zbuffer. In case when render are going from far surface to near one it will just write values that's never be bigger than previous, if code going from near surface to far, code works incorrectly because it should combine all transparent surfaces. Final pixel color could be incorrect because code has only 8 bits for color and not all combination of color provides something correct as result.

Set damage to zbuffer looks good idea, as code further checked only this section to understand that frame should be sent to render.

Have I missed something?

Update: not sure that we should touch damage values also as code will not change z value in such case

src/client/refresh/soft/sw_part.c
Shockblast: > Looks like add brackets should be enough, to resolve issues that width multiply to count and only after it divided.
> 
> ```
> if (pz[(vid_buffer_width * (count / 2) + (count / 2)] > izi)
> ```
> 
> Looks like it's not related to view angle, have I missed something?

It was my micro-optimization obsession. I know the compiler handles it anyway. I can revert it if you prefer.

0lvin:
Could you please provide some additional examples where combining color is broken, like with some transparent window or static effect and additional effect from weapon?

Like example in case of water and BFG sprite before looks like color is combined, and after effect from BFG is not applied to part under water.

Shockblast:
> Could you please provide some additional examples where combining color is broken, like with some transparent window or static effect and additional effect from weapon?
> 
> Like example in case of water and BFG sprite before looks like color is combined, and after effect from BFG is not applied to part under water.

Let me prepare the old build and a simple test map and post a screenshot.

0lvin:
Also could you please add to each of commits mention that it created with AI, like:
```
Co-authored-by: Copilot <https://copilot.microsoft.com>
```

Of course if commit contains some changes created with help from AI :-)

Shockblast:
> Also could you please add to each of commits mention that it created with AI, like:
> 
> ```
> Co-authored-by: Copilot <https://copilot.microsoft.com>
> ```
> 
> Of course if commit contains some changes created with help from AI :-)

Sure, once we're done going over the details here, I'll do it (For the 3 commits I made, I received help from the AI with feedback and editing).

Now the screenshots in the old build.

after1
after2

Now with my “fixes”

before1

before2


The sorting is clearly better than before, but I noticed the BFG ball can still occlude translucent surfaces that should be in front of it (image bellow, the lava pillar and water panel on the left). So it's an "improvement" but not a complete fix.

happy to hear your suggestions on the correct approach.

problem

0lvin:
After code produced incorrect result, BFG sprite is transparent and after change red transparent box is not combined as result transparent BFG hides transparent red box.

More notes:
Transparent objects do not write to the depth buffer (Z-buffer) because their final appearance relies on blending colors with whatever is behind them. To achieve this correctly and optimize performance, transparent objects are sorted and rendered last in the pipeline.

Expected Render Sequence in R_DrawEntitiesOnList
* To ensure correct visual depth and blending, the rendering pipeline should follow this specific order:
Solid Background/Walls: Draw opaque environment geometry. Write directly to the Z-buffer and set the initial pixel colors.
* Solid Models: Draw opaque dynamic objects. Check against the existing Z-buffer to discard occluded pixels, write new depths to the Z-buffer, and overwrite the pixel colors.
* Transparent Objects: Draw transparent models, sprites, and particles. Check against the Z-buffer to ensure they are not hidden behind solid geometry, and combine (blend) their colors with the existing pixel data without writing to the Z-buffer.

First step of the sequence focuses on writing the z + color buffer and never read, while the final step strictly reads from it and write combined color.

Shockblast:
> After code produced incorrect result, BFG sprite is transparent and after change red transparent box is not combined as result transparent BFG hides transparent red box.
> 
> More notes: Transparent objects do not write to the depth buffer (Z-buffer) because their final appearance relies on blending colors with whatever is behind them. To achieve this correctly and optimize performance, transparent objects are sorted and rendered last in the pipeline.
> 
> Expected Render Sequence in R_DrawEntitiesOnList
> 
> * To ensure correct visual depth and blending, the rendering pipeline should follow this specific order:
>   Solid Background/Walls: Draw opaque environment geometry. Write directly to the Z-buffer and set the initial pixel colors.
> * Solid Models: Draw opaque dynamic objects. Check against the existing Z-buffer to discard occluded pixels, write new depths to the Z-buffer, and overwrite the pixel colors.
> * Transparent Objects: Draw transparent models, sprites, and particles. Check against the Z-buffer to ensure they are not hidden behind solid geometry, and combine (blend) their colors with the existing pixel data without writing to the Z-buffer.
> 
> First step of the sequence focuses on writing the z + color buffer and never read, while the final step strictly reads from it and write combined color.

Thank you for the detailed explanation! I now understand why writing to the Z-buffer for transparent objects breaks the blending. I'll look into fixing the back-to-front sorting instead.
I'll update the PR once I have a proper solution, also update the commits co-author. 😅

EDIT: Sorry, I misunderstood what I read and what I said; I'm still looking at the code for this.

Shockblast:
Thanks for the feedback, @0lvin. You were right, translucent surfaces should not write to the depth buffer, as doing so causes sorting artifacts with objects rendered behind them.

I've updated the commit: depth writing has been removed from the 33% translucent model functions. The VID_DamageZBuffer calls remain, as they are needed for correct viewport damage tracking.

If this fix still doesn't seem right to you, I'm happy to drop it and only apply the particle clipping fix.

Here are updated screenshots showing the current behavior: 
Captura de pantalla_20260520_225403
Captura de pantalla_20260520_225626

0lvin:
Run of VID_DamageZBuffer for transparent or without write to zbuffer itself is not required, code used getter for this values only for solid spans, to check that update zbuffer is not required. Look to VID_CheckDamageZBuffer usage.

Shockblast:
Okay, I tested it and it works fine here, at least. Do you need the test minimap?

0lvin:
Could optimization with skip iteration through entities without transparent entities restored? Before changes code set flag if found transparent entity, and skip second iteration through entities list without such.

Shockblast:
@0lvin I've implemented the changes based on your suggestion. The AI suggested this changes to me, and it's just a simple flag in the code. Personally, I like it.
(It was much simpler than it looked, and here I was, racking my brain, I should have asked him first :rage4:)

I have a quick question for you (or more).

How do you conduct performance tests?
Do you use QEMU to simulate low-performance hardware?
Or do you use other metrics?
Or is this just theory you’re already familiar with?

Because, for example, measuring milliseconds on a modern machine isn't useful to me—the results are almost identical to those from a PC that's five years old or older.

0lvin:
Open benchmark suite could be used https://openbenchmarking.org/test/pts/yquake2&eval=3dd4b4c319ca9bd797063e1792c2600eae46e345#metrics
for tests.

Yamagi:
Any progress here?

Shockblast:
> Any progress here?

I remember that the requested revisions have been made, but if you want, I can take another look at the sprite commit, just in case.

Yamagi:
@0lvin, are you okay with this?

0lvin:
> @0lvin, are you okay with this?

Yes, looks good.

Yamagi:
Thank you both :)

PR #1314, opened 2026-05-20
Clean up function list and type-safe function saving

### LINKED ISSUE #1310: Saving / loading functions and type safety
The savegame system has a type safety issue with functions (for example it could link a think function to the touch pointer). I made an experiment sokving this problem: https://github.com/BjossiAlfreds/yquake2/tree/sg-funcs

It also removes all the unnecessary entries, reducing the size by about 2/3.

Some key points:
* The function list is now segmented based on which function pointer they are for.

* The `structdef_t` type holds static info that describes a struct. Currently it has a list of fields, struct size and a `fptrList_t` pointer.

* The `fptrList_t` type maps field offsets to function lists. Example: `{FOFS(blocked), &fnlist_blocked}`.

* `ReadStruct` and `WriteStruct` generalize reading/writing struct instances, using the struct definition.

* The `ReadField` and `WriteFieldX` functions take the struct definition's `fptrList_t` which is passed to `GetFunctionList` to get the correct list for the field if it is a function pointer field.

Is this something you would like in master?

Another task I'd want to do at some point is to mark functions with a blank macro, like this for example:

```
void FNTYPE(mi_stand)
soldier_stand(edict_t *self)
```

And then extend the extractfunctions tool to optionally support looking for these markers and using them to organize into several lists based on the type. Or maybe write a new tool from scratch for this purpose.

What do you guys think? Too ambitious maybe?

BjossiAlfreds:
Fixes https://github.com/yquake2/yquake2/issues/1310

This PR cleans up the function list of the savegame system, addresses the type-safety issue descrived in the issue report and adds new shell scripts, `fngen` and `mmgen`, that mechanically generate the function and mmove lists respectively.

* Splits the function list into several smaller lists, one for each function pointer. This ensures only compatible functions can be linked to a given pointer
* Removes unnecessary entries from the function list
* Marks `moveinfo.endfunc` functions in `g_func.c` with a special `/* ->moveinfo.endfunc = func_name; */` comment. This is needed because these functions are not assigned to the pointer directly but through the `Move_Calc` functions instead

Data generators:
* `stuff/fngen` are shell scripts for generating function tables and prototypes for each function pointer type
* `stuff/mmgen` are shell scripts for generating the mmove table and prototype list
* Both have readmes files that go into more details

With these changes are some extensions to the savegame system
* `fptrList_t` maps function pointers (as field offsets) to `functionList_t` objects
* `structdef_t` objects hold static context about structs that get saved: list of fields, pointer to a `fptrList_t` and their `sizeof` value
* `ReadStruct` and `WriteStruct` generalize reading / writing structs. Previously this code was copy-pasted in a few places. Takes in a pointer to a `structdef_t` object
* A little bit of code refactoring, more `const` etc.

Some more code cleanup and stability:
* Removed unused `F_GSTRING` field type
* Moved spawntemp fields to their own list and removed `FFL_SPAWNTEMP` flag
* Removed unused `F_CLIENT` field type
* Prevent writing `client` pointer value to savegames
* Sanitize `client` pointer to `NULL` after reading `edict_t` struct


BjossiAlfreds:
I pushed a commit to separate spawntemp foelds from entity fields. Did it for these reasons:
* Having fields from 2 different structs in the same list was just a forgetting-to-mark-FFL_SPAWNTEMP away from trouble
* The entity fields list is used in 2 different contexts (savegames and entity spawning) but spawntemp fields have no relevance to savegames
* The `structdef_t` API gives field lists a more structured purpose

0lvin:
In theory fields could be sorted and used binary search instead iteration, but it failed as some fields had incorrect values as result when I have tried it sometimes [ago](https://github.com/yquake2/yquake2remaster/commit/9a13efe9b491adc76507daaf3a7ecf662b615b3c). So I reverted such change. I suppose it was related to some dependency between fields.


if you found dependency between fields it could improve load time. Just idea;-)

BjossiAlfreds:
There are 2 fields named `"item"` - `FOFS(item)` and `STOFS(item)`. Maybe that is why the binary search was not working correctly? This commit would fix that. You could also simply rename the `FOFS(item)` field since it isn't used as a spawn field. Like `"item_ptr"`.

As for load times, pretty much all the load time goes into client-server networking and pak loading. Optimizing field search wouldn't have any effect on that.

BjossiAlfreds:
I made a commit that removes the `F_GSTRING` field type. It was unused and the code for it was incomplete. We wouldn't need a new field for this any ways, could just pass the memory tag to `ReadStruct`.

Also made a commit that removes the `F_CLIENT` field type. The `client` pointer was not in the `entfields` list and this pointer does not need to be saved because it is restored implicitly after reading the entities in `ReadLevel`.

I also set the `client` pointer to `NULL` before saving and after loading. The un-serialized memory address was being written into savegames and was present in the read struct until being properly restored after the entity reading loop in `ReadLevel`.

That also avoids a possible security issue. Only the `client` pointers of player entities are restored. So if a save contained an arbitrary value in the `client` pointer of a non-player entity it would carry over into the game code and get dereferenced.

Yamagi:
Thank you :)

PR #1317, opened 2026-05-22
client, server: Optionally use system copies of libminizip and Zlib

smcv:
Some Linux distributions have a system copy of the minizip library, either built from zlib's contrib directory or as a separate package. In such distributions, having a private copy in applications is generally discouraged, because it means more places that need to be updated every time a security issue is found.

---

For the moment this only works for the `Makefile` build system, and the CMake build system is still hard-coded to use the vendored minizip and miniz.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,10 @@ WITH_RPATH:=yes
 # When disabled SDL 2 is used instead of SDL 3.
 WITH_SDL3:=yes
 
+# WITH_SYSTEM_MINIZIP
+# When disabled, an included copy of minizip and miniz is used.
+WITH_SYSTEM_MINIZIP:=no
+
 # WITH_SYSTEMWIDE
 # Enable systemwide installation of game assets.
 WITH_SYSTEMWIDE:=no
@@ -1048,11 +1052,6 @@ CLIENT_OBJS_ := \
 	src/common/shared/flash.o \
 	src/common/shared/rand.o \
 	src/common/shared/shared.o \
-	src/common/unzip/ioapi.o \
-	src/common/unzip/unzip.o \
-	src/common/unzip/miniz/miniz.o \
-	src/common/unzip/miniz/miniz_tdef.o \
-	src/common/unzip/miniz/miniz_tinfl.o \
 	src/server/sv_cmd.o \
 	src/server/sv_conless.o \
 	src/server/sv_entities.o \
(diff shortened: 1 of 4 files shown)


Yamagi:
I'm totally fine with this. minizip should be stable enough for future updates not causing problems and as you say, it might make the security guys a little bit more happy. I think not touch the CMakeLists.txt is okay, it has been officially unmaintained for years and is (or should) most used for generating IDE projects. Like the MSVC solution files.

Can you add a short paragraph to https://github.com/yquake2/yquake2/blob/master/doc/070_packaging.md explaining using the system wide copy of minizip for package maintainers?

smcv:
> Can you add a short paragraph to https://github.com/yquake2/yquake2/blob/master/doc/070_packaging.md explaining using the system wide copy of minizip for package maintainers?

Sure, I'll try to get to that at some point and update this PR.

If yquake2 has any other vendored dependencies that also exist as third-party projects, then a similar technique would probably be applicable to them. I haven't done the equivalent for the STB stuff, because I'm not sure to what extent those are API-/ABI-stable.

smcv:
> I think not touch the CMakeLists.txt is okay, it has been officially unmaintained for years and is (or should) most used for generating IDE projects. Like the MSVC solution files.

Conveniently, it's exactly the OSs with IDE-based workflows where vendoring all the dependencies tends to be most appealing!

Yamagi:
> If yquake2 has any other vendored dependencies that also exist as third-party projects, then a similar technique would probably be applicable to them. I haven't done the equivalent for the STB stuff, because I'm not sure to what extent those are API-/ABI-stable.

To my experience STB isn't stable enough to pull the headers from an external source. The API might be backward compatible but in the past nearly every update of the headers broke something. That's more or less expected, STB is designed to be bundled with the applications. The same goes for HandmadeMath.

All other dependencies are external:

* cURL
* libGL
* OpenAL
* SDL

All of them are linked at runtime, SDL through the rtld and the other ones are opened with `dlopen()`.

smcv:
> To my experience STB isn't stable enough to pull the headers from an external source. The API might be backward compatible but in the past nearly every update of the headers broke something.

I'm glad I pushed back against requests to switch from vendored STB stuff to a shared copy in other games in Debian, then!

> All of [the other deps] are linked at runtime, SDL through the rtld and the other ones are opened with `dlopen()`.

That reminds me that at some point I should also propose a PR that adds `SDL_ELF_NOTE_DLOPEN` for each dlopen'd library, so that distros can use https://github.com/systemd/package-notes to generate mandatory or optional dependencies as appropriate. (Only SDL3 provides that feature, but `#ifdef`'ing it to keep the SDL2 build working shouldn't be difficult.)

Yamagi:
> Sure, I'll try to get to that at some point and update this PR.

Any progress here?

smcv:
> Can you add a short paragraph to https://github.com/yquake2/yquake2/blob/master/doc/070_packaging.md explaining using the system wide copy of minizip for package maintainers?

Done, sorry for the delay.

Yamagi:
Many thanks!

PR #1318, opened 2026-05-24
sound: Random music playback

apartfromtime:
This is to fix a mistake on the initial commit, true randomness can playback the same track twice. To keep things simple I have removed the "truly random" option, keeping "random", as a skip or reshuffle key-bind would be more appropriate.

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -348,9 +348,7 @@ it's `+set busywait 0` (setting the `busywait` cvar) and `-portable`
   - `0`: Loop the current track (the default).
   - `1`: Play the current track once, then stop.
   - `2`: Play all available tracks in a linear sequence.
-  - `3`: Shuffle through the available tracks, never play the same track
-         twice in a row.
-  - `4`: Shuffle through the available tracks, may play the same track
+  - `3`: Shuffle through the available tracks, may play the same track
          multiple times in a row.
 
 * **s_doppler**: If set to `1` doppler effects are enabled. This is only
--- a/src/client/menu/menu.c
+++ b/src/client/menu/menu.c
@@ -2706,7 +2706,6 @@ Options_MenuInit(void)
 		"play once",
 		"sequential",
 		"random",
-		"truly random",
 		0
 	};
 
--- a/src/client/sound/ogg.c
+++ b/src/client/sound/ogg.c
@@ -531,7 +531,6 @@ OGG_PlayTrack(const char *track, qboolean cdtrack, qboolean immediate)
 		newtrack = (curtrack + 1) % (ogg_maxfileindex + 1) != 0 ? (curtrack + 1) : 2;
 	} break;
 	case 3:			// random
-	case 4:			// random with true randomness
 	{
 		int retries = 100;
 		newtrack = 0;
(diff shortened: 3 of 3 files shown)


protocultor:
This is more of an issue of "_unclear text_" or "_confusing documentation_", but the point of `truly random` is precisely that: it is truly random, meaning the same track is able to appear twice in a row. As it can be seen in the code you deleted, the point of the other "random" option (number 3) is to **force a re-attempt** of the shuffle if the current track appears again.

apartfromtime:
> This is more of an issue of "_unclear text_" or "_confusing documentation_", but the point of `truly random` is precisely that: it is truly random, meaning the same track is able to appear twice in a row. As it can be seen in the code you deleted, the point of the other "random" option (number 3) is to **force a re-attempt** of the shuffle if the current track appears again.

I agree that the confusion is in "truly random" as "truly random" is by definition "random". A sample size of one track would loop forever (or 100 in the code) if "truly random" was selected. I can switch the defintions if there is enough resistance to the idea, but i'd be inclined to offer the end-user a key-bind that prompts a reshuffle of any playback tracks, keeping the single definition of "random", if that clears up any confusion.

Yamagi:
We had this before in #1143. After the new *true random* code was merged people - okay, one user - started complaining that the same track could be played multiple times in a row. Because of that I altered the code again, `3` was reverted to the old behavior skipping over duplicated tracks and `4` added to allow the same track to be played several times.

We can remove `4` and add a `ogg random` sub command which plays a random track. That could be bind to a key if decided. But we shouldn't change the behavior of `3`. That will lead to complains.

j4reporting:
I wasn't  complaining. I was simply pointing out that most users  expect  'shuffle mode' rather than a  truly random playback mode.  
That's why Apple introduced shuffle mode for its ipods --  too many customers were confused when a song played a second time.

Yamagi:
I've implemented an `ogg skip` command in 56d86ecd9c9bc3e96379234122e82aca6aa4c322. This can be bound a key. Depending on `ogg_shuffle` it either skips back to the beginning of the current track, replays the last track or skips to another random track.

Yamagi:
> I wasn't complaining. I was simply pointing out that most users expect 'shuffle mode' rather than a truly random playback mode.

And that was a valid point.

protocultor:
Perhaps removing the current `4` option was the simplest solution from the very beginning.

apartfromtime:
> Perhaps removing the current `4` option was the simplest solution from the very beginning.

That's all I'm suggesting. Keep it simple, make "random", a dice roll and provide a function that can be bound to a key if you want to skip or re-roll the dice.

protocultor:
> That's all I'm suggesting. Keep it simple, make "random", a dice roll and provide a function that can be bound to a key if you want to skip or re-roll the dice.

But what you actually did in your PR was to get rid of `3`, and replace it with what `4` did before. That is... not the same thing I'm saying.
Many thanks for the work, but I'm in the boat of "this PR is no longer needed" after https://github.com/yquake2/yquake2/commit/56d86ecd9c9bc3e96379234122e82aca6aa4c322 .

Yamagi:
So, do we agree that `4` was a bad idea which confuses users and can be removed now that `ogg skip` is implemented?

protocultor:
> So, do we agree that `4` was a bad idea which confuses users and can be removed now that `ogg skip` is implemented?

Definitely yes, but unlike the present PR, please keep the current `3`.

BjossiAlfreds:
I agree with you guys, get rid of 4 but keep 3 as-is.

BjossiAlfreds:
On a side note, there is a simpler and more efficient way to randomly choose the next track and avoid same track playing twice in a row. Like this:

```
newtrack = randk() % ogg_maxfileindex;

if (newtrack >= curtrack)
{
	newtrack++;
}
```

PR #1319, opened 2026-05-24
client: Add 2 server frame delay on lerping newly created client lasers

### LINKED ISSUE #1316: [Bug] Monster lasers in xatrix lerp repeatedly
This was posted in https://github.com/yquake2/yquake2/issues/1286
https://github.com/yquake2/yquake2/issues/1286#issuecomment-4514751264

I can't confirm this visually since my eyesight is extremely bad, but with debug printing I think I know what is being observed in-game.

The laser guards (and other xatrix laser monsters) create new laser entities repeatedly where each one only lives for 2 server frames. This essentially "restarts" the lerping each time, lerping from the laser's start position at the gun to the destination once it has traced.

Here is what I believe is the sequence of events:
1. Server: The laser is spawned. `origin` is set to start pos but `old_origin` is either `0 0 0` or the same as start pos, depending on if the laser comes after the monster in the entity list (`origin` is copied into `old_origin` in `G_RunFrame`)
2. Client No lerping or lerps from `0 0 0` to start pos
3. Server: The laser thinks and traces. `old_origin` gets the value of `tr.endpos`
4. Client: Lerps from prev `old_origin` to the updated one
5. Sever: Laser is freed by `G_FreeEdict`

There are a few things we could do about this:
* Use the `frame` field to identify monster lasers. As far as I know they are the only lasers that use the value 2, level lasers use 4 or 16
* Determine how "old" the entity is client-side and disable lerping for the first couple of its server frames
* Make laser lerping opt-in. By default, lasers are flagged with `RF_BEAM` but not `RF_FRAMELERP`. We could make it so if both flags are set then we use the beam lerping. That way we can mark level lasers with `RF_FRAMELERP` but skip the monster lasers in xatrix.
* Revert to the old behavior. This is not an important change

Thoughts?

BjossiAlfreds:
Fixes https://github.com/yquake2/yquake2/issues/1316

This fixes toggled lasers lerping every time they turn on and lerp-spamming xatrix monster lasers that are short-lived and spawned repeatedly.

Now there is a `serverframe_created` in `centity_t`, set when a client receives an entity that was previously absent in packet(s). Subtracting this value from `cl.frame.serverframe` gives the client lifetime of the entity relative to number of server frames that have passed.

**Known minor issue**: `sv.framenum` increases even while the game is paused. This means if the game is paused during one of the delay frames of the entity, it will appear lerped on the paused screen. This is a minor issue and so won't be fixed unless someone knows a better way to do this.

Thanks to @protocultor for reporting this issue, suggesting using server frames and testing the fix.

### DIFF
--- a/src/client/cl_entities.c
+++ b/src/client/cl_entities.c
@@ -27,6 +27,13 @@
 #include <math.h>
 #include "header/client.h"
 
+/* skip lerping the first server frames after client receives it
+ * avoids lerping toggled lasers
+ * and fixes lerp-spamming with xatrix monster lasers
+ * they spawn short-lived lasers repeatedly
+ */
+#define LASERLERP_NSKIPS 2
+
 void
 CL_AddPacketEntities(frame_t *frame)
 {
@@ -122,7 +129,8 @@ CL_AddPacketEntities(frame_t *frame)
 		ent.oldframe = cent->prev.frame;
 		ent.backlerp = 1.0f - cl.lerpfrac;
 
-		if (renderfx & RF_BEAM)
+		if ((renderfx & RF_BEAM) &&
+			(cl.frame.serverframe - cent->serverframe_created) >= LASERLERP_NSKIPS)
 		{
 			for (i = 0; i < 3; i++)
 			{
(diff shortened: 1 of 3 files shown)


Yamagi:
Many thank to all involved :)

PR #1321, opened 2026-05-29
workflows: fix coverity and arm64 builds

### LINKED ISSUE #1320: CI builds stopped working on ARM64
I tried latest release from CI builds for ARM64 (specifically RPi5) and I found out, that latest releases aren't working on it. When I try to run it, I get following error:

`bash: ./quake2: cannot execute binary file: Exec format error`

This error was suspicious, so I run file info on that file, and got this:

`quake2: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=52c1bc1161935030576844c0a86923ad7e3d0eea, for GNU/Linux 3.2.0, not stripped`

This doesn't seems right, so I tried some older builds and found out, that following one is last one that works just fine:
https://github.com/yquake2/yquake2/actions/runs/25928760011

When I tried to do a file info on this one, I got something that actually make sense on ARM64:

`quake2: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, BuildID[sha1]=b50c9a80b205800fc01b1dea3cf8eedb5e7d0c81, for GNU/Linux 3.7.0, not stripped`

0lvin:
Fixes: https://github.com/yquake2/yquake2/issues/1320

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -33,32 +33,31 @@ jobs:
       run: |
         if [ -z "${{ secrets.COVERITY_SCAN_TOKEN }}" ]; then
           echo "COVERITY_SCAN_TOKEN is not set. Skipping workflow."
+          echo "skip=true" >> $GITHUB_ENV
         else
-            wget https://scan.coverity.com/download/cxx/linux-ARM64 \
-              --post-data "token=${{secrets.COVERITY_SCAN_TOKEN}}&project=${{github.repository}}" \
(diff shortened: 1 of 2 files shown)


0lvin:
Fixes arm64 build base image and add ignore of coverity if site is unavailable.

Yamagi:
Thank you!

PR #1322, opened 2026-05-30
game: Fixed chick, flipper, infantry and parasite not playing their search sounds

### LINKED ISSUE #1315: [Legacy bug] Infantry and Parasite don't play their walk-search sound effects
Infantry and Parasite both precache `sound_search` sounds but never use them. Parasites even have a `parasite_search` function that plays the sound but is not linked to the `monsterinfo.search` pointer.

Should I do a PR to finish these lose ends? Judging by the code this seems like an oversight on id's part. I see no reason why they would want to remove these sounds for these two enemy types specifically.

BjossiAlfreds:
Fixes https://github.com/yquake2/yquake2/issues/1315

Discovered that `chick` and `flipper` also precached search sounds that they never used.

EDIT: And discovered that flipper was precaching a `sound_idle` and not using it.

### DIFF
--- a/src/game/monster/brain/brain.c
+++ b/src/game/monster/brain/brain.c
@@ -194,13 +194,25 @@ mmove_t brain_move_idle =
 void
 brain_idle(edict_t *self)
 {
+	int n;
+
 	if (!self)
 	{
 		return;
 	}
 
-	gi.sound(self, CHAN_AUTO, sound_idle3, 1, ATTN_IDLE, 0);
-	self->monsterinfo.currentmove = &brain_move_idle;
+	n = randk() & 3;
+
+	if (n <= 1)
+	{
+		self->monsterinfo.currentmove = &brain_move_idle;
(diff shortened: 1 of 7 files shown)


BjossiAlfreds:
I discovered that Brain has 3 idle sounds. All 3 are being precached but only `sound_idle3` was used. I figured I'd add a commit that makes `brain_idle` use the first 2 as well.

Basically a random number between 0-3 is chosen. If 0 or 1, his normal idle routine is used, otherwise he plays one of the new sounds, based on if the random number is 2 or 3.

While the 2 sounds were being precached, showing intention to use them, this code is more my interpretation than fixing a bug / oversight. You could skip this commit if you don't like it. If nothing else this does add to the sound experience of the game and should be fairly harmless.

BjossiAlfreds:
Made another discovery and a commit that may be controversial. `chick` precaches a `sound_fall_down` that was never being used. It was almost definitely meant for `chick_move_death2` so I added it there. I quite like the outcome. What do you think?

If you don't like this change I could convert this commit into a removal one, removing the static variable and adding a comment to mark this as unused.

PR #1324, opened 2026-06-07
game: Improved robustness and a bit of refactoring in physics code

BjossiAlfreds:
This PR improves robustness in the game's physics code. Also a bit of clean-up refactoring

* Unknown movetype no longer drops the server. Now the game prints a warning and defaults to `MOVETYPE_NONE`
* `SV_Push` can no longer corrupts memory if more than `MAX_EDICTS` entities are being pushed, possible now that the server supports more than that. Now there is a `pushed_t`API that is used to handle this situation gracefully, avoiding both the memory corruption and the `gi.error` call
* A bit of refactoring: Marked some functions `static`, reduced scope of some variables, removed some useless comments and `obstacle` is now a local variable and `SV_Push` returns the obstacle instead of true/false

### DIFF
--- a/src/game/g_phys.c
+++ b/src/game/g_phys.c
@@ -1301,6 +1301,10 @@ G_RunEntity(edict_t *ent)
 			SV_Physics_Toss(ent);
 			break;
 		default:
-			gi.error("%s: bad movetype %i", __func__, (int)ent->movetype);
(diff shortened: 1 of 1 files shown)


Yamagi:
Thank you! :)

PR #1325, opened 2026-06-10
GL4.6 Feature: Bloom

atsb:
This adds bloom to the video menu for the GL 4.6 renderer.  It also fixes a bug that did not allow the gl4 colorlights to be switched off in the menu.

### DIFF
--- a/src/client/menu/videomenu.c
+++ b/src/client/menu/videomenu.c
@@ -45,6 +45,7 @@ static cvar_t *gl_anisotropic;
 static cvar_t *gl_msaa_samples;
 static cvar_t *gl3_colorlight;
 static cvar_t *gl4_colorlight;
+static cvar_t *gl4_bloom;
 static cvar_t *vk_dynamic;
 
 static menuframework_s s_opengl_menu;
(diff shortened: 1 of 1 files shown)


atsb:
Will anything be done about this?  I really don't like maintaining a renderer where I need to wait a week for a PR review so end users can use features.

Or can I have write access so I can push these myself?

Yamagi:
I've got a life besides Quake II. I try to spend at least one afternoon in two weeks for the project. Nevertheless: Thank you :)

atsb:
Thanks, I do as well, I just work quite fast :)

protocultor:
> Will anything be done about this? I really don't like maintaining a renderer where I need to wait a week for a PR review so end users can use features.

I know you do a lot of stuff for yourself, but this statement makes it clear you are not a fan of doing PR for other people's projects. Because I do, and Yamagi is practically the only one who responds in less than a month. Just saying.

atsb:
Thanks for your opinion.  You're terribly wrong, but that's ok, you're just like that.

It's merged and we're done.

Yamagi:
So, another round of burnout for me is okay as long as users can download the CI asset from the main repo instead of this PR? Sounds like a sensible trade-off.

PR #1328, opened 2026-06-26
menu: Improve stability in some menu code and minor optimizations

BjossiAlfreds:
This PR does a bit of stability improvements for the `DoSlide` and `Draw` functions as well as a bit of cleanup.

* Fixed mismatch of `static` for prototype and declaration of some functions
* Removed unused `errorpic` field from `menubitmap_s` struct
* Replaced 0 with `NULL` for pointer assignments
* `Menu_SlideItem` now return true/false depending on if their value changed with the interaction
* `Slider_DoSlide` and `Slider_Draw` now gracefully handle a `NULL` `s->cvar`
* `SpinControl_DoSlide` and `SpinControl_Draw` now gracefully handle `NULL` or empty itemlist and invalid selection
* These functions were also optimized and cleaned up a bit
* Optimized out large `startmap` 1024 byte buffer from `StartServerActionFunc` and moved the hard-coded start spot code to `GetStartSpot` function and made it more data-driven

I was thinking, have you ever noticed how the sound always plays when pressing left/right arrow on a slider/option, even if the value didn't change? I could do a commit that makes the sound only play if the option changes. Would be good for accessibility, then the user can hear when they have reached the end of the available range.

### DIFF
--- a/src/client/menu/header/qmenu.h
+++ b/src/client/menu/header/qmenu.h
@@ -90,7 +90,6 @@ typedef struct
 {
 	menucommon_s    generic;
 	char *          focuspic;	
-	char *          errorpic;
 	int             width;
 	int             height;
 } menubitmap_s;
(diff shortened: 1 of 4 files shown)


Yamagi:
> I was thinking, have you ever noticed how the sound always plays when pressing left/right arrow on a slider/option, even if the value didn't change? I could do a commit that makes the sound only play if the option changes. Would be good for accessibility, then the user can hear when they have reached the end of the available range.

No, I haven't. I use the menu rarely. But I have tries and it's indeed a little bit annoying, so a PR would be welcome.

Thank you for this cleanup! :)

PR #1329, opened 2026-07-01
sound: hoist invariant cvar reads out of hot paths

devnexen:


src/client/sound/sound.c
```diff
@@ -1278,7 +1279,12 @@ S_StartSound(vec3_t origin, int entnum, int entchannel, sfx_t *sfx,
 		}
 	}
 
-	cvar_t *game = Cvar_Get("game",  "", CVAR_LATCH | CVAR_SERVERINFO);
+	static const cvar_t *game = NULL;
+
+	if (game == NULL)
+	{
+		game = Cvar_Get("game",  "", CVAR_LATCH | CVAR_SERVERINFO);
```
DanielGibson, line 1286: does this still work as expected when loading another mod or doing some other kind of reinitialization?

devnexen: I ve tried in game changing to xatrix and it had worked for me.

Yamagi:
Same as in #1331. I'm not convinced that this is save and doesn't cause problem when the user changes `game` without restarting or the server resets the cvar.

Yamagi:
Thank you :)

PR #1330, opened 2026-07-01
server: compare squared distance for sound attenuation cull

devnexen:


Yamagi:
Can you explain why this is necessary?

0lvin:
It removes non required square root operation.

Yamagi:
Ah, I see. Thank you :)

PR #1331, opened 2026-07-01
client: look up game cvar once in CL_AddPacketEntities

devnexen:


Yamagi:
Is this really save? What happens when the player switches from baseq2 to rogue without restarting? Or when the server resets the `game` cvar to something else?

BjossiAlfreds:
The cvar list is only freed when exiting the game so storing a static pointer to the cvar should be safe, and since the cvar's value is not being cached but the pointer to it,the code should always get the current value, as long as the optimizer doesn't do anything funny.

My gut feeling tells me this is safe.

devnexen:
I ve tried with xatrix and rogue changing after few minutes of baseq2, no issue so far

Yamagi:
Jepp, I stepped through it and this is indeed save because the pointer to the cvar struct stays the same and value inside the struct is updated.

Thank you :)

PR #1335, opened 2026-07-05
server: Prevent crash on invalid modelindex values

BjossiAlfreds:
The `sv.models` array was being indexed without any bounds checks in `SV_HullForEntity`. A buggy or malicious game library could assign invalid values to `ent->s.modelindex` to trigger a crash or returning an undefined headnode value.

And it is no longer a fatal error if a `SOLID_BSP` entity has no model. There will simply be no collision with the entity.


Yamagi:
Thank you!

PR #1337, opened 2026-07-10
menu: Sound related fixes/improvements and further stability

BjossiAlfreds:
This PR continues where https://github.com/yquake2/yquake2/pull/1328 left off, with further stability improvements and menu sound fixes/improvements.

Sound fixes/improvements
* Fixed video menu using the wrong move sound. All other menus use `misc/menu2.wav` but the video menu was using `misc/menu1.wav`
* Fixed menu exit sound playing louder for some menus. `Default_MenuKey` was causing `S_StartLocalSound` to be called twice - in `M_Keydown` and `M_PopMenu`
* Pressing Left/Right arrow now only plays a sound if the selected option changed its value
* Pressing Enter now only plays a sound if the option has a callback action

Other fixes
* Fixed main menu forgetting previous cursor position on each re-open. Commit 07fc65f9c51b02009693da3ace49c411f74a5e05 added a memset for `s_main` which reset the cursor value to 0. The commit comment doesn't indicate that this was intentional
* Fixed invert mouse/joystick and lookstrafe options changing state even if the option didn't change to a different value. Their callbacks were changing the cvar values without checking the current value of the option. So for example pressing Enter on invert mouse would turn on invert mouse even if the option still says "no"

Stability
* Additional checks in cursor drawing code
* Upper bound checks for spin control `curvalue`
* `S_StartLocalSound` arg is now const
* Video menu code now uses `Default_MenuKey` and `M_Banner` to avoid code copy-paste

PR #1338, opened 2026-07-11
Use curl instead wget to fix win32 build crash

0lvin:


0lvin:
Additionally contains small fixes for cppcheck, update SDL3 version and add const.

Issue with wget crash found in #1337

Yamagi:
Thank you!

Issue #954, opened 2023-01-07
Soft renderer doesn't support PNG transparency


lafoxxx92:
Using custom /pics 
[zzz_psxhud.zip](https://github.com/yquake2/yquake2/files/10428668/zzz_psxhud.zip)

![изображение](https://user-images.githubusercontent.com/34402711/211167259-c9600175-108e-4e0c-85a0-ace12bf4ef55.png)
gl1

![изображение](https://user-images.githubusercontent.com/34402711/211167250-b19e3bc3-0dcd-4fcc-a1ba-2870517bc3ac.png)

soft

0lvin:
Have you checked with latest master branch code?

atsb:
![yq2_software_psxhud](https://github.com/yquake2/yquake2/assets/44937323/3f886853-ef7a-4f97-ba0b-f852a7492cf7)

Looks like it can be closed.  Latest master doesn't have this issue.

runlevel5:
I could not re-produce this issue with master code as of March 12 2025. I think the issue could be closed

PR #1339, opened 2026-07-16
GL3: Significantly reduce draw calls with batching

DanielGibson:
.. and some small fixes in the GL1 code that came up

In addition to the batching, GL3 now also has a scrap texture for 2D GUI textures, like GL1.

GL3 now has a `gl3_show_draw_stats` CVar that prints the amount of draw calls and glBufferData() calls.

before:  
q2_before

q2_0033

after:  
q2_0034

q2_0032

marking as a draft because it needs more testing and probably some more cleaning up

(please test!)

### DIFF
--- a/src/client/refresh/gl3/gl3_draw.c
+++ b/src/client/refresh/gl3/gl3_draw.c
@@ -251,7 +251,11 @@ GL3_Draw_CharScaled(int x, int y, int num, float scale)
 
 	scaledSize = 8*scale;
 
-	drawTexturedRectangle(draw_chars->texnum, x, y, scaledSize, scaledSize, fcol, frow, fcol+size, frow+size);
+	drawTexturedRectangle( draw_chars->texnum, x, y, scaledSize, scaledSize,
+	                       draw_chars->sl + fcol * (draw_chars->sh - draw_chars->sl),
+	                       draw_chars->tl + frow * (draw_chars->th - draw_chars->tl),
+	                       draw_chars->sl + (fcol + size) * (draw_chars->sh - draw_chars->sl),
+	                       draw_chars->tl + (frow + size) * (draw_chars->th - draw_chars->tl) );
 }
 
 // DG: copy of DrawStringScaled(), so we can draw some stats right here in the render DLL
--- a/src/client/refresh/gl3/gl3_draw.c
+++ b/src/client/refresh/gl3/gl3_draw.c
@@ -496,20 +496,26 @@ void GL3_EndFrame(void)
 {
 	// TODO: draw last 2D batch
 
+	// by saving those values into a variable and setting them to 0 afterwards,
+	// gl3_show_draw_stats can include its own drawcalls (from previous frame)
+	int num3D = gl3_num3Ddraws;
+	int num2D = gl3_num2Ddraws;
+	int numBufVtx = gl3_numBufferVtxData;
+	int numBufUni = gl3_numBufferUniforms;
+	gl3_num3Ddraws = 0;
+	gl3_num2Ddraws = 0;
+	gl3_numBufferVtxData = 0;
+	gl3_numBufferUniforms = 0;
+
 	if(gl3_show_draw_stats->value)
 	{
 		float factor = 2.0f; // TODO: like SCR_GetConsoleScale()
 		char stbuf[128] = {0};
 		snprintf(stbuf, sizeof(stbuf), "3D drawcalls: %d - 2D drawcalls: %d - buffer vtx data: %d - buffer uniforms: %d",
-		         gl3_num3Ddraws, gl3_num2Ddraws, gl3_numBufferVtxData, gl3_numBufferUniforms);
+		         num3D, num2D, numBufVtx, numBufUni);
 
 		GL3_DrawStringScaled(10, 5, stbuf, factor);
 	}
-
-	gl3_num3Ddraws = 0;
-	gl3_num2Ddraws = 0;
-	gl3_numBufferVtxData = 0;
-	gl3_numBufferUniforms = 0;
 
 	if(gl3config.useBigVBO)
 	{
--- a/src/client/refresh/gl3/gl3_draw.c
+++ b/src/client/refresh/gl3/gl3_draw.c
@@ -172,11 +172,6 @@ drawTexturedRectangle(GLuint texNum, float x, float y, float w, float h,
 	addIdx[3] = firstIdx+1; // second triangle
 	addIdx[4] = firstIdx+3;
 	addIdx[5] = firstIdx+2;
-
-	// FIXME: right now all this is still a bit buggy, e.g. the player model
-	// in the multiplayer menu is hidden when not enabling the next line..
-	// probably need to call drawLastBatch() before several other kinds of drawcalls
-	//drawLastBatch();
 }
 
 // bind the texture before calling this
@@ -193,6 +188,10 @@ drawTexturedRectangleNow(float x, float y, float w, float h,
 	 * sl,tl--------sh,tl
 	 *  x,y        x+w,y
 	 */
+
+	// in case some batched 2D draws are outstanding, draw them now
+	// to preserve draw order
+	GL3_DrawCurrent2Dbatch();
 
 	gl3_drawVert2D vBuf[4] = {
 	//    X,   Y,   S,  T
@@ -388,7 +387,6 @@ GL3_Draw_Fill(int x, int y, int w, int h, int c)
 		unsigned c;
 		byte v[4];
 	} color;
-	int i;
 
 	if ((unsigned)c > 255)
 	{
@@ -396,6 +394,10 @@ GL3_Draw_Fill(int x, int y, int w, int h, int c)
 	}
 
 	color.c = d_8to24table[c];
+
+	// in case some batched 2D draws are outstanding, draw them now
+	// to preserve draw order
+	GL3_DrawCurrent2Dbatch();
 
 	GLfloat vBuf[8] = {
 	//  X,   Y
@@ -405,7 +407,7 @@ GL3_Draw_Fill(int x, int y, int w, int h, int c)
 		x+w, y
 	};
 
-	for(i=0; i<3; ++i)
+	for(int i=0; i<3; ++i)
 	{
 		gl3state.uniCommonData.color.Elements[i] = color.v[i] * (1.0f/255.0f);
 	}
@@ -436,7 +438,9 @@ GL3_Draw_Flash(const float color[4], float x, float y, float w, float h)
 		return;
 	}
 
-	int i=0;
+	// in case some batched 2D draws are outstanding, draw them now
+	// to preserve draw order
+	GL3_DrawCurrent2Dbatch();
 
 	GLfloat vBuf[8] = {
 	//  X,   Y
@@ -448,7 +452,8 @@ GL3_Draw_Flash(const float color[4], float x, float y, float w, float h)
 
 	glEnable(GL_BLEND);
 
-	for(i=0; i<4; ++i)  gl3state.uniCommonData.color.Elements[i] = color[i];
+	for(int i=0; i<4; ++i)
+		gl3state.uniCommonData.color.Elements[i] = color[i];
 
 	GL3_UpdateUBOCommon();
 
--- a/src/client/refresh/gl3/gl3_draw.c
+++ b/src/client/refresh/gl3/gl3_draw.c
@@ -33,6 +33,8 @@ gl3image_t *draw_chars;
 
 static GLuint vbo2D = 0, vao2D = 0, vao2Dcolor = 0; // vao2D is for textured rendering, vao2Dcolor for color-only
 
+int gl3_num3Ddraws = 0, gl3_num2Ddraws = 0, gl3_numBufferVtxData = 0, gl3_numBufferUniforms = 0;
+
 void
 GL3_Draw_InitLocal(void)
 {
@@ -116,8 +118,9 @@ drawTexturedRectangle(float x, float y, float w, float h,
 	//       implicitly bind the vbo, so I need to explicitly bind it before glBufferData()
 	GL3_BindVBO(vbo2D);
 	glBufferData(GL_ARRAY_BUFFER, sizeof(vBuf), vBuf, GL_STREAM_DRAW);
-
 	glDrawArrays(GL_TRIANGLE_STRIP, 0, 4);
+	++gl3_numBufferVtxData;
+	++gl3_num2Ddraws;
 
 	//glMultiDrawArrays(mode, first, count, drawcount) ??
 }
@@ -158,6 +161,18 @@ GL3_Draw_CharScaled(int x, int y, int num, float scale)
 	GL3_UseProgram(gl3state.si2D.shaderProgram);
 	GL3_Bind(draw_chars->texnum);
 	drawTexturedRectangle(x, y, scaledSize, scaledSize, fcol, frow, fcol+size, frow+size);
+}
+
+// DG: copy of DrawStringScaled(), so we can draw some stats right here in the render DLL
+void
+GL3_DrawStringScaled(int x, int y, const char *s, float factor)
+{
+	while (*s)
+	{
+		GL3_Draw_CharScaled(x, y, *s ^ 0x80, factor);
+		x += 8*factor;
+		s++;
+	}
 }
 
 gl3image_t *
@@ -246,6 +261,9 @@ GL3_Draw_PicScaledCol(int x, int y, const char *pic, float factor, const float c
 void
 GL3_Draw_TileClear(int x, int y, int w, int h, const char *pic)
 {
+	if(w <= 0 || h <= 0)
+		return;
+
 	gl3image_t *image = R_FindPic(pic, (findimage_t)GL3_FindImage);
 	if (!image)
 	{
@@ -323,6 +341,9 @@ GL3_Draw_Fill(int x, int y, int w, int h, int c)
 	glBufferData(GL_ARRAY_BUFFER, sizeof(vBuf), vBuf, GL_STREAM_DRAW);
 
 	glDrawArrays(GL_TRIANGLE_STRIP, 0, 4);
+
+	++gl3_numBufferVtxData;
+	++gl3_num2Ddraws;
 }
 
 // in GL1 this is called R_Flash() (which just calls R_PolyBlend())
@@ -360,6 +381,9 @@ GL3_Draw_Flash(const float color[4], float x, float y, float w, float h)
 	glBufferData(GL_ARRAY_BUFFER, sizeof(vBuf), vBuf, GL_STREAM_DRAW);
 
 	glDrawArrays(GL_TRIANGLE_STRIP, 0, 4);
+
+	++gl3_numBufferVtxData;
+	++gl3_num2Ddraws;
 
 	glDisable(GL_BLEND);
 }
@@ -432,4 +456,38 @@ GL3_Draw_StretchRaw(int x, int y, int w, int h, int cols, int rows, const byte *
 	glDeleteTextures(1, &glTex);
 
 	GL3_Bind(0);
+}
+
+/*
+ * Called at the end of the frame, after 2D (UI) rendering is done.
+ * Does some internal housekeeping, then swaps the buffers
+ * and shows the next frame.
+ */
+void GL3_EndFrame(void)
+{
+	if(gl3_show_draw_stats->value)
+	{
+		float factor = 2.0f; // TODO: like SCR_GetConsoleScale()
+		char stbuf[128] = {0};
+		snprintf(stbuf, sizeof(stbuf), "3D drawcalls: %d - 2D drawcalls: %d - buffer vtx data: %d - buffer uniforms: %d",
+		         gl3_num3Ddraws, gl3_num2Ddraws, gl3_numBufferVtxData, gl3_numBufferUniforms);
+
+		GL3_DrawStringScaled(10, 5, stbuf, factor);
+	}
+
+	gl3_num3Ddraws = 0;
+	gl3_num2Ddraws = 0;
+	gl3_numBufferVtxData = 0;
+	gl3_numBufferUniforms = 0;
+
+	if(gl3config.useBigVBO)
+	{
+		// I think this is a good point to orphan the VBO and get a fresh one
+		GL3_BindVAO(gl3state.vao3D);
+		GL3_BindVBO(gl3state.vbo3D);
+		glBufferData(GL_ARRAY_BUFFER, gl3state.vbo3Dsize, NULL, GL_STREAM_DRAW);
+		gl3state.vbo3DcurOffset = 0;
+	}
+
+	GL3_SwapWindow();
 }
--- a/src/client/refresh/gl3/gl3_draw.c
+++ b/src/client/refresh/gl3/gl3_draw.c
@@ -27,13 +27,26 @@
 
 #include "header/local.h"
 
+#include "header/DG_dynarr.h"
+
 unsigned d_8to24table[256];
 
 gl3image_t *draw_chars;
 
-static GLuint vbo2D = 0, vao2D = 0, vao2Dcolor = 0; // vao2D is for textured rendering, vao2Dcolor for color-only
+static GLuint vbo2D = 0, ebo2D = 0, vao2D = 0, vao2Dcolor = 0; // vao2D is for textured rendering, vao2Dcolor for color-only
 
 int gl3_num3Ddraws = 0, gl3_num2Ddraws = 0, gl3_numBufferVtxData = 0, gl3_numBufferUniforms = 0;
+
+typedef struct gl3_drawVert2D_s {
+	float x, y, s, t;
+} gl3_drawVert2D;
+
+DA_TYPEDEF(gl3_drawVert2D, Vtx2DArray_t);
+DA_TYPEDEF(GLushort, UShortArray_t);
+// dynamic arrays to batch all consecutive 2D draws with same texture to reduce drawcalls
+static Vtx2DArray_t vtxBuf = {0};
+static UShortArray_t idxBuf = {0};
+static GLuint lastBatchTexture = 0;
 
 void
 GL3_Draw_InitLocal(void)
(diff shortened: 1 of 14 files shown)


0lvin:
Great news:-) Looks like it's solution for https://github.com/yquake2/yquake2remaster/issues/60 :-) I will try to merge it to remaster code. 

It already has scrap logic for gl3 with 32bit images as if user enabled retexturing, scrap will be unused. 

I will create PR with 32bit scrap later.

DanielGibson:
Yeah 32bit scrap makes sense, it's uploaded to the GPU as 32bit anyway

DanielGibson:
I just implemented 32bit scrap because I felt like it, sorry ;)

Should work with retexturing packs, but didn't test that yet (is there a q2 retexturing pack for UI textures?)

0lvin:
I have used attachment from https://github.com/yquake2/yquake2/issues/954 zzz_psxhud.pk3

DanielGibson:
yeah seems to work: 
q2_0036

(without scratch there would be more 2D drawcalls because of more different textures)

DanielGibson:
the build failures are github fuckups, checking out the code fails because some Github REST API used for that returns an error page instead

the log contains lots of raw HTML code, incl. base64 encoded images...

somewhere in the middle it says
```
      <p><strong>No server is currently available to service your request.</strong></p>
      <p>Sorry about that. Please try refreshing and contact us if the problem persists.<
```

DanielGibson:
rebased to latest master, added a fix for ARM (**signed** char, because there it's unsigned by default -_-)

on my RPi4 at 800x600 I now get 99fps instead of 77fps in a demo1.dm2 timedemo
(smaller differences at higher resolutions, I guess there the GPU rendering the shaders is the bottleneck)

on my desktop it's 3200fps instead of 1800fps (at fullhd or whatever)

0lvin:
What do think about extend renders API to have draw string additionally to draw chat? I have such in remaster just because rerelease has translation for strings, but it has additional benefit as provide ability to draw whole string by single call.

Maybe it will be useful, `gl3_show_draw_stats` could be replaced with `GALLIUM_HUD=draw-calls,fps` and so on with open-source linux drivers.

protocultor:
I've been looking for this since #1124! Glad you could make it @DanielGibson .
I was doing [my own version](https://github.com/protocultor/yquake2/tree/temporal) but messing around with dynamic arrays, shaders shenanigans, and especially real life, slowed me down.

This PR gave me almost a 40% perf improvement on desktop, but where this will shine is on SoCs; I wouldn't be surprised if some models achieve almost double the previous performance.

if you would be so kind, please include the following patch for GL1:
[drawing_order.patch](https://github.com/user-attachments/files/30132639/drawing_order.patch)
This is a fix for big HUD elements, textures that don't fit in the _scrap_, something that also happened to you. Found this problem by using _KMQuake2_'s HUD. Thanks.

EDIT: something to add, I just tested the PSX hud 0lvin mentioned, and it is a quicker alternative to test the patch.
Funny thing is, this made me realize the hud is fumbled on GL3.
Compare the picture above with how it appears in GL1:
q2_0000
This looks more appropiate. Also tested the KMQuake2 hud on GL3, and their numbers appear even bigger than PSX's.
EDIT 2: Current GL3 renderer on master show both huds in their correct proportions.

DanielGibson:
> What do think about extend renders API to have draw string additionally to draw chat? I have such in remaster just because rerelease has translation for strings, but it has additional benefit as provide ability to draw whole string by single call.

At least for GL3 this doesn't matter anymore, they'll be batched into one drawcall anyway

> Maybe it will be useful, gl3_show_draw_stats could be replaced with GALLIUM_HUD=draw-calls,fps and so on with open-source linux drivers.

but gallium won't show the difference between 2D and 3D draw calls, I think?

> this made me realize the hud is fumbled on GL3.

ok, looks like I need to investigate that, thanks :)

DanielGibson:
the PSX HUD proportions got broken with "GL3: Make scrap 32bit (convert 8bit in GL3_LoadPic())", so the problem is apparently not the batching but the scrap

maybe scrap + different texture size (retextured one bigger than original texture) doesn't cooperate

DanielGibson:
fixed

(UPDATE: Just checked @0lvin's #1340 and maybe I should've just used his 32bit scrap, this issue was already fixed there)

0lvin:
I have merged your code to remaster repository. It could have several bugs/regressions but It has improved fps for some maps like mgu2m1 from 0.5-1.5 fps to 50-60fps.

DanielGibson:
1.5fps?! on what device? what kind of map is that?

> but where this will shine is on SoCs; I wouldn't be surprised if some models achieve almost double the previous performance

The only SoC I have for testing is my **Raspberry Pi 4**, and it wasn't that great, so I investigated a bit further just to figure out what's making it slow.
To be clear, I won't do anything about it because I have no better solutions, I just disabled some code (mostly in the `fragmentSrc3Dlm` shader) to see what costs performance, but doing so of course made the game look wrong.

My test was running YQ2 at 1280x720 on current Raspberry Pi OS, windowed mode on a 2440x1440 desktop, with the `gles3` renderer. There I ran `demomap demo1.dm2` with `timedemo 1`, 3-5 times in a row and then wrote down the best result.

On the unmodified `master` branch, I got 56fps.
On the unmodified `gl3-improvements` branch, I got 59fps.

I wondered if it would help to disable/ignore lmscale updates - doing so reduces drawcalls in some scenes (e.g. in base1 when looking at the first doorway with the flickering light, 3D drawcalls were reduced from ~160 to ~40). Not really, *maybe* 1fps difference, at least in demo1.dm2 - no matter if I do this change on an otherwise unmodified gl3-improvements-state or a state that runs at ~92fps by disabling dynlights and reading only the first lightmap.
I'm just mentioning this here so it gets mentioned, it's the only optimization that was outside of the shader and not also tested on `master`. The code change for this looked like:
<details>

```diff
diff --git a/src/client/refresh/gl3/gl3_main.c b/src/client/refresh/gl3/gl3_main.c
index f45757f6..7ab61b93 100644
--- a/src/client/refresh/gl3/gl3_main.c
+++ b/src/client/refresh/gl3/gl3_main.c
@@ -832,6 +832,7 @@ void GL3_Draw3DBatchesNow()
                        updateUni3D = true;
                }
 
+#if 0
                if((flags & DCFlag_UseLmStyles) && !lmstylesEqual(cmd->styles, curState.styles))
                {
                        hmm_vec4 lmScales[MAX_LIGHTMAPS_PER_SURFACE] = {0};
@@ -845,6 +846,7 @@ void GL3_Draw3DBatchesNow()
                        }
                        UpdateLMscales(lmScales, shader);
                }
+#endif
 
                if(flags & DCFlag_UseColor)
                {
@@ -922,8 +924,8 @@ static qboolean drawStateEqual(const gl3drawCmd_t* a, const gl3drawCmd_t* b)
        if(a->alpha != b->alpha)
                return false;
 
-       if((flags & DCFlag_UseLmStyles) && !lmstylesEqual(a->styles, b->styles))
-               return false;
+       //if((flags & DCFlag_UseLmStyles) && !lmstylesEqual(a->styles, b->styles))
+       //      return false;
 
        return true;
 }
```

</details>

---

So, about changes that did make a difference, all in `fragmentSrc3Dlm` in gl3_shaders.c:

- Disabling dynamic light calculations in the shader (`if(false)` instead of `if(passLightFlags != 0u)`)
- Reading only the first lightmap instead of all 4 (comment out `lmTex     += texture(lightmap1, passLMcoord) * lmScales[1];` and the same lines for `lightmap2 and lightmap3`)
- Reading no lightmaps at all (in addition to previous change, `vec4 lmTex = vec4(1.0,1.0,1.0,0.0);` instead of `vec4 lmTex = texture(lightmap0, passLMcoord) * lmScales[0];`), makes the world look fullbright

I tested those changes with both `master` and the current `gl3-improvements` state:

Changes to the shader | master | gl3-impr.
----------------------|--------|------
None of these (all lightmaps, dynlights enabled), the default state | 56 fps | 59 fps
Only first lightmap, dynlights enabled | 62 fps | 67 fps
No lightmaps, dynlights enabled        | 62 fps | 68 fps
All lightmaps, dynlights **disabled**  | 66 fps | 77 fps
One lightmap, dynlights disabled       | 76 fps | 92 fps
No lightmaps, dynlights disabled       | 77 fps | 104 fps

So the dynamic lights have the biggest impact with both the old and the new code.
Without dynamic lights, the the number of lightmaps read starts having a bigger impact and so do the optimizations of this PR.

0lvin:
> 1.5fps?! on what device? what kind of map is that?
> 

ReRelease maps often are huge, have 4k+ coordinates and lacks visibility lump or whole map is visible by bsp from start point and so on.

Example of bug report: [Very bad performance on some maps in Call of The Machine](https://github.com/yquake2/yquake2remaster/issues/60)

protocultor:
@DanielGibson I recommend you to watch the Digital Foundry presentation for the Quake 2 rerelease:
https://youtu.be/ck3pqrTAphs
Or at least, see just between:
- 13:16 and 13:49, for a small tech intro (_no BSP used!_)
- 22:15 and 23:41, for the reason of the tech modifications

That will make clear the low numbers 0lvin gets on the _Call of the Machine_ maps; these are not meant to be rendered by this engine in the traditional way.

> The only SoC I have for testing is my **Raspberry Pi 4**, and it wasn't that great, so I investigated a bit further just to figure out what's making it slow.

I forgot you already batched draw calls for the meshes; yeah, performance improvement won't be as noticeable as in GL1.

About the lightmaps, I just can share what I did for GL1, which was forcing to **change them only once for each frame**, and no more; the improvement after doing this was ridiculous. Maybe some shaders are being called more than once needlessly, over the same brushes/meshes?

See function `R_RegenAllLightmaps()` here (changed a lot today, but its objective remains the same):
https://github.com/yquake2/yquake2/commit/ba71af2af83fc2536cc9e773d0b925baa0cf02a3#diff-867f66295166e68d13ad6d73613c300dd577fdd2338cfdebfd02c896d8aaa350R743-R829
...which does a single call to `glTexSubImage2D()` for all the changes in every lightmap present in the lightmap chain.

But also `R_GetBrushesLighting()` here (also changed today):
https://github.com/yquake2/yquake2/commit/4503333b3a5c170a1388066a40d9c82daea35053#diff-867f66295166e68d13ad6d73613c300dd577fdd2338cfdebfd02c896d8aaa350R1195-R1284
...which gathers all the changes to the lightmaps for the brushes in the entity list and adds the lightmap to the lightmap chain, so `R_RegenAllLightmaps()` can do its magic later.

Compare to Vanilla Q2, where after drawing the lightmaps, there were additional changes to the "dynamic lightmap" + drawing this dynamic lightmap on top of current brushes, which is unoptimal, to say the least. I'm too dumb to follow your code, but maybe you are treating the entity list' lighting the same way?

About your RPi4 numbers, I wanted to compare with my RPi3 numbers on GL1, using `gl1_lightmapcopies 1` and `gl1_discardfb 1` (cvars for SoC optimizations) with the game at 1280x720 resolution:

- RaspiOS Trixie over *KMSDRM* (console, no X11 or Wayland), fullscreen: 138.7 fps
- RaspiOS Bookworm on a *labwc* desktop (Wayland) at 1080p, window: 64.3 fps.

Both are "first attempt" only, numbers improve by just a little on repeated attempts.

Note that GL1 doesn't allocate memory for batching anything. Maybe dynamic array allocation can be improved?
Can you give GL1 / GLES1 a try on your RPi4?

DanielGibson:
I don't think the dynamic arrays can be improved much - they are just *cleared* when done, which only resets the internal count to 0 and does not free memory, so the memory is reused next time. Only if the dynamic array is too small to hold the current number of elements, new memory is allocated (with realloc()).

I'll have to try gl1 on RPi4.

@0lvin you could try if disabling the lmscale updates to reduce draw calls (as I showed in my post above but didn't help much in vanilla q2) helps in those insane remaster levels.
If that really makes a difference, we could try putting those lmscales into a uniform (it's just 256 vec4s) that's only updated once per frame, and pass the lmscales as 4 bytes per vertex to avoid frequent uniform updates and allow batching those drawcalls.

> About the lightmaps, I just can share what I did for GL1, which was forcing to change them only once for each frame, and no more; the improvement after doing this was ridiculous. 

IIRC Q2 has 4 lightmap textures and then those lmscales/lmstyles that select per surface which one(s) to use with which factor. I think the GL1 renderer merges them into one texture that is updated each frame, and also bakes the dynamic lights into that merged lightmap.

In GL3 I don't update the lightmap textures at all.
Dynamic lights are calculated in the pixelshader instead of being baked into a lightmap, so they have a much higher resolution and don't require updating any textures.
All 4 lightmaps are passed into the shader, and so are the 4 lmscales, and the shader then merges them: https://github.com/yquake2/yquake2/blob/e3289e558c43d3999ac75a8abd996494acbc145a/src/client/refresh/gl3/gl3_shaders.c#L593-L597

Maybe at least on crappy GPUs merging the lightmaps would be faster, probably even doing dynamic lights in the merged lightmap like in GL1 would make sense for them.
But then again, one could as well use the GL1 renderer on those devices...

protocultor:
> IIRC Q2 has 4 lightmap textures and then those lmscales/lmstyles that select per surface which one(s) to use with which factor. I think the GL1 renderer merges them into one texture that is updated each frame, and also bakes the dynamic lights into that merged lightmap.

That's correct; the important part is that the "consolidated lightmap" is now updated only once every frame, not multiple times like before.
I didn't mean you to do the same as GL1; `glTexSubImage2D()` is the slowest thing in life, and the first thing to do if you're chasing performance is NOT to use it (like _Quake 3_ did, only 2 years after _Quake 2_). I just mentioned this to give you ideas on how to avoid "reprocessing" lighting constantly, which is always a pain.
Like, what you mention about "_putting lmscales into a uniform_" sounds like something worth trying...

DanielGibson:
> Like, what you mention about "putting lmscales into a uniform" sounds like something worth trying...

Well, that's the question. It does further reduce drawcalls, but in my tests that didn't give any further performance advantage (in demo1.dm2 on RPi 4) - I tested it by just not updating the lmscales and using `lmscales[0] = (1,1,1,1);` and `0,0,0,0` for lmscales[1] to get at least basic lightmaps.

But maybe other maps/scenes than what demo1 shows, like maybe those insane remastered maps, benefit more.

It could theoretically also be that those changes would make things slower because of the indirect lookup (`texture(lightmap0, passLMcoord) * lmScales[0]` becomes `texture(lightmap0, passLMcoord) * lmStyles[lmScaleIdx[0]]`), no idea how much GPUs like that.

0lvin:
> @0lvin you could try if disabling the lmscale updates to reduce draw calls (as I showed in my post above but didn't help much in vanilla q2) helps in those insane remaster levels. If that really makes a difference, we could try putting those lmscales into a uniform (it's just 256 vec4s) that's only updated once per frame, and pass the lmscales as 4 bytes per vertex to avoid frequent uniform updates and allow batching those drawcalls.
>

Thank you, I will look.

Could you please look https://github.com/yquake2/yquake2remaster/commit/b88f33c1c3a3215ab09ff8b1cdaee10c65e7151f? Remaster repository has showtris implementation for gl3, maybe it will be useful for your testing.

DanielGibson:
I've implemented that lmscales change, no idea if it helps, on my desktop it doesn't seem to make a big difference (but framerates for demo1 vary between 3200 and 3700 fps anyway between consecutive runs)

But maybe it helps with q2 remastered or on other GPUs

UPDATE: On my laptop the latest commit seems to reduce the performance a bit, but there the results also fluctuate a lot - but it seems like without that change I usually get around 200fps (10%) more

DanielGibson:
On RPi4, the latest commit costs about 10fps in demo1 and demo2 - I think I'll remove it again (will keep it here for a bit so @0lvin can get it to test with q2remaster, maybe it helps there).

I also tested GL1 at 720p.
- gl1_lightmapcopies 0, gl1_discardfb 0 (default state): 79fps
- gl1_lightmapcopies 0, gl1_discardfb 1: 95fps
- gl1_lightmapcopies 1, gl1_discardfb 0: 109fps
- gl1_lightmapcopies 1, gl1_discardfb 1: 126fps

protocultor:
> On RPi4, the latest commit costs about 10fps in demo1 and demo2 - I think I'll remove it again (will keep it here for a bit so @0lvin can get it to test with q2remaster, maybe it helps there).

Shame about it; don't let it discourage you to try other things.

> I also tested GL1 at 720p.

Those are good numbers for a desktop environment; I guarantee you they will almost double outside _Wayland_/_X11_.
`gl1_lightmapcopies` is a trick to avoid `glTexSubImage2D()`'s adverse behavior on SoCs, but you might want to take a look at what `gl1_discardfb` does, it'll probably help performance in GLES3 on the RPi4.

0lvin:
With yesterday build with enabled debug I had:
Soft: 3.64fps,
Gl4: 0.2fps, it's gl3 code mostly but without 3d draw combine,
Gl3: 14fps, 8k+ 3d draws,
Gl1: 13fps.

On start point of warehousejam1_start. 

For reproduce could be used https://www.moddb.com/mods/quake-2-re-release-warehouse-jam-1 with remaster code. ReRelease assets are not required if you are ok with missing textures.

0lvin:
Results with latest code in remaster:
* demo1.dm2 177fps average,
* warehousejam1_start start point 16.98fps 7858 3d draws

Results with latest + gl3-improvements(58f40178df41d4544c85b5b232dccc42eec23e8f)
* demo1.dm2 165.5fps
* warehousejam1_start start point 28.50fps 339 3d draws

IMHO: I prefer twice less fps but more shared code between renders and search way how to combine render commands in future and reuse between renders :-)

protocultor:
I wanted to mention: if you have normal/classic Quake 2 on either Steam or GOG, you have the remaster as well, for free.
Which likewise runs on Proton, so you don't require Windows to run it.

DanielGibson:
I implemented discard framebuffers.
After (locally) reverting that "GL3: lightmap styles as indices in vertex attributes" commit, enabling discard improves the framerate in demo1.dm2 from around 58fps to around 69fps.

Unlike in GL1 discarding the color buffer had to be done after swapbuffers, otherwise it rendered garbage on the RPi4 - probably makes sense, after all the color buffer is what's getting rendered to the screen?
It actually was fastest to discard stencil- and depthbuffer before swapbuffers and the colorbuffer afterwards, instead of all afterwards (I think about 4fps more)

src/client/refresh/gl3/gl3_draw.c
```diff
@@ -79,17 +95,91 @@
...
+	if((lastBatchTexture != 0 && texNum != lastBatchTexture) || da_count(vtxBuf)+4 > UINT16_MAX)
+		GL3_DrawCurrent2Dbatch();
+
+	lastBatchTexture = texNum;
+
+	GLushort firstIdx = da_count(vtxBuf);
+
+	gl3_drawVert2D* addVtx = da_addn_uninit(vtxBuf, 4);
+	//                            X,   Y,   S,  T
+	addVtx[0] = (gl3_drawVert2D){ x,   y+h, sl, th };
```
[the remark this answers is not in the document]
DanielGibson: there is no nullptr.
if allocating memory fails, the process will be terminated (see `#define DG_DYNARR_ASSERT(cond, msg) ...` in local.h; on out of memory DG_DYNARR_ASSERT() is called).

DanielGibson:
I removed the "GL3: lightmap styles as indices in vertex attributes" commit, this should be ready for review/final testing/merging

Srandista:
Wow, the performance difference in certain situations is absolutely brutal. I tried this in my VM and the results were insane (30 fps vs 300 fps looking at the skybox on map mgu2m1 or literal seconds per frame vs 17 fps looking behind my starting position on mgu2m1). RPi5 also saw a big performance bump (5 fps vs 33 fps looking behind my starting position on mgu2m1).

Yamagi:
I have played through the first unit and checked some known to be difficult places in later maps:

* The pool in `jail1`
* The black hole generator in `command`
* The moveable brush next to the door to the forcefield controls in `city3`

So far everything looks fine. I'll do some more testing with the addons over the next few days.

This also helps a lot with Zink. Master reaches 2604 FPS on my overpowered Radeon 9070XT, Daniels branch 6530 FPS... For comparison the ref_vk renderer yields 3293 FPS.

0lvin:
> This also helps a lot with Zink. Master reaches 2604 FPS on my overpowered Radeon 9070XT, Daniels branch 6530 FPS... For comparison the ref_vk renderer yields 3293 FPS.

Could you please also check remaster repo with vulkan?

Yamagi:
> Could you please also check remaster repo with vulkan?

With remaster I get 3204 FPS.

But Daniel pointed out that on his laptop Zink is slower than the native radeonsi Mesa backend and I can reproduce it. I get 1195 FPS native and only 997 FPS with Zink on an Ryzen AI 350.

0lvin:
> > Could you please also check remaster repo with vulkan?
> 
> With remaster I get 3204 FPS.
> 
> But Daniel pointed out that on his laptop Zink is slower than the native radeonsi Mesa backend and I can reproduce it. I get 1195 FPS native and only 997 FPS with Zink on an Ryzen AI 350.

Thank you! 90fps drop, looks like we still have room to improve :-)

Yamagi:
Arg, that was a typo. It's 997 FPS.

protocultor:
Agree with the rest of the comments; your latest commits have made a notorious change for the better. Nice job!

> Unlike in GL1 discarding the color buffer had to be done after swapbuffers, otherwise it rendered garbage on the RPi4 - probably makes sense, after all the color buffer is what's getting rendered to the screen? It actually was fastest to discard stencil- and depthbuffer before swapbuffers and the colorbuffer afterwards, instead of all afterwards (I think about 4fps more)

I wanted to say that your `gl_discardfb` work is good, but missing something added in #1152: the ability to call `glClear()` with a black color at the beginning of every frame. This is the reason why it works with GL1 on the RPi4 (without the need for GLES1 / `glDiscardFramebufferEXT`), but is also probably the reason of the problem you're describing.
Additionally, `glDiscardFramebufferEXT()` exists as a core functionality in OpenGL ES 3.0, as [glInvalidateFramebuffer()](https://registry.khronos.org/OpenGL-Refpages/es3.0/html/glInvalidateFramebuffer.xhtml).

Please give a try to this patch on your RPi4:
[gles3_clear.patch](https://github.com/user-attachments/files/30278626/gles3_clear.patch)
...and if it works better than the current state, remove `glDiscardFramebufferEXT` from your _glad_ installation :)

DanielGibson:
> Please give a try to this patch on your RPi4:
> [gles3_clear.patch](https://github.com/user-attachments/files/30278626/gles3_clear.patch)
> ...and if it works better than the current state, remove glDiscardFramebufferEXT from your glad installation :)

yeah, it's up to around 76fps now (from 69, which was nicer though)

why does clearing the color make a performance difference after discarding the color buffer? O_o

protocultor:
> why does clearing the color make a performance difference after discarding the color buffer? O_o

These links give simple, concise explanations to that:
https://stackoverflow.com/q/15506110
https://stackoverflow.com/q/2538662 (first two answers only, second is interesting)

About why to keep using `glInvalidateFramebuffer()`:
https://stackoverflow.com/q/59892796

Must say that these links speak about _Tile-Based Rendering_ GPUs (mobile / embedded), not _Immediate-Mode Rendering_ ones (traditional desktop cards), which behave differently.

I was thinking: since you are not using it in GL3, and it is being useful in GLES3, it is too crazy to delete the `gl_discardfb` cvar and force an "_always 1_" bahavior in GLES3? It doesn't make much sense to disable it in GLES3-only devices. It's good to have this "switch" in GL1 since many devices like the RPi have taken out support for GLES1 but offer support for GL1, but this is not the case with GL3/GLES3.

0lvin:
I have released new version of remaster with your code, 
https://github.com/yquake2/yquake2remaster/releases/tag/8.71RR15. it should resolve performance issues, thank you!

Yamagi:
As discussed, I've merged the #1345 and #1340. I'll do some more testing with this one later this week. Afterwards (and when the conflicts are resolved, of course) I'll merge it.

0lvin:
> As discussed, I've merged the #1345 and #1340. I'll do some more testing with this one later this week. Afterwards (and when the conflicts are resolved, of course) I'll merge it.

Merged version with resolved conflicts https://github.com/0lvin/yquake2/commit/484c1536c23945f1ce732e1c6e03bcfba3d45265 / https://github.com/0lvin/yquake2/tree/gl3-improvements

@DanielGibson do you like to have PR for your repository?

Update: I have created https://github.com/DanielGibson/yquake2/pull/7

Yamagi:
I did some more testing. Everything seems to be fine, including the always problematic negative light in rogue. @DanielGibson will look for some more improvements, afterwards we are ready to go.

DanielGibson:
I rebased this to the current master branch and did some cleanups on the way (squashed some commits, replaced GL_EXT_discard_framebuffer with @protocultor's patch to use the GLES3-native glInvalidateFramebuffer(), ...)

DanielGibson:
failing mingw32 build is because of 

```
C:\Windows\system32\cmd.exe /D /S /C D:\a\_temp\setup-msys2\msys2.cmd -c "'pacman' '--noconfirm' '-S' '--needed' '--overwrite' '*' 'git' 'make' 'zip' 'unzip' 'mingw-w64-i686-curl' 'mingw-w64-i686-gcc' 'mingw-w64-i686-make' 'mingw-w64-i686-openal' 'mingw-w64-i686-pkgconf'"
  error: target not found: mingw-w64-i686-openal
```

i.e. not my fault

Yamagi:
> i.e. not my fault

Crap, they don't publish the `mingw-w64-i686-openal` package anymore.

Yamagi:
I've played two levels and no regressions so far. I no one complains I'll merge this in the next few days.

src/client/refresh/gl1/gl1_image.c
```diff
@@ -439,6 +439,8 @@ R_ImageList_f(void)
 
 		char imageType = '?';
 
+		Com_Printf(image->scrap ? "S" : " ");
```
0lvin: Will write 'S' twice, as 'S' in isScrap also.

DanielGibson: oh, right - when I originally made this change it didn't, I missed that you modified that function as well

fixed

Yamagi:
Thank you all! :)

PR #1341, opened 2026-07-21
menu: API refactoring and more bug fixes and stability

BjossiAlfreds:
This PR is a 3rd round of stability improvements and fixes for the menu system. Also ncludes some API refactoring.

Simplified menu layer code
* Replaced `m_active` variable with `M_GetActiveMenu` function
* Removed `menulayer_t` struct and made `m_layers` an array of `menuframework_s *`

menuframework_s refactoring
* Added `banner` string member
* `->draw` and `->key` callbacks now take a self-pointer as argument
* Removed a large amount of `*_MenuKey` and `_MenuDraw` in favor of reworked `Default_MenuKey` and `Default_MenuDraw` functions
* Added `Default_MenuDraw` that provides baseline menu drawing behavior (banner, menu and popup)
* `Default_MenuKey` is now called directly by the callback and it includes popup closing
* Added a `->close` callback which is now used by credits, address book and player config menus. This callback is called from within `M_ForceMenuOff` and `M_PopMenu` functions

Popup API
* Added `menupopup_s` struct and `m_popup` variable
* Reworked `M_Popup` into `Menu_DrawPopup` and added `Menu_StartPopup` functions
* Added `Menu_PopupActive` and `Menu_ClosePopup` macros
* Moved some utility code to `qmenu.c`

Misc refactoring
* Delete and backspace are no longer treated the same in `Key_GetMenuKey`
* Added `case K_DEL:` to keep old behavior in key binding / savegame / etc. menus
* Start server menu elements now positioned relatively rather than absolutely and `capture limit` option now pushes down instead of up
* Moved `StripExtension` to shared, renamed it `COM_StripExtension2` and made it smarter and return the extension part
* Moved `ContainsFile` to filesystem, renamed it `FS_FileExists` and made it more robust and path arg can be NULL
* Removed `ReplacedCharacters` and added `Q_strrchrs` to shared that looks for the last instance of one of the specified characters

Stability
* `Field_InitBuffer` function for setting up input field state
* Make sure `menu->nitems >= 0` in `Menu_AddItem`
* `f` is now initialized to 0 at the start of `Fs_FOpenFile` to ensure it is defined after returning

Bug fixes
* Fixed possible resource leaks or discarded changes if a menu is closed without pressing Escape
* Fixed delete key not working correctly in input fields. It was behaving like backspace
* Fixed credits menu playing a louder menu closing sound. It was calling `S_StartLocalSound` twice
* Fixed popup state not being cleared when menu is forced off
* (maybe a bug) Fixed start server menu banner missing
* Fixed `Q_strchrs` not always returning the earliest instance of one of the specified characters

Since this is refactoring core API logic, more people testing this would be welcome.

I might get more ideas for refactoring or discover more issues, but what is currently here seems stable during my testing.

### DIFF
--- a/src/client/menu/header/qmenu.h
+++ b/src/client/menu/header/qmenu.h
@@ -65,10 +65,11 @@ typedef struct _tag_menuframework
 	int nitems;
 	void *items[64];
 
+	const char *banner;
 	const char *statusbar;
 
-	void (*draw)(void);
-    	const char *(*key)(struct _tag_menuframework *m, int k);
+	void (*draw)(struct _tag_menuframework *m);
+	const char *(*key)(struct _tag_menuframework *m, int k);
 	void (*cursordraw)(struct _tag_menuframework *m);
 } menuframework_s;
 
--- a/src/client/menu/header/qmenu.h
+++ b/src/client/menu/header/qmenu.h
@@ -153,10 +153,25 @@ qboolean Menu_SelectItem(menuframework_s *s);
 void Menu_SetStatusBar(menuframework_s *s, const char *string);
 qboolean Menu_SlideItem(menuframework_s *s, int dir);
 
+void Menu_DrawCharacter(int cx, int cy, int num);
 void Menu_DrawString(int, int, const char *);
 void Menu_DrawStringDark(int, int, const char *);
 void Menu_DrawStringR2L(int, int, const char *);
 void Menu_DrawStringR2LDark(int, int, const char *);
+
+typedef struct
+{
+	const char *string;
+	int endtime;
+} menupopup_s;
+
+void Menu_DrawTextBox(int x, int y, int width, int lines);
+void Menu_DrawText(int x, int y, const char *str);
+void Menu_DrawPopup(int x, int y, const menupopup_s *pup);
+void Menu_StartPopup(menupopup_s *pup, const char *string, int lifetime);
+
+#define Menu_PopupActive(pup) ((pup)->string && ((pup)->endtime <= 0 || (pup)->endtime >= cls.realtime))
+#define Menu_ClosePopup(pup) (pup)->endtime = cls.realtime - 1
 
 float ClampCvar(float min, float max, float value);
 
--- a/src/client/menu/header/qmenu.h
+++ b/src/client/menu/header/qmenu.h
@@ -68,7 +68,7 @@ typedef struct _tag_menuframework
 	const char *statusbar;
 
 	void (*draw)(void);
-    	const char *(*key)(int k);
+    	const char *(*key)(struct _tag_menuframework *m, int k);
 	void (*cursordraw)(struct _tag_menuframework *m);
 } menuframework_s;
 
--- a/src/client/menu/header/qmenu.h
+++ b/src/client/menu/header/qmenu.h
@@ -142,6 +142,7 @@ typedef struct
 
 void M_PushMenu(menuframework_s* menu);
 
+void Field_InitState(menufield_s *f, const char *s, int len, int vislen);
 void Field_ResetCursor(menuframework_s *m);
 qboolean Field_Key(menufield_s *field, int key);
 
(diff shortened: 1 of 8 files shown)


Yamagi:
Wow. This is essential what I had in mind when I wrote that infamous comment stating that the menu used should be rewritten. And what I never even tried...

> Since this is refactoring core API logic, more people testing this would be welcome.

I've clicked through all menus and options. I didn't find any problems so far. I'll have a deeper look at the weekend.

BjossiAlfreds:
If you have any more rewrite ideas I could take a look. I really want to clean up player config code for example, it is really messy to put it mildly. That would probably be a separate PR though.

Also, I made an experimental commit, giving the start server menu a previously unused banner `m_banner_start_server`. See what you think. I had to make the ctf `capture limit` box push down instead of up but I don't think anyone would notice the difference.

Yamagi:
> If you have any more rewrite ideas I could take a look.

No, not really. The first part would have been a general cleanup and refactoring. That is more or less done right now, by some previous pull requests and in depth by this one. The second I part envisioned would have been some scripting system which allows mods to define their own menus. That is out of scope, scripting is way to much work and and it would likely need changes to API between game and server. And maybe even some network protocol changes.

Yamagi:
> Also, I made an experimental commit, giving the start server menu a previously unused banner m_banner_start_server. See what you think. I had to make the ctf capture limit box push down instead of up but I don't think anyone would notice the difference.

I wouldn't have noticed if I didn't know it. The same goes for the slightly different quality of the banners. 'multiplayer' <-> 'join server' <-> 'start server' all look a little bit different. I guess this is totally okay and since id Software shipped the 'start server' we can use it.

BjossiAlfreds:
Menu scripting is something I have only kind of dreamed about, because as you said it would be a lot of work and need API / net protocol changes (if we want to let game lib invoke custom menus).

Even though I don't have the eyesight to see any differences between the banners, this is something that crossed my mind, that maybe the start server banner is unfinished / low quality. But didn't consider the other banners also had differences between each other.

Yamagi:
No, no, it doesn't look better or worse than the other. It's hard to see, look at the id Software logo to the right: 'multiplayer' has a slightly different scale than 'join server' and 'start server'.

BjossiAlfreds:
Oh I see it now. The Q2 logo on the left side also seems to vary slightly.

BjossiAlfreds:
Just found and fixed another small menu bug: Text popups were not closed by `M_ForceMenuOff`.

Example:
1. Go to options menu
2. Press Enter on sound backend option to open the popup
3. Quickly open the console and return to the options menu
4. If you are fast enough, the popup would still be there

Yamagi:
>Just found and fixed another small menu bug: Text popups were not closed by M_ForceMenuOff.

Now that I know it it's fairly easy to reproduce -_-

Is this PR ready?

BjossiAlfreds:
Yeah I'd say so.

0lvin:
> If you have any more rewrite ideas I could take a look. I really want to clean up player config code for example, it is really messy to put it mildly. That would probably be a separate PR though.
> 
> Also, I made an experimental commit, giving the start server menu a previously unused banner `m_banner_start_server`. See what you think. I had to make the ctf `capture limit` box push down instead of up but I don't think anyone would notice the difference.

Just some idea to implement in menu: ability to select start map for single player, code has map select in menu for multiplayer, but not for single player. For ReRelease it  makes more sense as player could select expansion/level, and for classic game could be selected map using names from maps.lst or [maps.json](https://github.com/yquake2/yquake2remaster/commit/b19aac3e7c08d7bc4875a4b0c67529a6ba16d2e5). What do you think?

BjossiAlfreds:
Would players find that useful though? Quake 2 maps aren't really designed for cold starts except the first map of each unit. Starting mid-unit can even require cheats to progress at certain points.

Another problem is neither maps.lst nor the map itself says whether a map is a DM map, a campaign map or both. We would have to add a hard-coded filter, look for the letters `dm` in the map filename or just have DM maps inside the list.

I think the best that could be done is a *start from* spin control, where the available choices could be *Beginning*, *Unit 2: Warehouse*, *Unit 3: Jail*, etc.

Maybe you could open an issue about this where we can discuss and share ideas?

In any case, this PR is ready for a merge as far as I know.

Yamagi:
> In any case, this PR is ready for a merge as far as I know.

Thank you!

PR #1343, opened 2026-07-21
menu: Fix menu sounds not playing in SDL mode when a game session is active

### LINKED ISSUE #1342: Main menu sounds does not playing when there is demo playing and my audio backend is set to SDL

BjossiAlfreds:
Fixes https://github.com/yquake2/yquake2/issues/1342

`S_StartLocalSound` uses an `entnum` of -1 to indicate a sound event that is not spatial, such as menu sounds. This is a regression I introduced last year when working on BSP sound positioning, if I remember correctly.


Yamagi:
Thank you :)

PR #1351, opened 2026-08-05
workflows: disable openal for win32

0lvin:
* Build artifacts are not unused for final release, disable OpenAL for win32 should be safe, as OpenAL also used in win64 and should catch possible regressions.
* Update SDL3 and workflow dependencies. 
* Add skip system include check for cppcheck
* Put crosschar model to build artifacts.
* Build debug version additionally to release version.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -38,11 +38,11 @@ jobs:
       run: |
         sed -i 's|option(SDL3_SUPPORT "Build against SDL 3 instead of SDL2" ON)|option(SDL3_SUPPORT "Build against SDL 3 instead of SDL2" OFF)|g' CMakeLists.txt
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v4
+      uses: github/codeql-action/init@v4.37.4
       with:
         languages: ${{ matrix.language }}
         build-mode: ${{ matrix.build-mode }}
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@v4
+      uses: github/codeql-action/analyze@v4.37.4
       with:
         category: "/language:${{ matrix.language }}"
--- a/.github/workflows/linux_aarch64.yml
+++ b/.github/workflows/linux_aarch64.yml
@@ -40,14 +40,20 @@ jobs:
         # Public runners come with 4 CPUs.
         make WITH_SDL3=no -j4
         make WITH_SDL3=no -j4 ref_gles1
+        # Debug version
+        make WITH_SDL3=no DEBUG=1 -j4
+        make WITH_SDL3=no DEBUG=1 -j4 ref_gles1
     - name: Create testbuild package
       run: |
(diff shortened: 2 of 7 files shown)


Yamagi:
Thank you.

I'll need to think about the OpenAL problem. Maybe something similar to what we are doing with SDL3, download it from somewhere(tm) and teach the Makefile were to find it.

j4reporting:
that's upstream  and not affiliated with msys2 at all.

msys2  is  here    https://packages.msys2.org/search?t=binpkg&q=openal

btw:  fedora: 
```
$ dnf list mingw32-openal\* mingw64-openal\*
Available packages (available for reinstall, available for upgrade)
mingw32-openal-soft.noarch 1.25.0-2.fc44 fedora
mingw64-openal-soft.noarch 1.25.0-2.fc44 fedora
```
Does ubuntu also provide 3rd-party mingw packages?  That could be a viable alternative for cross-compiling  32bit and perhaps also 64bit windows binaries.

protocultor:
Yeah, didn't see the MSYS2 packages, apologies for that.

Yamagi:
The problem here is MSYS2 dropping 32 bit support. They announced it over 3 years ago and so far my "strategy" to ignore it was successful. ;)

I still think that we should continue to ship 32 bit binaries for Windows as long as possible. Switching to 64 bit would break compatibility with all mods that come with their own game dll. And most mods never released their source code and were the source code is available it's often in really bad shape.

For a short term solution we could do the same as we did for SDL3. Put the official OpenAL package to /opt and implement `OPENALCFLAGS` in the Makefile to tell the compiler about the include directory. And maybe do the same for curl before they also drop the curl package.

So far I can think of three ways to a longer term solution:

* Switch to another mingw-w64 distribution, maybe something smaller and less complicated than msys2. We don't need much beside the compilers and linkers. SDL, OpenAL and curl can be taken from their vendors, that leaves `make` and `windres`.
* Do something with cross compilation with mingw-w64 from Linux. I have no experience with it, may work reliable or may introduce interesting problems. Maybe not something we want for release builds.
* Switch roles between mingw-w64 and MSVC, declare MSVC as the preferred way to build Windows binaries and demote mingw-w64 to the alternative way. Will require some work. The CMakeLists.txt may not be in the best shape and the addons were never ported to modern MSVC. There shouldn't be to much changes needed, though.

Another point is backward compatibility to older Windows versions. The main reason for mingw-w64 was the ability to build the the Windows binaries against the classic MSVCRT shipped with older Windows version. They should work on Windows XP and newer out of the box, without installing some C++ runtime. 

Switching to URTC - either by using MSVC to compile or by switching to a mingw-w64 distribution with UCRT - would break compatibility for everything before Windows 10, but Windows 7 reached EOL in 2020 and Windows XP in 2014. URTC will properly supported for the years to come, were MSVCRT might get dropped somewhere in the foreseeable future. Either by mingw-w64 dropping support or dependencies forcing us to switch to URTC.

PR #1360, opened 2026-08-12
client: download filter, only check file presence after sanity checks

devnexen:
Backport of yquake2remaster e70b076. The extension blocklist becomes a table (adding .lib and .a) and the local presence check is moved behind the path and blocklist checks, so a hostile path is rejected before it is looked up on the filesystem.


Yamagi:
Thank you.

PR #1361, opened 2026-08-16
New cvar to centralize SBC optimizations

protocultor:
In #1128 I added two new cvars to handle two different optimization options for SoC / embedded devices. Since they both are used towards the same final purpose (optimize for SoC or not), and both are always enabled or disabled simultaneously, I believe the user would be more comfortable using only one.

The new cvar, `gl1_tilerendering`, is only available for GL1. This is mainly for two reasons:
1. New (official) Raspberry Pi OS versions have taken out the possibility to use GLES1. Having this switch makes practical for those OSs to enable the needed optimizations for this hardware.
2. GLES1 operates as if `gl1_tilerendering` is always enabled, just to avoid "_the paradox of choice_", since if you want this option disabled, using GL1 is probably the best choice from the start.

Related to this last point, I took the liberty to delete the `gl_discardfb` cvar of GLES3, recently added in #1339, and based in the now "removed" `gl1_discardfb`. Again, now it operates like it's "always 1", since you'll always need these _tile rendering_ optimizations if you have to choose GLES3 over GL3. It didn't have documentation anyway.

Also included in this PR:

- GL1 general code cleanup, plus a little bit of the same on GL3.
- Removed a few unnecesary OpenGL calls on GL3 init.
- Default for `gl1_pointparameters` in GLES1 is now `0`; in GL1 it remains `1`. Newer GPU drivers for SBCs show the needed extension as "_supported_" but [they actually don't do it](https://youtu.be/Wbp7OPADyUc?t=181).
- Fixed `vid_pauseonfocuslost` documentation. This didn't follow how it actually operates, according to the old menu option (31db2f7f).
- Added `GL_LINEAR` to the "_texture filter_" option in the Video menu. It's a user-friendly way to solve a graphical bug on Intel G41 Express Chipset ([see discussion](https://www.reddit.com/r/quake/comments/1tdhfot/help_quake_2_lighting_problem/)).

### DIFF
--- a/doc/040_cvarlist.md
+++ b/doc/040_cvarlist.md
@@ -530,9 +530,11 @@ it's `+set busywait 0` (setting the `busywait` cvar) and `-portable`
   show. Defaults to `300`.  
   Related to this: *cl_maxfps* and *cl_async*.
 
-* **vid_pauseonfocuslost**: When set to `1` the game is paused as soon
-  as it's window looses focus. It will work only in situation were the
-  game can be paused, e.g. not in multiplayer games. Defaults to `0`.
+* **vid_pauseonfocuslost**: When set to `0` (default) the game is paused
+  as soon as its window loses focus. It will only work where the game can
+  be paused, e.g. not in multiplayer. If `1`, the game continues to run
+  in the background. Setting it as `2` is the same as `0`, but the game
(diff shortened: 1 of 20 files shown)


Yamagi:
Thank you :)

PR #1371, opened 2026-09-10
game: fix floater activate frame array being one short

devnexen:


Yamagi:
Nice find. Do we need the same in xatrix and rogue?

BjossiAlfreds:
The activate move is never used in the code so the array and mmove object could simply be removed. But yes, I imagine this is also relevant to rogue, xatrix and zaero.

Yamagi:
> The activate move is never used in the code so the array and mmove object could simply be removed. But yes, I imagine this is also relevant to rogue, xatrix and zaero.

Let's merge it anyways to record the fix in the repo history. We can remove both at a later time.

Yamagi:
Thank you :)

PR #1376, opened 2026-09-22
github: make dependency update monthly

0lvin:


0lvin:
Security updates should come without limitation, but minor or patches should be skipped.

